Issue #13, opened 2015-07-03
Ranks not displayed in statistics


seeebek:
Hi,
EDMC 1.5.3.0 Windows shows no information about ranks in statistics (Trading, Combat, Exploring). It's a minor issue but issue non the less. 

Keep up the great work. You rock. ;-)

Marginal:
Sebastian,

It should look something like:
![trade](https://cloud.githubusercontent.com/assets/1374740/8505757/8eb6e394-21e9-11e5-90b1-2e68147fc3cf.jpeg)

If not, please [run from source](https://github.com/Marginal/EDMarketConnector#running-from-source), press "Update", and mail me the .json file that it produces in the working directory.

seeebek:
It's completely empty in my case. I will send you the data on sunday.

seeebek:
I sent you my json dump.

Here again the error:
```
Traceback (most recent call last):
  File "C:\Users\Sebastian\Downloads\EDMarketConnector-master\EDMarketConnector-
master\stats.py", line 332, in addranking
    if rank>=len(progress) or not progress[rank]['ts']:
TypeError: string indices must be integers
```

seeebek:
I added a PR fixing this issue for me:
#14

PR #14, opened 2015-07-05
Fix statistics

seeebek:
Fix rank indication in statistics (issue #13 )
It should generally work. You should test it anyway.


Marginal:
Thanks. Looks like this is an issue of how Frontier encode sparse arrays so I went with a more general solution in 8ce49fa.

PR #31, opened 2015-10-19
FIX: Error: Can't connect to EDSM

AnthorNet:


Marginal:
Oops. Thanks for that.
Fixed in Release 1.83.

PR #57, opened 2015-12-22
Update EDSM url

AnthorNet:


Marginal:
Ah I was already testing a patch to do the same thing - 75cf0d6
My version doesn't bother with the url lookup - it seems like extra work for the EDSM servers that will not be used much of the time ?

AnthorNet:
Your version is cleaner indeed ^^
Note that I sent fromSoftware every time and added a fdId parameter to track Frontier Id.

Marginal:
I've added fromSoftware in f239aec .
I will incorporate your fdId change.

Marginal:
Have implemented the same thing in parallel in 75cf0d6 and f239aec , so closing.

PR #64, opened 2015-12-27
EDMC++

SixByNine:
I really like EDMC, but I've been thinking that it would be nice to use the "live" data to feed a kind of trade computer that computes nearby good trades. I've started this by computing the current ship mass and jump range (not quite so easy as one would think). You probably don't want to merge this right now, but wondered if you were interested in this kind of feature and this seems the easiest way to discuss it.

I had to make a few minor changes to the core code to make the companion data structure available from outside, but most of the code lives in a separate module, activated from the file/apple menu.

Have a look and let me know what you think. The code quality is pretty rough, but it will get polished up as I go on.


Marginal:
I'm really not sure about this. It seems to be going down the path of duplicating the functionality of other existing tools both online (eddb.io, ed-td.space, ETN, Thrudd's, etc) and offline (Trade Dangerous, TCE, etc).

What are we bringing to the party?

wolverine2710:
I like the Unix philosopy. Small higly specialized and high quality tools which can be combined using the piping mechanism. Kind of what is happening now with EDMC. I'm all for new tools and I applaude Mikes initiave but personally  I would prefer EDMC to stay lean and mean. 

@Mike: An API (or plugin structure) for EDMC could be useful but then all functionality should be exposed, not only data which your or a specific tool needs. Imho such tools should not be part of the EDMC distribution but could be distributed as normal in the forums/websites - prerequisite EDMC installed ofc.. Haven't used EDMC in a while so can't recall. When the JSON structure received from FD is saved to a file you have all the info you need BUT you need to code more which is suboptimal.... Just my 2 eurocent.....

SixByNine:
Yeah... I agree about the "unix philosophy". In any case I really don't mind if you don't want to include this functionality - I'm happy to just keep it as a private tool. I have used tradedangerous, but it is a bit tedious to get it set up and is very slow to run. The online eddb.io tools are also good, but i) limited options and ii) also takes a while to set up. My aim was to basically run real-time analysis based on the current state of your ship. E.g. if you have cargo it will look up the best place nearby to sell. If you are at a station it will suggest good routes coming from here. I'd also like it to identify nearby stations that have not had the market updated in a while so that I can hunt for new markets. As far as I know there is not an easy way to do this with the current tools. I also want to make edmc detect when I have docked and automatically update the database as I'm really bad at remembering to press the update button.

For now I'm going to keep experimenting with this, but perhaps the best thing in the long run is to develop an alternative tool that interfaces with edmc somehow, though I'm also in wary of making "yet another trade tool".

Marginal:
> perhaps the best thing in the long run is to develop an alternative tool that interfaces with edmc somehow, though I'm also in wary of making "yet another trade tool"

With this patch I think you're already going down the path of making yet another trade tool!
You can interface to EDMC via the [command line](https://github.com/Marginal/EDMarketConnector/wiki/Running-from-source) executable on Windows. The OSX app bundle doesn't currently include the command-line executable, but should be fairly easy to include it.

Marginal:
If you're on Windows you may be interested in v2 of [Trade Computer Extension](https://forums.frontier.co.uk/showthread.php?t=219385) which uses the command-line EDMC to get its data.

inorton:
Hiya, I wonder if my plugin branch might fill some holes for you here? (you could write a plugin that feeds your other tool)

Marginal:
Closing since the [plugins](https://github.com/Marginal/EDMarketConnector/blob/master/PLUGINS.md) mechanism mentioned by inorton above and included in Release 2.05 seems like the best route to achieve this.

Issue #53, opened 2015-12-21
Settings screen too tall for some laptop resolutions


TinyBigJacko:
I'm currently running EDMC on a laptop with a native 1366 x 768 screen resolution, and the EDMC app is almost too tall to be able to fit into that screen under Windows 10. Probably the same in all Windows versions, I expect. It will *just* squeeze in and can be dragged up to reach the very tippy-top of the OK button when making changes, but it's a pain in the neck to move around (even using keystrokes and tab-presses to reach hidden boxes).

It'd be really useful if the screen's layout could be auto-resized, or reflowed onto a two-column, wider dialog-window, if at all possible.

Certainly if the screen height gets any taller, it will become impossible to use with a mouse, so please bear this in mind!

Thanks for an otherwise excellent and helpful app. Much appreciated.

Marginal:
Yes I'm struggling to keep it under 768, and will have to re-design if I add any more options.

inorton:
In my fork I've collapsed the authentication entries into tabs so might ease this problem

Marginal:
Fixed in Release 2.02

PR #67, opened 2015-12-30
Twitter Support

inorton:
When updating optionally tweet the current system / station.

Also reformatted the credentials for ED and Starmap (and twitter) into a notebook.

### DIFF
--- a/EDMarketConnector.py
+++ b/EDMarketConnector.py
@@ -15,6 +15,8 @@ import ttk
 import tkFont
 from ttkHyperlinkLabel import HyperlinkLabel
 
+import tweet
+
 if __debug__:
     from traceback import print_exc
 
@@ -307,10 +309,15 @@ class AppWindow:
                     if not self.status['text']:
                         self.status['text'] = _("You're not docked at a station!")
 
+                    if config.getint('output') & config.OUT_LOG_TWEET:
+                        tweet.systemchanged(self.system['text'])
+
                 else:
                     # Finally - the data looks sane and we're docked at a station
                     (station_id, has_market, has_outfitting, has_shipyard) = EDDB.station(self.system['text'], self.station['text'])
 
+                    if config.getint('output') & config.OUT_LOG_TWEET:
+                        tweet.stationchanged(self.system['text'], self.station['text'])
 
                     # No EDDN output at known station?
                     if (config.getint('output') & config.OUT_EDDN) and station_id and not has_market and not has_outfitting and not has_shipyard:
@@ -429,6 +436,9 @@ class AppWindow:
             self.station['text'] = EDDB.system(system) and self.STATION_UNDOCKED or ''
             if config.getint('output') & config.OUT_LOG_FILE:
                 flightlog.writelog(timestamp, system)
+            if config.getint('output') & config.OUT_LOG_TWEET:
+                tweet.systemchange(system)
+
             if config.getint('output') & config.OUT_LOG_EDSM:
                 try:
                     self.status['text'] = _('Sending data to EDSM...')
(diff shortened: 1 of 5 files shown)


Marginal:
Hi,

I suspect that registering for Consumer Keys and Access Tokens is going to be too daunting for the average user, and it isn't really appropriate to ask them to "[Create an Application](https://apps.twitter.com/app/new)". I believe the [standard approach](https://dev.twitter.com/oauth/3-legged) is to embed the Consumer Key + Secret in the app, ask the user to sign in to Twitter and grant access, and store the returned Access Token + Secret. See for example oauth_dance() in the twitter module. 

I'd be happy to accept a pull request that worked in this manner. (Obviously I'd regenerate my own Consumer Key + Secret for released binaries of the app).

Also, please address the following issues:
- Make text displayed to the user localizable by wrapping it in `_()` and using `format()` instead of piecing together strings. See [here](https://github.com/Marginal/EDMarketConnector/blob/master/EDMarketConnector.py#L472) for an example.
- Please test using the latest release of the Python twitter module installed with `easy_install twitter`. If you *do* need any of the patches since the last release, please add the twitter GitHub project as a submodule.
- I've re-worked the settings dialog to fix Issue #53. Sorry about that. Please re-integrate - I suggest by copying the code for the "EDSM" tab.

inorton:
Twitter have made it quite hard to have an open source client (you aren't supposed to publish the consumer secret in readable form)

One thing from this that jumped out. It would be quite nice to offer a kind of plugin interface. There are a number of events that are applicable to lots of purposes ( enter system, enter station, manual update )

I can think of a handful of tasks that would be nice to implement as plugins (freeing up the core to be developed itself) eg:

* auto-screenshot when arriving in a new system/station
* integration with another app/phone app (offer a small local webservice?)

Making the config GUI pluggable might be the hardest bit

Marginal:
> Twitter have made it quite hard to have an open source client (you aren't supposed to publish the consumer secret in readable form)

It's certainly a fragile system since it's hard to effectively hide a secret key from a user that has a copy of your software (as we've seen in the ongoing DRM wars).

But I'm not sure that it's *especially* hostile to open source. I would expect the source code to an open source client not to contain the Key or Secret, and the packager or distributor of binaries to supply them in at build time. I can think of some similar situations:
- AWS keys: Amazon [recommends](http://docs.aws.amazon.com/general/latest/gr/aws-access-keys-best-practices.html) that AWS keys are not embedded in code (and particularly in [published source code](http://www.theregister.co.uk/2015/01/06/dev_blunder_shows_github_crawling_with_keyslurping_bots/)).
- Code signing: I code sign my projects on Mac since the OSX UX strongly discourages users from installing unsigned apps. But I don't publish my signing key in GitHub; instead I leave a [placeholder](https://github.com/Marginal/EDMarketConnector/blob/master/setup.py#L36) that I fill in.
- Firefox: If you want to distribute your own build of Firefox you can do so, but you have to rename it (e.g. [IceWeasel](https://wiki.debian.org/Iceweasel)).

> One thing from this that jumped out. It would be quite nice to offer a kind of plugin interface.

It would be possible to offer a plugin interface which searched for plugin DLLs, let them register for various events and gave them their own tab on the settings dialog. Patches welcome I guess.

inorton:
Hiya, I'm going to close this. I'll have a think about a plugins kind of thing instead

Marginal:
> I'll have a think about a plugins kind of thing instead

If you like, but seems a little bit overall to me.
And any Twitter plugin will still face the issues that I mentioned [above](https://github.com/Marginal/EDMarketConnector/pull/67#issuecomment-168359607).

PR #75, opened 2016-01-08
Very Basic Plugin Support

inorton:
I've added a few tiny hooks and a very simple "About" plugin that doesn't do much. But it should allow for extensions to be written without disturbing the core of EDMC


Marginal:
Merged (but I lost the merge commit in a rebase - must remember to stop doing that) so closing.

PR #100, opened 2016-03-03
Fix crash when system isn't already known to EDSM

endlessfalls:
EDSM api is now returning "[]" instead of "-1" when the system name doesn't exist in its database. Fixes "Can't connect to EDSM" errors when jumping to a new system.


Marginal:
Thanks. Fixed in Release 2.10.

PR #112, opened 2016-05-27
Add log coordinates support from 1.6/2.1

AnthorNet:
I'm not sure it's the cleanest thing ^^ But it should work.
Still waiting for the API to do more testing.
Also not sure if EDProxy has coorinates support yet? And how, so only
pass None to the event.

### DIFF
--- a/EDMarketConnector.py
+++ b/EDMarketConnector.py
@@ -387,7 +387,7 @@ class AppWindow:
                     try:
                         self.status['text'] = _('Sending data to EDSM...')
                         self.w.update_idletasks()
-                        edsm.export(data, lambda:self.edsm.lookup(self.system['text'], EDDB.system(self.system['text'])))	# Do EDSM lookup during EDSM export
+                        edsm.export(data, None, lambda:self.edsm.lookup(self.system['text'], EDDB.system(self.system['text'])))	# Do EDSM lookup during EDSM export
                         self.status['text'] = ''
                     except Exception as e:
                         if __debug__: print_exc()
@@ -510,7 +510,7 @@ class AppWindow:
         except:
             pass
 
-    def system_change(self, timestamp, system):
+    def system_change(self, timestamp, system, coordinates):
 
         if self.system['text'] != system:
             self.system['text'] = system
@@ -526,7 +526,7 @@ class AppWindow:
                 try:
                     self.status['text'] = _('Sending data to EDSM...')
                     self.w.update_idletasks()
-                    edsm.writelog(timestamp, system, lambda:self.edsm.lookup(system, EDDB.system(system)))	# Do EDSM lookup during EDSM export
+                    edsm.writelog(timestamp, system, coordinates, lambda:self.edsm.lookup(system, EDDB.system(system)))	# Do EDSM lookup during EDSM export
                     self.status['text'] = strftime(_('Last updated at {HH}:{MM}:{SS}').format(HH='%H', MM='%M', SS='%S').encode('utf-8'), localtime(timestamp)).decode('utf-8')
                 except Exception as e:
                     if __debug__: print_exc()
--- a/PLUGINS.md
+++ b/PLUGINS.md
@@ -73,11 +73,11 @@ Your events all get called on the main tkinter loop so be sure not to block for 
 This gets called when EDMC uses the netlog to notice that you have arrived at a new star system.
 
 ```
-def system_changed(timestamp, system):
+def system_changed(timestamp, system, coordinates):
    """
    We arrived at a new system!
    """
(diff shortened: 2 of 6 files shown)


PLUGINS.md
```diff
@@ -73,11 +73,11 @@ Your events all get called on the main tkinter loop so be sure not to block for
 This gets called when EDMC uses the netlog to notice that you have arrived at a new star system.
 
 ```
-def system_changed(timestamp, system):
+def system_changed(timestamp, system, coordinates):
```
inorton: perhaps default this to None?

plug.py
```diff
@@ -98,7 +98,7 @@ def notify_system_changed(timestamp, system):
         system_changed = _get_plugin_func(plugname, "system_changed")
         if system_changed:
             try:
-                system_changed(timestamp, system)
+                system_changed(timestamp, system, coordinates)
```
inorton: Hmm. I wonder how we could prevent breakage of existing plugins here..

Marginal: Use co_argcount to detect old plugins.
Add a **kwargs parameter instead of an explicit coordinates parameter to avoid this problem in future.

edsm.py
```diff
@@ -102,23 +159,46 @@ def worker(self, system_name, result):
...
 
         if system in EDSM.FAKE:
             return
-
-        r = requests.get('http://www.edsm.net/api-logs-v1/set-log?commanderName=%s&apiKey=%s&systemName=%s&dateVisited=%s&fromSoftware=%s&fromSoftwareVersion=%s' % (urllib.quote(config.get('edsm_cmdrname').encode('utf-8')), urllib.quote(config.get('edsm_apikey')), urllib.quote(system), urllib.quote(time.strftime('%Y-%m-%d %H:%M:%S', time.gmtime(timestamp))), urllib.quote(applongname), urllib.quote(appversion)), timeout=EDSM._TIMEOUT)
+            
+        if coordinates:
+            (x,y,z) = coordinates
+            url = '%s/api-logs-v1/set-log?commanderName=%s&apiKey=%s&systemName=%s&dateVisited=%s&x=%s&y=%s&z=%s&fromSoftware=%s&fromSoftwareVersion=%s' % (
+                self.endPoint,
```
spacerecycler: The code fails to lookup the endPoint value here since the writelog method is not part of the EDSM class above.  I had to move the endPoint field in the EDSM class above to a global variable in this file to use it in this method.  Other than that it seems to be working.

EDMarketConnector.py
```diff
@@ -510,7 +510,7 @@ def retry_for_shipyard(self):
         except:
             pass
 
-    def system_change(self, timestamp, system):
+    def system_change(self, timestamp, system, coordinates):
```
spacerecycler: I had to change line 521 `plug.notify_system_changed(timestamp, system)` to `plug.notify_system_changed(timestamp, system, coordinates)` otherwise it caused an exception when I jumped to a new system.

EDMarketConnector.py
```diff
@@ -1,618 +1,618 @@
...
-        if event.widget == self.w:
-            theme.apply(self.w)
-
-
-# Run the app
-if __name__ == "__main__":
-    root = tk.Tk()
-    app = AppWindow(root)
-    root.mainloop()
+#!/usr/bin/python
```
inorton: is this all just line endings? what was it before you changed it?

edsm.py
```diff
@@ -30,70 +30,132 @@ def __init__(self):
     def link(self, system_name):
         self.cancel_lookup()
         if system_name in self.FAKE:
-            self.result = { 'img': '', 'url': None, 'done': True, 'uncharted': False }
+            self.result = {
```
inorton: While we can fairly easily see what you've done this is noise that detracts from the real code changes, it's code style that doesn't really fit with any tools either I think.

edsm.py
```diff
@@ -13,9 +13,9 @@
     from traceback import print_exc
 
 class EDSM:
-
-    _TIMEOUT = 10
-    FAKE = ['CQC', 'Training', 'Destination']	# Fake systems that shouldn't be sent to EDSM
+    endPoint    = 'https://www.edsm.net'
```
inorton: the extra whitespace here upsets my editors' pep8 plugin  :)

edsm.py
AnthorNet: Yeah sorry, but it was unreadable ^^

EDMarketConnector.py
AnthorNet: Strange did not notice that!
Will look at my history

edsm.py
AnthorNet: pep8 always seems unreadable to me ;)

PR #146, opened 2016-09-25
Add requirements file for use with pip

devdave:
To make development easier on others, requirements freezes the dependancies for edmc and then someone else can use ` pip install -r requirements`

### DIFF
--- a/requirements
+++ b/requirements
@@ -1,0 +1,6 @@
+argh==0.26.2
+pathtools==0.1.2
+PyYAML==3.12
+requests==2.11.1
+watchdog==0.8.3
+wheel==0.24.0


Marginal:
Thanks for the PR.

EDMC isn't doing anything particularly esoteric that requires particular versions of its dependences, so I'm not sure about the need to "feeze" dependencies. So would `>=` not be more polite than `==` ?

devdave:
Let me see if I can edit it, I just went with pip freeze defaults but yeah no difference on => vs ==

devdave:
Right, I've never done an in situ edit and pull request using the github web interface so this was educational.  Changed over the comparison operations but I wish there was a "I don't care, just give me the latest" operator.  >= * didn't work locally for me.

Marginal:
Thanks.

PR #158, opened 2016-11-11
Fix a bug in edshipyard.py

AndreasGocht:
When refreshing data at a station, I got an error that the global variable querytime was not defined.

It looks like querytime should be defined a few lines further down than it actually was, because where it was it didn't do anything. Everything worked fine after this fix.


Marginal:
Thanks.

Issue #199, opened 2017-04-24
ImportError: No module named collections


inorton:
Since modules.p was added I can no longer run the source build of EDMC, I suspect this is because my version of python is not 100% exactly the same as your one.  

I've found some background on this at http://stackoverflow.com/questions/24497214/python-cpickle-import-error

inorton:
Aha! on closer inspection, it is trying to load "OrderedDict\r" from a module called "collections\r"

inorton:
Turns out this is solved by doing `git config core.autocrlf true`

JediRhymeTrix:
Is there any way of removing the unnecessary \r's from a .p file without git? I found one somewhere and am facing the same issue and was wondering how to solve it. Any help would be greatly appreciated.

JediRhymeTrix:
@inorton Is there any way of removing the unnecessary \r's from a .p file without git? I found one somewhere and am facing the same issue and was wondering how to solve it. Any help would be greatly appreciated.

_Sent from my Galaxy Note8 using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

Marginal:
@JediRhymeTrix 
Open the file in universal newlines mode for reading- `open('file.p', 'rU')`.
For writing it's best to open pickled files in binary mode to avoid this issue.

JediRhymeTrix:
@Marginal I've tried rU but I'm still getting "no module named collections". Strangely, I only get this issue in Docker. It works fine on Windows 10 and Ubuntu even without rU.

_Sent from my Galaxy Note8 using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

PR #200, opened 2017-04-24
re #199 load pickled files in text mode rather than binary mode.

inorton:
I'm not 100% sure if this is the right approach, my instinct too would have been to use binary file mode but this seems to work here.


inorton:
Hm... no, maybe there is a better way

PR #201, opened 2017-04-25
re #199 load pickled files in text mode rather than binary mode.

inorton:
This forces the .p files to be checked out with \n as line endings rather than git converting them to \r\n


inorton:
nope, i just needed to do "git config core.autocrlf true"

PR #202, opened 2017-04-25
Add a tab to the Settings dialog. Users can directly open the plugins…

inorton:
From here you can open the plugins directory.

Can also disable plugins by adding ".disabled" to the end of their folder name.

image


inorton:
Thanks 👍 :)

PR #206, opened 2017-05-04
Terrible Salting Technique

Rubysea:
Really bad salt to reduce the possibility for looking up commander names with online md5 dycrypters. Also reference to current Salt Harvester number 1.


Scyntrus:
Why

PR #210, opened 2017-05-09
Remove erroneous newline from Russian translation file

Scyntrus:
This newline would cause the regex at l10n.py:66 to fail, leading to attempting to print the translation message at l10n.py:70. Since the translation message contains unicode characters, the print would fail which would cause the entire settings window to silently fail loading. There is an error message but the message doesn't appear to be fatal. Since the settings window fails to load, user credentials cannot be updated which may lead to the "Invalid Credentials" error. This is what's been happening to me and I have no idea why other people aren't encountering this bug.

### DIFF
--- a/L10n/ru.strings
+++ b/L10n/ru.strings
@@ -470,8 +470,7 @@
 "Theme" = "Тема";
 
 /* Help text in settings. [prefs.py] */
-"Tip: You can disable a plugin by{CR}adding '{EXT}' to it's folder name" = "Подсказка:
-Отключить расширение можно {CR}добавив '{EXT}' к имени папки";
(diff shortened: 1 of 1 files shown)


Scyntrus:
To clarify, the settings window appears to load, but since everything after prefs.py:275 doesn't run, the OK button isn't loaded and the binds to save data aren't run.

Marginal:
Thanks. Translations are managed over at [OneSky](https://marginal.oneskyapp.com/collaboration/project?id=52710) where this is [fixed](https://marginal.oneskyapp.com/collaboration/translate/project/project/52710/language/11#/?keyword=tip), and the fix will be in the next version.

PR #235, opened 2017-08-20
Document where configuration is stored

Athanasius:
Someone managed to slightly badly set the config such that the program wouldn't run, so needed to know where to look to delete/edit it. This wasn't documented in README.md (where this patch adds it) or on the wiki.


inorton:
Could you rebase (squash) your commits into one? that would make this easier to think about

Athanasius:
OK, bear with me, it'll be the first time I've needed to do a rebase in anger ;).

Athanasius:
Is that what you wanted ?

Marginal:
Thanks

Issue #298, opened 2018-03-08
MissionCompleted event - MaterialRewards - Category issue


ryczypior:
Whenever I get a mission with material rewards and I choose to get materials I've got an error:
```
Invalid journal entry "'{ "timestamp":"2018-03-08T00:33:28Z", "event":"MissionCompleted", "Faction":"The Winged Hussars", "Name":"Mission_Delivery_Boom_name", "MissionID":336255223, "Commodity":"$FruitAndVegetables_Name;", "Commodity_Localised":"Fruit and Vegetables", "Count":120, "TargetFaction":"HIP 115888 Transport Inc", "DestinationSystem":"HIP 115888", "DestinationStation":"Pilyugin Landing", "Reward":433920, "MaterialsReward":[ { "Name":"ArchivedEmissionData", "Name_Localised":"Irregular Emission Data", "Category":"$MICRORESOURCE_CATEGORY_Encoded;", "Category_Localised":"Encoded", "Count":5 } ], "FactionEffects":[ { "Faction":"HIP 115888 Transport Inc", "Effects":[ { "Effect":"$MISSIONUTIL_Interaction_Summary_famine_down;", "Effect_Localised":"After sourcing more supplies $#MinorFaction; have reduced the chances of famine.", "Trend":"DownGood" } ], "Influence":[ { "SystemAddress":1733119840994, "Trend":"UpGood" } ], "Reputation":"UpGood" }, { "Faction":"The Winged Hussars", "Effects":[ { "Effect":"$MISSIONUTIL_Interaction_Summary_boom_up;", "Effect_Localised":"$#MinorFaction; are experiencing increased growth that could lead to an economic boom", "Trend":"UpGood" } ], "Influence":[ { "SystemAddress":2869977163201, "Trend":"UpGood" } ], "Reputation":"UpGood" } ] }\n'"
Traceback (most recent call last):
  File "D:\git\EDMarketConnector\monitor.py", line 566, in parse_entry
    self.state[self.canonicalise(reward['Category']).capitalize()][material] += reward.get('Count', 1)
KeyError: u'$microresource_category_encoded'
```
I've found that for MissionCompleted event the MaterialsReward object has a "$MICRORESOURCE_CATEGORY_" prefix and doesn't have a "_name" suffix for Category property. I've made some changes in the monitor.py file which I'll provide in pull request

PR #299, opened 2018-03-08
Issue #298 MissionCompleted event - MaterialRewards - Category issue

ryczypior:
This patch repairs MissionCompleted even when choosing Materials reward


Marginal:
Thanks. Fixed by 78dd0c5 in a slightly different way.

PR #304, opened 2018-03-18
Strip bad filename parts from ship names

### LINKED ISSUE #303: Can't save ship loadout to disk if ship name contains question mark
```
[Errno 22] invalid mode ('wt') or filename: u"C:\Users\sporebat\Documents\What's the point?.2018-03-18T13.38.30.txt". Which is funny, as you already took the colons out of the ISO 8601 timestamp.

sporebat:
Fixes #303 

I could have left `slugify` in `monitor` and imported it from there intro `edshipyard`, but `loadout` also needed it and having either `monitor` or `loadout` reach into each other for a copy felt odd.

If you're in the mood for a more extensive clean-up, there's a lot of duplicate `filename = join(...)` code that could be extracted into `filenames` for the safety.

### DIFF
--- a/edshipyard.py
+++ b/edshipyard.py
@@ -14,6 +14,8 @@ from config import config
 import companion
 import outfitting
 from monitor import monitor
+from filenames import slugify
+
 
 # Map API ship names to E:D Shipyard ship names
 ship_map = dict(companion.ship_map)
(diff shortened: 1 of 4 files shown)


Marginal:
Thanks. Committed slightly different fix in  b102b8c.

sporebat:
Oh, that'll do.

PR #313, opened 2018-04-08
Load plugins in sorted order and introduce the idea of "library" plugins

inorton:
User-installed plugins are loaded from two folders, first "plugins/libs" and then "plugins" as normal so that plugins like EDMCOverlay can be installed once and loaded before others.

Also stop EDMC attempting to load random files (like logs) as plugins.

### DIFF
--- a/PLUGINS.md
+++ b/PLUGINS.md
@@ -178,3 +178,33 @@ EDMC now lets you disable a plugin without deleting it, simply rename the plugin
 
 Disabled and enabled plugins are listed on the "Plugins" Settings tab
 
+
+# Python Package Plugins
+
+EDMC supports writing plugins that provide features and python code for others plugins to call. These plugins are
+loaded before any others.
+
+A _Package Plugin_ is simply one that also contains a `__init__.py` file.
+
+This should also contain a standard `load.py` file containing a `plugin_start()` function even if the function is empty
+otherwise the plugin will not appear in the EDMC plugins tab.
+
+Eg,
+
+```
+plugins\Fish
+  __init__.py
+  load.py
+  fingers.py
+  cakes.py
+
+plugins\Cook
+  load.py
+```
+
+Given the above, the `Cook` plugin would be able to do:
+
+```
+from Fish import fingers, cakes
+```
+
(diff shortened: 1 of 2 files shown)


inorton:
*nudge* :)

inorton:
@Marginal I'd really like to get this in if possible.  There are already quite a few plugins that depend on my one.

Marginal:
Sorry, been busy.
I don't really understand the need for the extra complication. Cannot clients of EDMCOverlay just wait 'til their `plugin_app` callback, by which time EDMCOverlay will be loaded and started?

BTW did you see c8f2573 - as used e.g.: https://github.com/Marginal/EDMarketConnector/blob/f203b8bc7e2b409ebd68983b830921a93bd077ec/plugins/edsm.py#L247

inorton:
Oh that is quite handy.  

My main reason for wanting a `lib` folder was that there can be pure python modules and packages added that different plugins can import.  I kind of wanted users of edmcoverlay to just be able to import it's client library

Marginal:
Thanks. That use-case makes sense. But I don't like the additional opportunities for screwup that the "lib" folder will enable users to perform - a significant number of them seem to struggle with the current setup.

How about instead I change the load order so that plugins that contain a specially named file, say "preload", in their folder are loaded before those that don't?

inorton:
Let me tidy up this PR to remove the libs part. I'll add the "preload.py" thing in, that's a much better idea.

Marginal:
Not sure it's worth trying to load preload.py - there's not much you can do in it since the Plugin doesn't exist until we load load.py. Was just intending to use the presence of a file as a flag.

inorton:
Uh.. i think i messed this up.. just a mo

inorton:
Here we go,  I went for a different idea. Plugins can now be regular python packages, so we load plugins first if they contain an `__init__.py` file

Marginal:
Went for a slightly simpler implementation in fa3cf23 . 
Please let me know whether this works for you.

Marginal:
Note: One limitation of this approach of plugins importing Package Plugins is that the Package Plugin must have the correct name, and not e.g. "PackageFoo-master".

inorton:
Thanks, I'll give it a go!

PR #343, opened 2018-08-25
Fix line endings

inorton:
Without this I can't rebase from the official master or fast forward anything.

No matter which way I clone the official repo whether on windows or linux it always changes the line endings of these files


Marginal:
I assume this issue is due to the change to .gitattributes in 6bffcf4

PR #402, opened 2019-01-25
Update Error: Can't connect to EDDN

De-Wohli:
After an whole day of trying to find that error. Following SetUp:
Ubuntu box as router running dnsmasq + squid.
Windows 10 PC running ED:Market Connector.
Proxy settings inside Windows 10 set to the Ubuntu box.
Unable to connect to EDDN until `eddn.edcd.io` added to the exemption list in windows proxy settings.


Marginal:
Thanks. Fixed with aa1cc6a.

PR #424, opened 2019-06-29
Allow using a default value if config.get() returns None

vStone:


vStone:
We'll see if I still need this after the python 3.x merge.

PR #447, opened 2019-09-11
Eddn mandatory checks

### LINKED ISSUE #446: Setting StarPos can't assume this.coordinates is set
It seems that if you start EDMC after the game is already running (as I'm doing a lot during this port to Python3) you can't assume that this.coordinates is set.  I'm guessing this is because the journal monitor picks up with the next line in the journal, and thus hasn't seen any Location/Docked/FSDJump event in order to set this.

Hopefully this just causes the message to not be sent to EDDN ?

https://github.com/Marginal/EDMarketConnector/blob/86cd6c0f2c0c360735da2842af4da7426af78127/plugins/eddn.py#L436

Athanasius:
Fixes #446


Athanasius:
If there's any confusion here, I noobed up trying to clean this commit out of my master branch, removing it from this branch instead.  So I've cherry-picked it again and pushed, you want commit 558748d306b693978f0899055bab9db5858ab63f

Marginal:
Fixed in 02d4da3

PR #453, opened 2019-09-22
Python3 final

Athanasius:
Here's the big python3 pull.  This branch is a cleaned up rebase of my python3-2019 branch (where I applied and reverted things multiple times, including code that just shouldn't be upstream).  See the TODO.md file in that branch for an idea of what I've done.

1. I've tested as thoroughly as I can on Windows using python3.7 (both 64-bit and 32-bit).  Specifically Windows 10, game installed via Frontier's launcher, not Steam, standard Journal location ('Saved Games' not moved or any such).
1. I've done some perfunctory testing on Linux using Debian's Python 3.7 packages.  I don't have the game installed via Linux Steam or WINE.  Exports to files works, and copying a renamed Journal file in caused the expected event spam.  CAPI access works.
1. I can't test MacOS/darwin at all, but think I've updated any code in those paths.
1. I tried to keep changes as minimal as possible, but have gratuitously changed some "%" style formatting to "{}" style.
1. Changes in eddb.py are more intrusive as I didn't want to download the 3.5GB systems.csv on each run, so now it's expecting the files locally.  I added command-line switches to specify their location.  This requires 64-bit python else it just bombs out with a MemoryError.  I tested with the resulting new .p files.  I may look into trying to optimise this, as it took 3 hours to run on my i7-7700k, but that's for another branch and PR.
1. I left the coriolis-data sub-repo at the specific commit you had upstream when testing coriolis.py's generation of .p files.  Specifically a493df6 .
1. I've yet to try the py2exe flow.
1. I've not yet tried any third party plugins to see how they cope.

I'll be un-surprised if those more familiar with the code find issues.


Athanasius:
I forgot in point 1 that the ctypes-mediated Registry access on Windows only works in 32-bit python.  If this is an issue then I'll look into replacing all of that with the winreg module.

Athanasius:
That conflict in outfitting.py is simple enough to resolve, we put the voicepacks bit in a different position is all.

Athanasius:
I've just rebased locally on origin/master, will now need to test things.  Stand by.

Athanasius:
Rebased on origin/master c5df83eb32785d25e9b4da9abc144119cd701d65 .

Athanasius:
That's the safe .wxs tcl/tk culling done.  I can get things to work with even fewer files, but don't know tcl/tk well enough to judge if future code/user plugins might need something.

BarneyKB:
this looks good, would this add support for python 3 plugin development? I'd love to make some but the hassle of working with py2 is too much :(

Athanasius:
All the code for this has now been merged into the `python3` branch, which at this point should also be function/bug synchronised with the current state of the `release-final-python27` branch.

So closing this.

PR #454, opened 2019-09-22
Adds "shell=True" to npm callout, so that it works on Windows.

Athanasius:


coriolis.py
```diff
@@ -25,7 +25,7 @@ def add(modules, name, attributes):
         modules[name] = attributes
 
     # Regenerate coriolis-data distribution
-    subprocess.check_call(['npm', 'install'], cwd='coriolis-data', stdout=sys.stdout, stderr=sys.stderr)
+    subprocess.check_call(['npm', 'install'], cwd='coriolis-data', stdout=sys.stdout, stderr=sys.stderr, shell=True)
```
Marginal: Needs to be `subprocess.check_call(['npm install'], cwd='coriolis-data', shell=True, stdout=sys.stdout, stderr=sys.stderr)` for portability.

Issue #475, opened 2020-01-20
Error: Inara setCommanderReputationMinorFaction, The reputation value exceeds the valid range


louisrousseau:
Running 3.4.3.0, I have gotten this error message in the status bar, did not notice exactly when it did so unsure what conditions might trigger it for the moment.

Inara setCommanderReputationMinorFaction, The reputation value exceeds the valid range

Wikibear:
I can confirm this issue.

TheInfamousAlk:
Can confirm this issue as well.

ZTiKnl:
same here, and here is the corresponding part of EDMarketConnector.log

`Inara	202 Everything was alright, the near-neutral status just wasn't stored.	{"eventName": "setCommanderReputationMinorFaction", "eventTimestamp": "2020-02-18T16:27:00Z", "eventData": [{"minorfactionName": "Mooramba aristocrats", "minorfactionReputation": 0.0}, {"minorfactionName": "Vagha Travel Commodities", "minorfactionReputation": 0.0}, {"minorfactionName": "Sirius Hyperspace", "minorfactionReputation": 0.0}, {"minorfactionName": "Cail Autocracy", "minorfactionReputation": 0.0}, {"minorfactionName": "Pilots' Federation Local Branch", "minorfactionReputation": 75.664299}, {"minorfactionName": "Leleng Musu Gold Family", "minorfactionReputation": 0.0}, {"minorfactionName": "Integrated Mooramba", "minorfactionReputation": 0.0}, {"minorfactionName": "Massagetae Blue Camorra", "minorfactionReputation": 0.0}]}`

ZTiKnl:
I recall something from the last (ED) update being another decimal point being added to some type of reputation?
Not sure if this is connected though...

ZTiKnl:
latest journal file (removed most unrelated data):

```
{ 
"timestamp":"2020-02-18T16:33:08Z", 
"event":"FSDJump", 
"StarSystem":"Varam", 
"SystemAddress":5068196095409, 
"Factions":[ 
  { 
    "Name":"Pilots' Federation Local Branch", 
    "Influence":0.000000, 
    "MyReputation":75.664299 
  }, { 
    "Name":"Varam Blue Legal Industries", "Influence":0.093000, "MyReputation":30.860001 
  }, { 
    "Name":"Varam People's Party", "Influence":0.110000, "MyReputation":5.600000
  }, { 
[...]

PR #496, opened 2020-05-27
Sync with master

bgol:
Includes #479 (bugfix for #475)


Athanasius:
I won't be merging this as-is, not least because that 738395d commit snuck into it.  It took me a PR or two to figure out that you really want to make a branch in your fork for literally only the commits you want in the PR.  Anything else you do in the branch will automatically become a part of the PR.

I will, however, use this as reference for ensuring no work gets left ouf of master when I merge the python3 branch into it.  Thanks.

bgol:
That's ok. Take what you want/need. You can cancel the current pull request(s) until you have a stable version which works for you and than I can sync and remake the carrier one. We wont need both anymore.

Athanasius:
I'll review the other PRs as I suspect they'll just merge cleanly once I'm at that stage anyway.  If not I'll be in touch on each specific PR.

PR #502, opened 2020-06-08
Added line to Linux installation instructions to tell Ubuntu 20.04 us…

alex-williams:
…ers to use python3 branch to avoid confusion.


Athanasius:
This will be moot in the next day or so, as I merge the python3 code into master.

PR #530, opened 2020-06-27
Issue/529 update translations

Athanasius:
Updates currently in sync with OneSky, and all new strings since last sync added.


Athanasius:
I still need to check the new translations definitely work.

Athanasius:
In testing that "other translations within a string" thing I briefly used French, and it did indeed all work.  This was specifically the `Popup body: Warning about plugins without Python 3.x support [EDMarketConnector.py]` translation.  Sure, most of it was still in English at this point, but the File/Settings/Plugins parts got translated, which indicates the whole string was correctly going through translation.

Athanasius:
As I've added "Serbian (Latin, Bosnia and Herzegovina)", which mIRCon has translated, I should also see about adding that to the list of Languages in EDMarketConnector.wxs.  Need to find where to look up the numeric code for it.

PR #540, opened 2020-07-01
Fix/407 upload output settings reset

### LINKED ISSUE #407: Upload/output settings re-enable themselves if user has disabled all upload/output settings
Due to the logic in these lines: https://github.com/Marginal/EDMarketConnector/blob/04563cadaefca0213faf8a133dc3aeb12b5df50c/prefs.py#L104
https://github.com/Marginal/EDMarketConnector/blob/1dc6d04954c7b9b0bad537510700299db50611c6/plugins/eddn.py#L350
Ship loadout output, and EDDN sending options, will attempt to restore their default (enabled) state if `output` is set to `0`, i.e. all file outputs and EDDN uploads have been disabled. The unexpected re-enabling of these settings may give the impression that the configuration failed to "stick" when the user has explicitly disabled them previously (e.g., if they use another tool for EDDN upload).

Athanasius:
Fixes #407 by using a 'PrefsVersion' to determine if default should be set or not.

### DIFF
--- a/plugins/eddn.py
+++ b/plugins/eddn.py
@@ -14,6 +14,8 @@ import tkinter as tk
 from ttkHyperlinkLabel import HyperlinkLabel
 import myNotebook as nb
 
+from prefs import prefsVersion
+
 if sys.platform != 'win32':
     from fcntl import lockf, LOCK_EX, LOCK_NB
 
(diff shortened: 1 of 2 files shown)


Athanasius:
NB: If someone happens to have the 'magic' combination of settings that means 'output' is 0x0 then the first time they open Settings they'll once more get the defaults.  After they close that 'PrefsVersion' will be saved and this code will start working.

PR #565, opened 2020-07-07
Cleanup companion.py

A-UNDERSCORE-D:
Code cleanup and PEP8 compatibility.

### DIFF
--- a/companion.py
+++ b/companion.py
@@ -22,10 +22,14 @@ import zlib
 from config import appname, appversion, config
 from protocol import protocolhandler
 
+from typing import TYPE_CHECKING
+if TYPE_CHECKING:
+    _ = lambda x: x # noqa
 
-holdoff = 60	# be nice
-timeout = 10	# requests timeout
-auth_timeout = 30	# timeout for initial auth
+
+holdoff = 60  # be nice
+timeout = 10  # requests timeout
+auth_timeout = 30  # timeout for initial auth
 
 # Currently the "Elite Dangerous Market Connector (EDCD/Athanasius)" one in
 # Athanasius' Frontier account
@@ -47,7 +51,7 @@ category_map = {
     'Narcotics'     : 'Legal Drugs',
     'Slaves'        : 'Slavery',
     'Waste '        : 'Waste',
-    'NonMarketable' : False,	# Don't appear in the in-game market so don't report
+    'NonMarketable' : False  # Don't appear in the in-game market so don't report
 }
 
 commodity_map = {}
@@ -105,47 +109,62 @@ ship_map = {
 # In practice these arrays aren't very sparse so just convert them to lists with any 'gaps' holding None.
 def listify(thing):
     if thing is None:
-        return []	# data is not present
+        return []  # data is not present
+
     elif isinstance(thing, list):
-        return list(thing)	# array is not sparse
+        return list(thing)  # array is not sparse
+
     elif isinstance(thing, dict):
         retval = []
-        for k,v in thing.items():
+        for k, v in thing.items():
             idx = int(k)
+
             if idx >= len(retval):
                 retval.extend([None] * (idx - len(retval)))
                 retval.append(v)
             else:
                 retval[idx] = v
+
         return retval
+
     else:
-        assert False, thing	# we expect an array or a sparse array
-        return list(thing)	# hope for the best
+        assert False, thing  # we expect an array or a sparse array
+        return list(thing)  # hope for the best
 
 
 class ServerError(Exception):
     def __init__(self, *args):
-        self.args = args if args else (_('Error: Frontier server is down'),)	# Raised when cannot contact the Companion API server
+        # Raised when cannot contact the Companion API server
+        self.args = args if args else (_('Error: Frontier server is down'),)
+
 
 class ServerLagging(Exception):
     def __init__(self, *args):
-        self.args = args if args else (_('Error: Frontier server is lagging'),)	# Raised when Companion API server is returning old data, e.g. when the servers are too busy
+        # Raised when Companion API server is returning old data, e.g. when the servers are too busy
+        self.args = args if args else (_('Error: Frontier server is lagging'),)
+
 
 class SKUError(Exception):
     def __init__(self, *args):
-        self.args = args if args else (_('Error: Frontier server SKU problem'),)	# Raised when the Companion API server thinks that the user has not purchased E:D. i.e. doesn't have the correct 'SKU'
+        # Raised when the Companion API server thinks that the user has not purchased E:D
+        # i.e. doesn't have the correct 'SKU'
+        self.args = args if args else (_('Error: Frontier server SKU problem'),)
+
 
 class CredentialsError(Exception):
     def __init__(self, *args):
         self.args = args if args else (_('Error: Invalid Credentials'),)
 
+
 class CmdrError(Exception):
     def __init__(self, *args):
-        self.args = args if args else (_('Error: Wrong Cmdr'),)	# Raised when the user has multiple accounts and the username/password setting is not for the account they're currently playing OR the user has reset their Cmdr and the Companion API server is still returning data for the old Cmdr
+        # Raised when the user has multiple accounts and the username/password setting is not for
+        # the account they're currently playing OR the user has reset their Cmdr and the Companion API
+        # server is still returning data for the old Cmdr
+        self.args = args if args else (_('Error: Wrong Cmdr'),)
 
 
 class Auth(object):
-
     def __init__(self, cmdr):
         self.cmdr = cmdr
         self.session = requests.Session()
@@ -166,19 +185,23 @@ class Auth(object):
                     'client_id': CLIENT_ID,
                     'refresh_token': tokens[idx],
                 }
+
                 r = self.session.post(SERVER_AUTH + URL_TOKEN, data=data, timeout=auth_timeout)
                 if r.status_code == requests.codes.ok:
                     data = r.json()
                     tokens[idx] = data.get('refresh_token', '')
                     config.set('fdev_apikeys', tokens)
-                    config.save()	# Save settings now for use by command-line app
+                    config.save()  # Save settings now for use by command-line app
                     return data.get('access_token')
+
                 else:
                     print('Auth\tCan\'t refresh token for %s' % self.cmdr)
                     self.dump(r)
+
             except:
                 print('Auth\tCan\'t refresh token for %s' % self.cmdr)
                 print_exc()
+
         else:
             print('Auth\tNo token for %s' % self.cmdr)
 
@@ -193,23 +216,26 @@ class Auth(object):
 
     def authorize(self, payload):
         # Handle OAuth authorization code callback. Returns access token if successful, otherwise raises CredentialsError
-        if not '?' in payload:
+        if '?' not in payload:
             print('Auth\tMalformed response "%s"' % payload)
-            raise CredentialsError()	# Not well formed
+            raise CredentialsError()  # Not well formed
 
         data = urllib.parse.parse_qs(payload[payload.index('?')+1:])
         if not self.state or not data.get('state') or data['state'][0] != self.state:
             print('Auth\tUnexpected response "%s"' % payload)
-            raise CredentialsError()	# Unexpected reply
+            raise CredentialsError()  # Unexpected reply
 
         if not data.get('code'):
             print('Auth\tNegative response "%s"' % payload)
             if data.get('error_description'):
                 raise CredentialsError('Error: %s' % data['error_description'][0])
+
             elif data.get('error'):
                 raise CredentialsError('Error: %s' % data['error'][0])
+
             elif data.get('message'):
                 raise CredentialsError('Error: %s' % data['message'][0])
+
             else:
                 raise CredentialsError()
 
@@ -222,6 +248,7 @@ class Auth(object):
                 'code': data['code'][0],
                 'redirect_uri': protocolhandler.redirect,
             }
+
             r = self.session.post(SERVER_AUTH + URL_TOKEN, data=data, timeout=auth_timeout)
             data = r.json()
             if r.status_code == requests.codes.ok:
@@ -232,8 +259,9 @@ class Auth(object):
                 tokens = tokens + [''] * (len(cmdrs) - len(tokens))
                 tokens[idx] = data.get('refresh_token', '')
                 config.set('fdev_apikeys', tokens)
-                config.save()	# Save settings now for use by command-line app
+                config.save()  # Save settings now for use by command-line app
                 return data.get('access_token')
+
         except:
             print('Auth\tCan\'t get token for %s' % self.cmdr)
             print_exc()
@@ -244,10 +272,13 @@ class Auth(object):
         self.dump(r)
         if data.get('error_description'):
             raise CredentialsError('Error: %s' % data['error_description'])
+
         elif data.get('error'):
             raise CredentialsError('Error: %s' % data['error'])
+
         elif data.get('message'):
             raise CredentialsError('Error: %s' % data['message'])
+
         else:
             raise CredentialsError()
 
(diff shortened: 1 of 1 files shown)


companion.py
```diff
@@ -4,32 +4,35 @@
...
 import time
 from traceback import print_exc
 import urllib.parse
 import webbrowser
-import zlib
 
 from config import appname, appversion, config
 from protocol import protocolhandler
 
+from typing import TYPE_CHECKING
```
A-UNDERSCORE-D, line 23: Dont forget to remove this before merge

companion.py
```diff
@@ -222,83 +250,85 @@ def authorize(self, payload):
...
         self.state = Session.STATE_INIT
         self.credentials = None
         self.session = None
         self.auth = None
-        self.retrying = False	# Avoid infinite loop when successful auth / unsuccessful query
+        self.retrying = False  # Avoid infinite loop when successful auth / unsuccessful query
 
         # yuck suppress InsecurePlatformWarning under Python < 2.7.9 which lacks SNI support
-        if sys.version_info < (2,7,9):
+        if sys.version_info < (2, 7, 9):
```
A-UNDERSCORE-D, line 310: This probably doesnt need to exist anymore, given that this file is not py2 compatible and the intent is to remove py2 support

companion.py
```diff
@@ -222,83 +250,85 @@ def authorize(self, payload):
...
     def login(self, cmdr=None, is_beta=None):
         # Returns True if login succeeded, False if re-authorization initiated.
         if not CLIENT_ID:
             raise CredentialsError()
+
         if not cmdr or is_beta is None:
             # Use existing credentials
             if not self.credentials:
-                raise CredentialsError()	# Shouldn't happen
+                raise CredentialsError()  # Shouldn't happen
```
A-UNDERSCORE-D, line 322: error message

companion.py
```diff
@@ -320,32 +352,38 @@ def login(self, cmdr=None, is_beta=None):
     # Callback from protocol handler
     def auth_callback(self):
         if self.state != Session.STATE_AUTH:
-            raise CredentialsError()	# Shouldn't be getting a callback
+            raise CredentialsError()  # Shouldn't be getting a callback
```
A-UNDERSCORE-D, line 355: error message

companion.py
```diff
@@ -320,32 +352,38 @@ def login(self, cmdr=None, is_beta=None):
...
             self.auth = None
-            raise	# Bad thing happened
+            raise  # Bad thing happened
 
     def start(self, access_token):
         self.session = requests.Session()
-        self.session.headers['Authorization'] = 'Bearer %s' % access_token
-        self.session.headers['User-Agent'] = 'EDCD-%s-%s' % (appname, appversion)
+        self.session.headers['Authorization'] = 'Bearer {}'.format(access_token)
+        self.session.headers['User-Agent'] = 'EDCD-{appname}-{version}'.format(appname=appname, version=appversion)
```
A-UNDERSCORE-D, line 369: this is done in a few places. Constant?

companion.py
```diff
@@ -355,33 +393,39 @@ def query(self, endpoint):
...
         elif 500 <= r.status_code < 600:
             # Server error. Typically 500 "Internal Server Error" if server is down
             self.dump(r)
             raise ServerError()
 
         try:
-            r.raise_for_status()	# Typically 403 "Forbidden" on token expiry
-            data = r.json()		# May also fail here if token expired since response is empty
+            r.raise_for_status()  # Typically 403 "Forbidden" on token expiry
+            data = r.json()  # May also fail here if token expired since response is empty
```
A-UNDERSCORE-D, line 404: These two are the lines that can throw exceptions, the below except clause should move to catching the HTTP exceptions (or just checking them instead of using raise_for_status, and the r.json can throw only so many errors

companion.py
```diff
@@ -355,33 +393,39 @@ def query(self, endpoint):
...
                 self.login()
                 raise CredentialsError()
+
             elif self.login():		# Maybe our token expired. Re-authorize in any case
                 self.retrying = True
                 return self.query(endpoint)
+
             else:
                 self.retrying = False
                 raise CredentialsError()
```
A-UNDERSCORE-D, line 423: error message

companion.py
```diff
@@ -391,29 +435,36 @@ def station(self):
         data = self.query(URL_QUERY)
         if data['commander'].get('docked'):
             services = data['lastStarport'].get('services', {})
+
             if services.get('commodities'):
+
                 marketdata = self.query(URL_MARKET)
-                if (data['lastStarport']['name'] != marketdata['name'] or
-                    int(data['lastStarport']['id']) != int(marketdata['id'])):
+                if (data['lastStarport']['name'] != marketdata['name'] or int(data['lastStarport']['id']) != int(marketdata['id'])):
```
A-UNDERSCORE-D, line 442: This is doing more weird magic with ors, check how it works and make it clearer

companion.py
```diff
@@ -391,29 +435,36 @@ def station(self):
...
                     raise ServerLagging()
+
                 else:
                     data['lastStarport'].update(marketdata)
+
             if services.get('outfitting') or services.get('shipyard'):
                 shipdata = self.query(URL_SHIPYARD)
-                if (data['lastStarport']['name'] != shipdata['name'] or
-                    int(data['lastStarport']['id']) != int(shipdata['id'])):
+                if (data['lastStarport']['name'] != shipdata['name'] or int(data['lastStarport']['id']) != int(shipdata['id'])):
```
A-UNDERSCORE-D, line 450: This is doing more weird magic with ors, check how it works and make it clearer

companion.py
```diff
@@ -425,45 +476,69 @@ def dump(self, r):
...
 
-# Returns a shallow copy of the received data suitable for export to older tools - English commodity names and anomalies fixed up
+# Returns a shallow copy of the received data suitable for export to older tools
+# English commodity names and anomalies fixed up
 def fixup(data):
-
     if not commodity_map:
         # Lazily populate
-        for f in ['commodity.csv', 'rare_commodity.csv']:
+        for f in ('commodity.csv', 'rare_commodity.csv'):
```
A-UNDERSCORE-D, line 484: need this be a loop over a simple function and two calls?

companion.py
```diff
@@ -425,45 +476,69 @@ def dump(self, r):
...
+
             elif not commodity['stockBracket'] in range(4):
-                if __debug__: print('Invalid "stockBracket":"%s" for "%s"' % (commodity['stockBracket'], commodity['name']))
+                if __debug__:
+                    print('Invalid "stockBracket":{!r} for {!r}'.format(commodity['stockBracket'], commodity['name']))
+
             else:
                 # Rewrite text fields
-                new = dict(commodity)	# shallow copy
+                new = dict(commodity)  # shallow copy
```
A-UNDERSCORE-D, line 541: `dict.copy()` does a shallow copy already, perhaps use that instead of this, for clarity.

companion.py
```diff
@@ -488,22 +563,28 @@ def fixup(data):
...
+            elif k in ('alive', 'cargo', 'cockpitBreached', 'health', 'oxygenRemaining',
+                       'rebuilds', 'starsystem', 'station'):
+                pass  # noisy
+
+            elif k in ('locDescription', 'locName') or k.endswith('LocDescription') or k.endswith('LocName'):
+                pass  # also noisy, and redundant
+
+            elif k in ('dir', 'LessIsGood'):
+                pass  # dir is not ASCII - remove to simplify handling
+
```
A-UNDERSCORE-D, line 581: This can probably be replaced with a `filter()` and a few constants.

A-UNDERSCORE-D: Looking at this some more, while it could be done that way, it wouldnt be any cleaner.

companion.py
Athanasius: As this is essentially just doing `TYPE_CHECKING = False` I don't see the harm in leaving it in.  If/when we make more use of `typing` we can address it then.

companion.py
Athanasius: Yes, rip out anything that's specific to Python < 3.0.

companion.py
```diff
@@ -355,33 +396,39 @@ def query(self, endpoint):
...
+            raise ServerError('Received error {} from server'.format(r.status_code))
 
         try:
-            r.raise_for_status()	# Typically 403 "Forbidden" on token expiry
-            data = r.json()		# May also fail here if token expired since response is empty
-        except:
+            r.raise_for_status()  # Typically 403 "Forbidden" on token expiry
+            data = r.json()  # May also fail here if token expired since response is empty
+
+        except (requests.HTTPError, ValueError) as e:
```
Athanasius, line 409: Perhaps `json.JSONDecodeError` instead of `ValueError`.  Heck, separate `except ..:` clause so that use can be made of the specific properties if needs be ?

Athanasius: Oh, hang on, that's still in *requests* not the actual json decode.

Athanasius: Except that *is* calling through to `json.loads()` if my reading of pydoc for `requests.Response.json` is correct.  But it must be catching the json specific exception in order to raise the ValueError.  So yeah, what you have is what it will need to be.

It's that or we explicitly call `json.loads()` ourselves on the r.content.  Leave it as is for now, we can start fiddling if we ever see the need for more fine-detailed error reporting.

companion.py
A-UNDERSCORE-D: True, my main concern is that Im not entirely sure how _ is hacked in, and whether or not there is any checking on whether or not there are any assignments ever. during normal runtime TYPE_CHECKING should always be false

companion.py
```diff
@@ -4,35 +4,40 @@
...
+
+holdoff = 60  # be nice
+timeout = 10  # requests timeout
+auth_timeout = 30  # timeout for initial auth
 
 # Currently the "Elite Dangerous Market Connector (EDCD/Athanasius)" one in
 # Athanasius' Frontier account
-CLIENT_ID   = os.getenv('CLIENT_ID') or 'fb88d428-9110-475f-a3d2-dc151c2b9c7a' 	# Obtain from https://auth.frontierstore.net/client/signup
+# Obtain from https://auth.frontierstore.net/client/signup
+CLIENT_ID   = os.getenv('CLIENT_ID') or 'fb88d428-9110-475f-a3d2-dc151c2b9c7a'
```
A-UNDERSCORE-D, line 35: This does or magic. Maybe replace? Undecided.

companion.py
```diff
@@ -105,51 +110,66 @@
...
                 retval.append(v)
             else:
                 retval[idx] = v
+
         return retval
+
     else:
-        assert False, thing	# we expect an array or a sparse array
-        return list(thing)	# hope for the best
+        assert False, thing  # we expect an array or a sparse array
```
A-UNDERSCORE-D, line 132: `assert False` always raises. Not entirely sure why this is here. Replacing it with a `raise ValueError` may be prudent

companion.py
Athanasius: Nah, this one is fine.  When it's just `<possibly false thing> or <alternative>` I'm OK with it.

Athanasius: It would be a different matter if we added a way to specify a CAPI CLIENT_ID on the commandline, then it'd be a three way and really require a proper if statement.  But that's for another PR.

companion.py
Athanasius: Just be sure to check callers.  If they're catching exceptions including AssertionError then we need to be sure they'll catch ValueError.  Of course they're probably just doing bare `except:` ;) .

A-UNDERSCORE-D: Appears that the only caller is https://github.com/EDCD/EDMarketConnector/blob/develop/stats.py#L157 (according to my IDE and a `grep 'listify' **/*.py`, anyway). So probably okay to switch to a ValueError?

companion.py
```diff
@@ -4,35 +4,40 @@
 import base64
 import csv
 import requests
-from http.cookiejar import LWPCookieJar	# No longer needed but retained in case plugins use it
+from http.cookiejar import LWPCookieJar  # No longer needed but retained in case plugins use it
```
Athanasius, line 7: I need to make a note to look into turning "in case plugins use it" into an extra explicit include in setup_py instead.

companion.py
Athanasius: Well, running this branch and switching to French appears to still be translating everything without throwing exceptions or errors.

Athanasius: Ah, but in companion.py itself the only use of `_()` is in errors that I haven't triggered yet.  Time to add something temporarily to be sure.

Athanasius: `print('Testing translation: "{KEY}" -> "{TRANSLATED}"'.format(KEY='Error: Wrong Cmdr', TRANSLATED=_('Error: Wrong Cmdr')))` does correctly yield `Testing translation: "Error: Wrong Cmdr" -> "Erreur : Cmdr incorrect"` when placed in the Session.login() method, right at the start.

companion.py
```diff
@@ -47,7 +52,7 @@
     'Narcotics'     : 'Legal Drugs',
     'Slaves'        : 'Slavery',
     'Waste '        : 'Waste',
-    'NonMarketable' : False,	# Don't appear in the in-game market so don't report
+    'NonMarketable' : False  # Don't appear in the in-game market so don't report
```
Athanasius, line 55: I prefer the trailing comma stay, makes adding a next line less work.  A diff would show only the new line, not the addition of the command to the prior line as well.

companion.py
Athanasius: Having traced that back up the possible call stacks, yeah, `raise ValueError` seems good.

companion.py
```diff
@@ -189,29 +213,35 @@ def refresh(self):
...
-            print('Auth\tMalformed response "%s"' % payload)
-            raise CredentialsError()	# Not well formed
+        # Handle OAuth authorization code callback.
+        # Returns access token if successful, otherwise raises CredentialsError
+        if '?' not in payload:
+            print('Auth\tMalformed response {!r}'.format(payload))
+            raise CredentialsError('malformed payload')  # Not well formed
 
-        data = urllib.parse.parse_qs(payload[payload.index('?')+1:])
+        data = urllib.parse.parse_qs(payload[payload.index('?') + 1:])
```
Athanasius, line 234: That `1:` looks weird.  I know what it's doing, but would it hurt to throw `()` around the calculation to make it obvious it's `<this value, plus 1>:` ?

companion.py
```diff
@@ -222,83 +252,80 @@ def authorize(self, payload):
...
-        if sys.version_info < (2,7,9):
-            from requests.packages import urllib3
-            urllib3.disable_warnings()
+        self.retrying = False  # Avoid infinite loop when successful auth / unsuccessful query
 
     def login(self, cmdr=None, is_beta=None):
         # Returns True if login succeeded, False if re-authorization initiated.
         if not CLIENT_ID:
-            raise CredentialsError()
+            raise CredentialsError('cannot login without a valid client ID')
```
Athanasius, line 314: "cannot login without a valid Client ID"

Yes, I spent some time looking up and settling on terms when documenting the Frontier oAuth2 stuff -> https://github.com/Athanasius/fd-api/blob/master/docs/FrontierDevelopments-oAuth2-notes.md

companion.py
```diff
@@ -355,33 +391,39 @@ def query(self, endpoint):
...
+                raise CredentialsError('query failed after refresh') from e
+
             elif self.login():		# Maybe our token expired. Re-authorize in any case
                 self.retrying = True
                 return self.query(endpoint)
+
             else:
                 self.retrying = False
-                raise CredentialsError()
+                raise CredentialsError('Invalid JSON or HTTP error') from e
```
Athanasius, line 421: This can be read as `Invalid (JSON or HTTP error)`.  Swap them around -> "HTTP error or invalid JSON".

companion.py
A-UNDERSCORE-D: Cool, then leaving it in should be fine, will do so

companion.py
A-UNDERSCORE-D: can do, though I feel that its clear enough without it, due to order of operations it'll never hit the : first

companion.py
A-UNDERSCORE-D: Raise added, with a message from the comment (and the given value, as the assert had)

A-UNDERSCORE-D:
Okay this is merge-able if everyone is happy

PR #573, opened 2020-07-09
Fixed updating inara cargo only on major events

### LINKED ISSUE #477: "Update" button doesn't send cargo info to inara
I've noticed that if I hit the "Update" button my cargo list isn't updated on inara -- it seems that cargo is only sent when I first log in to elite. 

This makes material trading frustrating because it's difficult to see what I still need after I've made trades.

Repro:

1. Make material trades
2. Click 'update'
3. Inara doesn't show updated cargo list.
4. Log out of elite
5. Log back in
6. Refresh inara -- updated cargo list is there.

A-UNDERSCORE-D:
This works by checking on _any_ event the status of the cargo vs the
current stored version. If the live cargo differs from the updated
cargo, _and_ we have not sent any update in the past 45 seconds (as this
would have contained the cargo anyway), we queue an actual send to the
API.

This unfortunately isn't checked only when docked, but on the plus side
miners will be happy to have their cargo updated relatively often.

Fixes #477. Possibly not quite what they wanted, but its about as close
as this will get

### DIFF
--- a/plugins/inara.py
+++ b/plugins/inara.py
@@ -413,14 +413,10 @@ def journal_entry(cmdr, is_beta, system, station, entry, state):
             # and our flood limit isnt covered, queue an update
             should_poll = this.cargo != cargo and time.time() - this.last_update_time > FLOOD_LIMIT_SECONDS
             
-            if should_poll:
-                print("[{}] INARA: should_poll is true".format(time.asctime()))
-
             # Send event(s) to Inara
             if entry['event'] == 'ShutDown' or len(this.events) > old_events or should_poll:
 
                 # Send cargo and materials if changed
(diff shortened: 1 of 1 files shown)


A-UNDERSCORE-D:
Note re WIP: Will remove the commented out code and debug print before merge, its currently how it stands in my dev branch while I'm testing

plugins/inara.py
```diff
@@ -50,6 +50,9 @@
 this.fleet = None
 this.shipswap = False	# just swapped ship
 
+this.last_update_time = time.time()  # last time since we updated (set to now because we're going to update quickly)
```
Athanasius, line 53: That ' since' doesn't belong there.

A-UNDERSCORE-D: Whoops! thats from me changing stuff around. will fix shortly

A-UNDERSCORE-D:
After a day or two of testing it appears that this change does not do anything funky

PR #590, opened 2020-07-12
sort internal plugins before loading

### LINKED ISSUE #589: Sort plugins name for loading
Some people have plugins that depend on one another.  Currently they're at the mercy of the order returned by `os.listdir()` in plug.py -> load_plugins():

        for name in os.listdir(config.internal_plugin_dir):
There's no reason we can't wrap that in a `sort()` so they can, for now, enforce order with name prefixes.

Athanasius:
This is already done for 'found' plugins, so let's avoid any potnetial
future surprise with the internal ones as well.

close #589

### DIFF
--- a/plug.py
+++ b/plug.py
@@ -168,7 +168,7 @@ def load_plugins(master):
     last_error['root'] = master
 
     internal = []
-    for name in os.listdir(config.internal_plugin_dir):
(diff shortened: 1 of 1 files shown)


Athanasius:
Bah, wrong branch to merge to.

PR #596, opened 2020-07-13
Cleanup EDMC

A-UNDERSCORE-D:
Step 1 in getting this testable will be a cleanup, after that breaking out into functions that can be mocked etc. So far this just contains step 1


A-UNDERSCORE-D:
This needs a good look over as its been rebased a few times, but should be good. simple run confirms it works

PR #600, opened 2020-07-13
Fix/512 eddb system url

Athanasius:
I'm still not sure how https://forums.frontier.co.uk/threads/elite-dangerous-market-connector-edmc.548869/post-8589773 could have happened, but time to be more paranoid here.

Contains misc cleanups too.


Athanasius:
I'll fold this directly into the release-4.0.1.0 branch, which will then get merged into stable.

Issue #538, opened 2020-06-30
[SemVer] - Allow for pre-release versions


Athanasius:
e.g. 4.0.0-1.0.0 would be version 1.0.0 as a pre-release to 4.0.0 and would always be considered a lower version than plain 4.0.0

See point 9 on https://semver.org/#semantic-versioning-specification-semver

- [X] Verify that msiexec supports this. - *It doesn't, we'll cope.*
- [X] Verify that WiX supports this.  - *It doesn't, we'll cope.*
- [X] Verify WinSparkle update check still works. - *It does, with caveats, see below.*
- [X] Verify setup.py still works. - *We have to pass only Major.Minor.Patch.*
- [X] Verify update.py still works.
- [x] Display full version on "Help" > "About ...".

Athanasius:
And py2exe doesn't like SemVer:

```Building 'dist.win32\SemVerTest.exe'.
Traceback (most recent call last):
  File "C:\Users\<user>\AppData\Local\Programs\Python\Python37-32\lib\site-packages\py2exe\versioninfo.py", line 63, in __init__
    self.dwFileVersionLS = int(fields[2]) * 65536 + int(fields[3])
ValueError: invalid literal for int() with base 10: '0-beta1'```

Athanasius:
The solution for py2exe is to pass the full version in product_version, and only the major.minor.patch(.revision) in version.

This also has to be done for WiX to be happy, and I suspect misexec too.

The only way I see to get the 'installer' step to actually check -<prerelease> will be custom WiX actions, possibly involving scripting.

I'm going to leave that for now and check how WinSparkle reacts to the extra fields.

Athanasius:
WinSparkle 'by accident' supports -prerelease properly, but not +build (it considers if part of the version for comparisons).

Solution is to force WinSparkle to not even see the +build part:

```python
            appversion_nobuildmetadata = appversion.split(sep='+')[0]
            self.updater.win_sparkle_set_app_build_version(appversion_nobuildmetadata)```

Athanasius:
[X] - We naturally use the entire version string on 'Help' > ' About ...'

*Assuming* a proper tag was put on a beta or rc GitHub Release then the 'Release Notes' URL will also work.

Issue #539, opened 2020-06-30
[SemVer] Allow for appending git hash to version


Athanasius:
e.g. 4.0.0+a7b34c2 to denote the code is based on that git hashtag.

There might be room for this to be added automatically as part of the build process, if e.g. there is no `Release/A.B.C` tag for the current HEAD, so as to indicate this version deviates from the tag the release number implies.

NB: Will *not* cause the version to be considered different from tha plain A.B.C version.

See point 10 on https://semver.org/#semantic-versioning-specification-semver

- [X] Verify that msiexec supports this. - *It doesn't, we'll cope.*
- [X] Verify that WiX supports this.  - *It doesn't, we'll cope.*
- [X] Verify WinSparkle update check still works. - *It does, with caveats, see below.*
- [x] Verify setup.py still works.
- [x] Verify update.py still works.
- [x] Display full version on "Help" > "About ...".

Athanasius:
All of py2exe, WiX and misexec are *not* going to like anything beyond Major.Minor.Patch(.Revision) in the version numbers they handle.

Athanasius:
WinSparkle 'by accident' supports -prerelease properly, but not +build (it considers if part of the version for comparisons).

Solution is to force WinSparkle to not even see the +build part:

```python
            appversion_nobuildmetadata = appversion.split(sep='+')[0]
            self.updater.win_sparkle_set_app_build_version(appversion_nobuildmetadata)
```

Athanasius:
Closing this per #607 - but note we're not yet actually appending any such hash.

Issue #537, opened 2020-06-30
[SemVer] Change appversion to A.B.C form


Athanasius:
- [X] Verify that msiexec supports A.B.C -> https://docs.microsoft.com/en-gb/windows/win32/msi/version
- [x] Verify that WiX supports this.
- [x] Verify that WinSparkle properly compares versions still.
- [x] Verify setup.py works with the shorter version.
- [x] Verify that update.py all works.

Athanasius:
All of py2exe, WiX and misexec are happy with Major.Minor.Patch. The most it will cause is a padding `.0` being added on the end to the .exe's "File version".

PR #607, opened 2020-07-17
Semantic Version support: py2exe & WinSparkle

Athanasius:
* py2exe only supports Major.Minor.Patch(.Revision), so pass only
Major.Minor.Patch to it.  'File version' of generated EXEs will have
a `.0` appended as .Revision.

* Set 'product_version' in py2exe with the full version string.  This
ends up in the EXE 'Product version' property.

* WinSparkle accidentally suppoorts semver -prerelease correctly, but
not +buildmetadata.  So trim the latter off and set that as the version
for checks.


Athanasius:
This is addressing:

#537 - This should be working, double check if we release 4.0.3(.0) !
close #538 - "-prelease" works throughout now
#539 - Still need to ensure 'Release Notes' URL works with +build.

EDMC.py
```diff
@@ -62,21 +63,41 @@ def versioncmp(versionstring):
...
-            #pass	# Quietly suppress timeouts etc.
-        if latest:
-            print('{CURRENT} ({UPDATE} is available)'.format(CURRENT=appversion, UPDATE=latest))
+        except requests.RequestException as ex:
+            sys.stderr.write('Error retrieving update_feed file: {}\n'.format(str(ex)))
+        else:
+            try:
+                feed = ElementTree.fromstring(r.text)
+            except SyntaxError as ex:
+                sys.stderr.write('Syntax error in update_feed file: {}\n'.format(str(ex)))
```
A-UNDERSCORE-D, line 75:
```suggestion
                print('Syntax error in update_feed file: {}'.format(str(ex)), file=sys.stderr)
```

Print makes it cleaner as to what you're doing

EDMC.py
```diff
@@ -62,21 +63,41 @@ def versioncmp(versionstring):
...
-            lastversion = sorted(items, key=versioncmp)[-1]
-            if versioncmp(lastversion) > versioncmp(appversion):
-                latest = items[lastversion]
-        except Exception as e:
-            sys.stderr.write('Exception in version check: {}'.format(str(e)))
-            #pass	# Quietly suppress timeouts etc.
-        if latest:
-            print('{CURRENT} ({UPDATE} is available)'.format(CURRENT=appversion, UPDATE=latest))
+        except requests.RequestException as ex:
+            sys.stderr.write('Error retrieving update_feed file: {}\n'.format(str(ex)))
```
A-UNDERSCORE-D, line 70:
```suggestion
            print('Error retrieving update_feed file: {}'.format(str(ex)), file=sys.stderr)
```

EDMC.py
```diff
@@ -62,21 +63,41 @@ def versioncmp(versionstring):
...
+                            {'title': item.find('title').text,
+                             'os': item.find('enclosure').attrib.get('{http://www.andymatuschak.org/xml-namespaces/sparkle}os'),
+                            }
+                        ) for item in feed.findall('channel/item')
+                    ]
+                )
+
+                # Filter on the sparkle:os attribute
+                os_map = {'darwin': 'macos', 'win32': 'windows', 'linux': 'linux'}  # Map sys.platform to sparkle:os
+                items = { k:v for (k,v) in items.items() if v['os'] == os_map[sys.platform]}
```
A-UNDERSCORE-D, line 91:
```suggestion
                items = {k: v for (k, v) in items.items() if v['os'] == os_map[sys.platform]}
```

update.py
```diff
@@ -28,17 +26,47 @@ def checkForUpdates(self):
...
+            except requests.RequestException as ex:
+                sys.stderr.write('Error retrieving update_feed file: {}\n'.format(str(ex)))
+            else:
+                try:
+                    feed = ElementTree.fromstring(r.text)
+                except SyntaxError as ex:
+                    sys.stderr.write('Syntax error in update_feed file: {}\n'.format(str(ex)))
+                else:
+
+                    items = dict()
```
A-UNDERSCORE-D, line 45:
```suggestion
                    items = {}
```

We use literals elsewhere, so we should be consistent

EDMC.py
Athanasius: Changed in my local copy, although all of this will likely change as I'm trying to make one common function for this checking, which means no assumption of writing error messages being appropriate.

update.py
Athanasius: I think I prefer the more obvious (especially to any Python newbie) `dict()`.  Also PyCharm comes with a style check on by default to convert such literals *to* the constructor.  Not that PEP8 seems to mention it....

Yes, that would mean converting all the *others* to the constructor form.

A-UNDERSCORE-D: That works

Athanasius: On the other hand, reviewing all those PyCharm options now I don't think I want the verboseness of always using dict(...) form, so yeah, literal it is.

A-UNDERSCORE-D: makes it easier to expand if we ever do so as well

setup.py
```diff
@@ -150,26 +155,26 @@ def iterRecipes(module=recipes):
 setup(
     name = APPLONGNAME,
     version = VERSION,
-    app = [APP],
```
Athanasius, line 153: This and the setup_requires have caused warnings for a while.  Harmless, but also harmless to remove them.

update.py
```diff
@@ -1,58 +1,102 @@
 import os
 from os.path import dirname, join
 import sys
-from time import time
 import threading
 from traceback import print_exc
+import semantic_version
+from typing import Optional
+import tkinter as tk
```
Athanasius, line 8: Importing tkinter *purely* to reference the type .Tk

update.py
```diff
@@ -61,59 +105,103 @@ def __init__(self, master):
...
+        Handles the EDMarketConnector.AppWindow.onexit() request.
+
+        NB: We just 'pass' here because:
+         1) We might have a worker() going, but no way to make that
+            co-operative to respond to a "please stop now" message.
+         2) If we're running frozen then we're using (Win)Sparkle to check
+            and *it* might have asked this whole application to quit, in
+            which case we don't want to ask *it* to quit
+
+        :return: None
```
Athanasius, line 205: I was part way through putting code in to kill the worker and close down (Win)Sparkle when I realised why we shouldn't.  So, documenting it.  May as well be docstring instead of just a comment.

.flake8
```diff
@@ -0,0 +1,33 @@
+[flake8]
```
Athanasius, line 1: This is in develop already, wtf GitHub?  Here because I rebased *this* on develop after the .flake8 PR merge

docs/Releasing.md
```diff
@@ -67,20 +67,20 @@ Before you create a new install each time you should:
 Preparing to Package
 ===
 
-**Version numbers should always be referred to in full, e.g. A.B.C.D not the
-old A.BC scheme.**
+**Version numbers should always be referred to as, e.g. A.B.C not the
```
Athanasius, line 70: I should fully review this, link to documentation for Semantic Versions, etc.

.flake8
A-UNDERSCORE-D: I had this annoyance too. Fix is `rebase -i` and removing  the commit in the rebase (or it was for me)

update.py
A-UNDERSCORE-D: In this case a forward reference would be better I think (Simply because tk is a large import that we do nothing else with in this file). wrap the tkinter import in `if False` or `if typing.TYPE_CHECKING` and where you use it for types wrap them in strings.

Athanasius:
I think is is finally all done.  It'll need merging into main as well simply because of the documentation changes.  I don't want someone accidentally using non-SV strings because they read the old docs.

PR #613, opened 2020-07-19
cleanup monitor.py

A-UNDERSCORE-D:


A-UNDERSCORE-D:
This is ready ready for testing and any reviews people want to throw at it

monitor.py
```diff
@@ -52,9 +51,10 @@ class EDLogs(FileSystemEventHandler):
     _POLL = 1		# Polling is cheap, so do it often
     _RE_CANONICALISE = re.compile(r'\$(.+)_name;')
     _RE_CATEGORY = re.compile(r'\$MICRORESOURCE_CATEGORY_(.+);')
+    _RE_LOGFILE = re.compile(r'^Journal(Beta)?\.[0-9]{12}\.[0-9]{2}\.log$')
 
     def __init__(self):
-        FileSystemEventHandler.__init__(self)	# futureproofing - not need for current version of watchdog
+        FileSystemEventHandler.__init__(self)  # futureproofing - not need for current version of watchdog
```
Athanasius, line 57: I wonder how old that comment is and if it's no longer futureproofing....

monitor.py
```diff
@@ -71,7 +71,7 @@ def __init__(self):
         # If 3 we need to inject a special 'StartUp' event since consumers won't see the LoadGame event.
         self.live = False
 
-        self.game_was_running = False	# For generation the "ShutDown" event
+        self.game_was_running = False  # For generation the "ShutDown" event
```
Athanasius, line 74: generation **of** the

monitor.py
```diff
@@ -226,80 +250,102 @@ def worker(self):
...
+
             else:
-                self.event_queue.append(None)	# Generate null event to update the display (with possibly out-of-date info)
+                # Generate null event to update the display (with possibly out-of-date info)
+                self.event_queue.append(None)
                 self.live = False
 
         # Watchdog thread
-        emitter = self.observed and self.observer._emitter_for_watch[self.observed]	# Note: Uses undocumented attribute
+        emitter = self.observed and self.observer._emitter_for_watch[self.observed]  # Note: Uses undocumented attribute
```
Athanasius, line 272: I wonder if there's a documented method to achieve this now ?

monitor.py
```diff
@@ -355,54 +404,61 @@ def parse_entry(self, line):
...
-                    'Credits'      : entry['Credits'],
-                    'FID'          : entry.get('FID'),	# From 3.3
-                    'Horizons'     : entry['Horizons'],	# From 3.0
-                    'Loan'         : entry['Loan'],
-                    'Engineers'    : {},
-                    'Rank'         : {},
-                    'Reputation'   : {},
-                    'Statistics'   : {},
-                    'Role'         : None,
+                # Don't set Ship, ShipID etc since this will reflect Fighter or SRV if starting in those
```
Athanasius, line 407: You completely lost the alignment on this one :P

monitor.py
```diff
@@ -411,197 +467,258 @@ def parse_entry(self, line):
...
+
+            elif event_type == 'ModuleBuy':
                 self.state['Modules'][entry['Slot']] = {
-                    'Slot'     : entry['Slot'],
-                    'Item'     : self.canonicalise(entry['BuyItem']),
-                    'On'       : True,
-                    'Priority' : 1,
-                    'Health'   : 1.0,
-                    'Value'    : entry['BuyPrice'],
+                    'Slot': entry['Slot'],
```
Athanasius, line 478: Alignment again

monitor.py
A-UNDERSCORE-D: Yeah I have no idea. on linux FileSystemEventHandler is nonexistent, but given that its "not needed" Im inclined to remove it (or at least replace with `super().__init__()`

monitor.py
```diff
@@ -411,197 +467,258 @@ def parse_entry(self, line):
...
                 payload.pop('timestamp')
-                self.state[entry['event']] = payload
+                self.state[event_type] = payload
+
+            elif event_type == 'EngineerProgress':
+                engineers = self.state['Engineers']
+                if 'Engineers' in entry:  # Startup summary
+                    self.state['Engineers'] = {
+                        e['Engineer']: (e['Rank'], e.get('RankProgress', 0))
+                        if 'Rank' in e else e['Progress'] for e in entry['Engineers']
```
Athanasius, line 562: It feels like the conditional should be indented to emphasise it's attached to the previous line.

monitor.py
```diff
@@ -411,197 +467,258 @@ def parse_entry(self, line):
...
                 module['Engineering'] = {
-                    'Engineer'      : entry['Engineer'],
-                    'EngineerID'    : entry['EngineerID'],
-                    'BlueprintName' : entry['BlueprintName'],
-                    'BlueprintID'   : entry['BlueprintID'],
-                    'Level'         : entry['Level'],
-                    'Quality'       : entry['Quality'],
-                    'Modifiers'     : entry['Modifiers'],
-                    }
+                    'Engineer': entry['Engineer'],
```
Athanasius, line 656: Alignment

monitor.py
A-UNDERSCORE-D: Does not appear to be, went looking

monitor.py
A-UNDERSCORE-D: Oh whoops, I assumed that was part of the comprehension

monitor.py
Athanasius: It's a 'darwin' block of code, and is `watchdog.events.FileSystemEventHandler` which definitely does exist.  But, yeah, using `super().__init__()` is slightly more indicative of the intent.

monitor.py
Athanasius: Add a comment about self.observer.emitters being a Set, and could check for a member with a 'watch' attribute. But as this is working, leave it be for now, we'll address it if watchdog internals ever change.

PR #625, opened 2020-07-22
Switched to using a timer for inara updates

A-UNDERSCORE-D:
This adds a new thread that will run in a loop sending inara events once
ever 30 seconds. The old method of sending once for "special" events has
been removed.

**THIS IS UNTESTED** it needs testing before it is merged

### DIFF
--- a/plugins/inara.py
+++ b/plugins/inara.py
@@ -103,7 +103,8 @@ def plugin_app(parent):
 
 def plugin_stop():
     # Send any unsent events
-    call(force=True)
+    call()
+    time.sleep(0.1)  # Sleep for 100ms to allow call to go out, and to force a context switch to our other threads
     # Signal thread to close and wait for it
     this.queue.put(None)
     this.thread.join()
--- a/plugins/inara.py
+++ b/plugins/inara.py
@@ -459,10 +459,6 @@ def journal_entry(cmdr, is_beta, system, station, entry, state):
             logger.debug('Adding events', exc_info=e)
             return str(e)
 
-        #
-        # Events that don't need to be sent immediately but will be sent on the next mandatory event
-        #
-
         # Send credits and stats to Inara on startup only - otherwise may be out of date
         if entry['event'] == 'LoadGame':
             add_event('setCommanderCredits', entry['timestamp'],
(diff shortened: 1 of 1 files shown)


plugins/inara.py
```diff
@@ -411,23 +419,17 @@ def journal_entry(cmdr, is_beta, system, station, entry, state):
             # if our cargo differers from last we checked, we're at a station,
             # and our flood limit isnt covered, queue an update
```
Athanasius, line 420: differ**ER**s ?

isnt ?

A-UNDERSCORE-D: That code should have been removed anyway :D

A-UNDERSCORE-D:
Im gonna need to rebase this before merging

A-UNDERSCORE-D:
okay rebase done, should be mergeable

plugins/inara.py
```diff
@@ -96,12 +104,14 @@ def plugin_app(parent):
 
 def plugin_stop():
     # Send any unsent events
-    call()
+    call(force=True)
```
Athanasius, line 107: I thought we agreed that forcing on close might be bad, as the user *could* conceivably restart EDMC fast enough to then send something else inside the flood limit.

Better to delay/not send the events on close (they can log back in if needs be) than hit the 1-hour timeout because of this.

plugins/inara.py
```diff
@@ -189,7 +199,7 @@ def journal_entry(cmdr, is_beta, system, station, entry, state):
 
     # Send any unsent events when switching accounts
     if cmdr and cmdr != this.cmdr:
-        call()
+        call(force=True)
```
Athanasius, line 202: But, yes, this force is fine as Cmdr changes, and thus should be using a different Inara API key anyway, so won't cause issues.

In *theory* someone could be on CmdrA and have CmdrB in a different install sat in the launcher ready to go, but I still think they'd not be able to actually get in-game within 30s to bust the flood limit.  If they're using different Windows users for this then they need to do that also for EDMC anyway, so no issue.

A-UNDERSCORE-D:
Okay this should be good to merge I think.

PR #627, opened 2020-07-23
Fixed issues with super long links for shipyards

A-UNDERSCORE-D:
This works by creating a temp file at config.app_dir and storing the
link in there, followed by directing the local browser to open the file.

HTML meta tags are then used to direct the browser to refresh to a URL
of our choosing (which is HTML escaped, just in case someone tries
something clever)

This should work everywhere, and on any browser (as the file:// format
is defined at https://tools.ietf.org/html/rfc1738 which was posted in
1994).

The URI used (`file://localhost/path`) ensures that we only ever
ask for a localhost file at our path.

The HTML format should be completely compliant with all major browsers
as well, ensuring that behaviour is consistent (assuming they support
HTML meta tags)

The commented out windows specific code is left there for now, it should be removed before merging

### DIFF
--- a/EDMarketConnector.py
+++ b/EDMarketConnector.py
@@ -11,6 +11,7 @@ import json
 from os import chdir, environ
 from os.path import dirname, expanduser, isdir, join
 import re
+import html
 import requests
 from time import gmtime, time, localtime, strftime, strptime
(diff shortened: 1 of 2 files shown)


Athanasius:
Looks good, but now I need to work on this some more in a branch, so I'll close this and open a new PR from an EDCD branch.

Issue #617, opened 2020-07-20
Ship load out link error


deltaw77:
**Describe the bug**
when clicking on your ship to get the build (either Coriolis or EDSY) both gives an import error :

Coriolis:
Jameson, we have a problem..
Failed to import ship

This does not happen to all my ships just some, not sure why some work and some don't

**To Reproduce**
1. Click on ship name'
2.  See error

**Expected behavior**
Ship loadout should show in either Coriolis or EDSY.

**Screenshots**
![edmc ship loadout error](https://user-images.githubusercontent.com/68562108/87972606-7d6fc100-cac7-11ea-9a66-dfa41c6361dd.jpg)

**Desktop (please complete the following information):**
 - OS: Windows 7
 - If applicable: Browser Firefox
 - Version 4.0.2.0

Not sure if it is an EDMC or Coriolis or FD problem
![edmc ship loadout error](https://user-images.githubusercontent.com/68562108/87972851-ea835680-cac7-11ea-83a6-9130d1e82a84.jpg)

Athanasius:
Could you please copy and paste the full URL from the Firefox URL bar in each failing case ?  Do it between backticks

```
`like this`
```

Athanasius:
Example of bugged loadout event from journal

[edmc-bugged-ship-loadout-617.txt](https://github.com/EDCD/EDMarketConnector/files/4951703/edmc-bugged-ship-loadout-617.txt)

deltaw77:
[...]

Issue #616, opened 2020-07-20
EDMC Not Showing "Station" after Update


cmdrredhouse:
Hello

I installed the EDMC and everything works fine, except, now, after an update to EDMC, the "Station" no longer shows. It does show if I go into File, Status, and Ships... there it does show my current Station, but on the main screen... no Station info. When I first installed this EDMC, worked fine and showed all info, including Station. Then, like literally a day later, there was an update that took out the Station on the main interface screen. Then there was another update maybe two days later but still, no Station information.

I am using the latest version, have tried with ED running and not, have hit Update 10000000 times, went through every setting... no luck. Just stopped working after that update.

![EDMC](https://user-images.githubusercontent.com/68560854/87969779-9f415780-ca77-11ea-9296-457a19993522.png)

![EDMC](https://user-images.githubusercontent.com/68560854/87970473-a7e65d80-ca78-11ea-9d55-f857d4f8e61e.png)

Athanasius:
What do you have set for station provider?  'File' > 'Settings' > 'Configuration' ... 'Preferred websites' > 'Station'.

cmdrredhouse:
HI. I have INARA.

![7-20-2020 11-25-26 AM](https://user-images.githubusercontent.com/68560854/87972508-db76b700-ca7b-11ea-94ca-e7d27bb609cd.png)

cmdrredhouse:
Hmm, I changed it and now it works. 

![7-20-2020 11-28-51 AM](https://user-images.githubusercontent.com/68560854/87972777-55a73b80-ca7c-11ea-81f5-2e6fbdc8f403.png)

Athanasius:
Well, yes, different code is run depending on your selection there.

We did make a few tweaks to `plugins/inara.py` (All Inara, EDDB and EDSM stuff for System/Station is in the internal plugins).  I'll have a look into this tomorrow.

cmdrredhouse:
When I changed Station to EDDB, I could no longer click on the System link, was just a circle with a line, so that stopped working. However, I changed BOTH System and Station to EDSM and now it all works! :)

![7-20-2020 11-28-51 AM](https://user-images.githubusercontent.com/68560854/87973671-baaf6100-ca7d-11ea-975f-35caf6c54ed1.png)

cmdrredhouse:
Nope, sorry... just stopped working again. I give up.

[...]

PR #628, opened 2020-07-23
Fix/616 alternate url opening

Athanasius:
Renamed branch from @A-UNDERSCORE-D 's work.


Athanasius:
NB: Branch name is wrong, this addresses #617  not #616

Issue #563, opened 2020-07-06
[Enhancement] Add proper logging 


A-UNDERSCORE-D:
Currently all logging is done via print statements, which works fine for simple information but lacks time (and optionally file/line) information. Python3 provides a [very powerful logging engine](https://docs.python.org/3.7/library/logging.html) that would even allow plugins to hook in and augment their logging to fit our structure.

- [x] Replace most `__debug__` stuff with appropriate `logging` level statements.
- [ ] A_D : maybe add "THIS IS A BUG, PLEASE REPORT ON EDMC GITHUB", so if users ever see it, we can see it
- [ ] Consider some level of log file rotation, rather than losing all previous logging on a fresh run.

TODO:

- [x] - Check that 'found' third-party plugins *can* use this without issue.  There shouldn't be any problem, I just want to be paranoid about them loading from a different path not blowing anything up.
- [x] - Update non-wiki Plugins.md for how to use logging.
- [x] - Add both EDMC.py commandline and EDMarketConnector.py UI for changing logging level.  Really just toggling between INFO and DEBUG.
- [x] - Check method for detecting (in plugins) if logging is available:
    `l = logging.getLogger(); if l.hasHandlers: <something to make all logging functions just be print, or set up basic logger?>`
    This would then need to go in non-wiki Plugins.md in main/stable for plugin developers to see/check.

Athanasius:
You read my mind, I'd been musing on this change already :) .

A-UNDERSCORE-D:
Ill add it to the list :D

Athanasius:
You're possibly more familiar with the module than me.

I'm assuming it will, of course, just end up using print() anyway, and that must normally be going through some sort of mutex else threads would all print over and in the middle of each other, right ?

Basically, is it OK for each file/module to just `import logging` and get on with it, or should we make a global `logger` instance available, including through to plugins, and document that as the way to do it?  It would mean easy enforcement of the format/other logging options.

Our own edmclogging.py that is imported everywhere ?

A-UNDERSCORE-D:
Yeah its safe for multiple threads (at the very least by way of the GIL), there are two ways we can do that, either as you say, a global logger instance, or you can ask the logging lib for the instance by name: `logging.getlogger("name")`

[...]

PR #630, opened 2020-07-24
Enhancement/563 proper logging

Athanasius:
It will be easier to diagnose user issues with proper logging.  This branch adds that.

See #563 for further comments and thoughts.

### DIFF
--- a/EDMarketConnector.py
+++ b/EDMarketConnector.py
@@ -947,6 +947,7 @@ def enforce_single_instance() -> None:
 # Logging
 
 # Has to be here to be defined for other modules importing
+# TODO: now nothing should import this, can it move back into __main__ block ?
 logger = logging.getLogger(appname)
 
 class EDMCContextFilter(logging.Filter):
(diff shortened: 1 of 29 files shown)


EDMarketConnector.py
```diff
@@ -935,11 +943,141 @@ def enumwindowsproc(hWnd, lParam):
...
+
+        def stack_(frame):
+            framelist = []
+            while frame:
+                framelist.append(frame)
+                frame = frame.f_back
+            return framelist
+
+        stack = stack_(sys._getframe(1))
+        # TODO: Make this less fragile by not depending on a magic number of
```
Athanasius, line 986: See, I'm aware we don't want to rely on a magic number :)

Athanasius:
This is ready for review.  Rather than take a week or more to complete let's get the logging in place and available for use in other cleanup work.

Athanasius:
NB: flake8 fails more and more lines on each push due to the diffs expanding and catching more lines.

So, leaving all of that for now.  The actual code changes have been tested.

PR #632, opened 2020-07-26
Cleanup eddb plugin

A-UNDERSCORE-D:
Usual string of PEP8 and readability improvements

### DIFF
--- a/plugins/eddb.py
+++ b/plugins/eddb.py
@@ -47,8 +47,10 @@ this.station_marketid = None
 def system_url(system_name: str) -> str:
     if this.system_address:
         return requests.utils.requote_uri(f'https://eddb.io/system/ed-address/{this.system_address}')
+
     elif system_name:
         return requests.utils.requote_uri(f'https://eddb.io/system/name/{system_name}')
+
     else:
         return ''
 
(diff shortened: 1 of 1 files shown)


plugins/eddb.py
```diff
@@ -24,55 +24,64 @@
...
-this.system_address = None
-this.system_population = None
-this.station_link = None
-this.station = None
-this.station_marketid = None
+this.system_link: Optional[str] = None
+this.system: Optional[str] = None
+this.system_address: Optional[str] = None
+this.system_population: Optional[int] = None
+this.station_link = None  # tk thing, not annotated
```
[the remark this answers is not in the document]
A-UNDERSCORE-D: was not sure what it was and was half asleep enough to not want to go digging to figure it out. I'll poke at that today

A-UNDERSCORE-D:
This needs a rebase

plugins/eddb.py
```diff
@@ -24,55 +24,68 @@
 
 
 import sys
+import tkinter
```
Athanasius, line 27: Is this necessary given the conditional 'from' import below ?

A-UNDERSCORE-D: removed -- was added in an auto import while I was typing

PR #636, opened 2020-07-27
Continuing work on passing flake8 in EDMarketConnector.py

Athanasius:


Athanasius:
Just the 'too complex' which need a greater refactor to resolve, along with the full addition of type annotations still needed.

PR #644, opened 2020-07-28
Enhancement/remove keyring

A-UNDERSCORE-D:
Removed keyring dependency
    
This removes all dependencies on the keyring lib, updates the requirements.txt to reflect that, and ensures that setup.py does not
attempt to package it.
    
Any use of the "old" keyring code will now return None and warn about its deprecation.

### DIFF
--- a/config.py
+++ b/config.py
@@ -1,5 +1,6 @@
 import numbers
 import sys
+import warnings
 from os import getenv, makedirs, mkdir, pardir
 from os.path import expanduser, dirname, exists, isdir, join, normpath
 from sys import platform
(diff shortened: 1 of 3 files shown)


Athanasius:
I wonder if any plugin is making use of it.  We package it, so they could.

A-UNDERSCORE-D:
That is true, maybe continue to package it? or package a wrapper that just exports all of its stuff, but throws up a warning?

Athanasius:
I checked all the "works with Python 3.x" plugins from the wiki page, and none of them have the string 'keyring' in any file.

So feel free to remove it.  Anyone who it breaks it for can just package keyring in their plugin anyway.

A-UNDERSCORE-D:
that being the case, merge away :D

PR #647, opened 2020-07-28
Try to only run flake8 annotate on Athanasius' PRs

Athanasius:
It only works with sufficient access anyway.


Athanasius:
OK, I'm done with this.

The use? Well, anyone else reviewing my PRs can benefit from the in-PR annotations without having to run flake8 themselves.

Athanasius:
The real fix is some code to check if an annotation can be added. If so remove it again (let's more permission required is equivalent) and exit cleanly, if not signal that and exit cleanly.

PR #648, opened 2020-07-28
Cleanup inara.py

A-UNDERSCORE-D:
Standard pep8 treatment et al.

Still considering replacing all OrderedDicts with simple dicts.

### DIFF
--- a/plugins/inara.py
+++ b/plugins/inara.py
@@ -1206,7 +1206,6 @@ def worker():
                 plug.show_error(_("Error: Can't connect to Inara"))
 
 
-# TODO(A_D) event is unused
 def update_location(event=None):
     """
     Call inara_notify_location in this and other interested plugins with Inara's response when changing system
@@ -1223,7 +1222,6 @@ def inara_notify_location(eventData):
     pass
 
 
-# TODO(A_D) event is unused
 def update_ship(event=None):
     """
     Call inara_notify_ship() in interested plugins with Inara's response when changing ship
--- a/plugins/inara.py
+++ b/plugins/inara.py
@@ -958,8 +958,8 @@ def journal_entry(cmdr: str, is_beta: bool, system: str, station: str, entry: Di
 
     if config.get('station_provider') == 'Inara':
         to_set = this.station
-        if not to_set and this.system_population is not None:
-            if this.system_population > 0:
+        if not to_set:
+            if this.system_population is not None and this.system_population > 0:
                 to_set = STATION_UNDOCKED
             else:
                 to_set = ''
--- a/plugins/inara.py
+++ b/plugins/inara.py
@@ -13,7 +13,6 @@ from operator import itemgetter
 from queue import Queue
 from threading import Thread
 import logging
-import dataclasses
 
 import tkinter as tk
 from ttkHyperlinkLabel import HyperlinkLabel
@@ -145,7 +144,7 @@ def plugin_prefs(parent: tk.Tk, cmdr: str, is_beta: bool):
     this.log_button.grid(columnspan=2, padx=BUTTONX, pady=(5, 0), sticky=tk.W)
 
     nb.Label(frame).grid(sticky=tk.W)  # big spacer
-    
+
     # Section heading in settings
     this.label = HyperlinkLabel(
         frame,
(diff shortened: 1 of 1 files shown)


plugins/inara.py
```diff
@@ -21,38 +22,41 @@
...
-this.materials = None
-this.lastcredits = 0	# Send credit update soon after Startup / new game
-this.storedmodules = None
-this.loadout = None
-this.fleet = None
-this.shipswap = False	# just swapped ship
+this.events: List[OrderedDictT[str, Any]] = []  # Unsent events
+this.cmdr: Optional[str] = None
+this.FID: Optional[str] = None		# Frontier ID
+this.multicrew = False  # don't send captain's ship info to Inara while on a crew
```
Athanasius, line 45: type hints on bools are optional ?

plugins/inara.py
```diff
@@ -112,25 +123,40 @@ def plugin_stop():
...
+        url='https://inara.cz/settings-api',
+        underline=True
+    )
 
-    nb.Label(frame).grid(sticky=tk.W)	# big spacer
-    this.label = HyperlinkLabel(frame, text=_('Inara credentials'), background=nb.Label().cget('background'), url='https://inara.cz/settings-api', underline=True)	# Section heading in settings
     this.label.grid(columnspan=2, padx=PADX, sticky=tk.W)
 
-    this.apikey_label = nb.Label(frame, text=_('API Key'))	# EDSM setting
+    this.apikey_label = nb.Label(frame, text=_('API Key'))  # EDSM setting
```
Athanasius, line 159: This is inara.py, not edsm.py.  EDSM -> Inara in comment

plugins/inara.py
```diff
@@ -584,11 +726,12 @@ def journal_entry(cmdr, is_beta, system, station, entry, state):
             if this.storedmodules != modules:
                 # Only send on change
                 this.storedmodules = modules
-                this.events = [x for x in this.events if x['eventName'] != 'setCommanderStorageModules']	# Remove any unsent
+                this.events = [x for x in this.events if x['eventName']
+                               != 'setCommanderStorageModules']  # Remove any unsent
```
Athanasius, line 730: Breaking the line mid conditional feels *so* wrong.

1) Move the comment to its own line above?
2) Split at the 'if ' otherwise/if you still need to ?

plugins/inara.py
```diff
@@ -614,97 +758,134 @@ def journal_entry(cmdr, is_beta, system, station, entry, state):
...
+                ])
+            )
 
         # Community Goals
-        if entry['event'] == 'CommunityGoal':
-            this.events = [x for x in this.events if x['eventName'] not in ['setCommunityGoal', 'setCommanderCommunityGoalProgress']]	# Remove any unsent
-            for goal in entry['CurrentGoals']:
+        if event_name == 'CommunityGoal':
+            this.events = [x for x in this.events if x['eventName'] not in [
+                'setCommunityGoal', 'setCommanderCommunityGoalProgress']]  # Remove any unsent
```
Athanasius, line 886: Split conditional again.  Heck, split mid array in the worst way !

plugins/inara.py
```diff
@@ -757,30 +954,42 @@ def journal_entry(cmdr, is_beta, system, station, entry, state):
         this.system_link.update_idletasks()
 
     if config.get('station_provider') == 'Inara':
-        this.station_link['text'] = this.station or (this.system_population and this.system_population > 0 and STATION_UNDOCKED or '')
+        to_set = this.station
+        if not to_set and this.system_population is not None:
+            if this.system_population > 0:
+                to_set = STATION_UNDOCKED
+            else:
+                to_set = ''
```
Athanasius, line 962: This doesn't look equivalent.  Only the `this.system_population > 0` should be gated by the `this.system_population is not None`.  The latter is in case we plain don't have that variable set yet.  It shouldn't be preventing the else doing `to_set = ''`

plugins/inara.py
```diff
@@ -757,30 +954,42 @@ def journal_entry(cmdr, is_beta, system, station, entry, state):
...
 
     # Always store initially, even if we're not the *current* system provider.
     if not this.station_marketid:
         this.station_marketid = data['commander']['docked'] and data['lastStarport']['id']
+
     # Only trust CAPI if these aren't yet set
-    this.system = this.system or data['lastSystem']['name']
-    this.station = this.station or data['commander']['docked'] and data['lastStarport']['name']
+    this.system = this.system if this.system else data['lastSystem']['name']
+    this.station = data['lastStarport']['name'] if data['commander']['docked'] and not this.station else this.station
```
Athanasius, line 978: `this.station = this.station if this.station elif data['commander']['docked'] data['lastStarport']['name']`
Or is elif not allowed on such embedded conditionals ?

Basically it should be more obvious that `this.station` will stay as that if it's set.

Heck, just make it multi-line:

```
if not this.station and data['commander']['docked']:
    this.station = data['lastStarport']['name']
```
If this.station isn't None then we leave it be.

plugins/inara.py
A-UNDERSCORE-D: They can be inferred more easily by a type checker

plugins/inara.py
```diff
@@ -846,26 +1070,52 @@ def make_loadout(state):
...
+    """
+
     this.events.append(OrderedDict([
         ('eventName', name),
         ('eventTimestamp', timestamp),
         ('eventData', data),
     ]))
 
+
 def call_timer(wait=FLOOD_LIMIT_SECONDS):
```
Athanasius, line 1090: type hint of `int` ?

plugins/inara.py
```diff
@@ -918,40 +1179,57 @@ def worker():
...
         else:
             if callback:
                 callback(None)
+
             else:
                 plug.show_error(_("Error: Can't connect to Inara"))
 
 
-# Call inara_notify_location() in this and other interested plugins with Inara's response when changing system or station
+# TODO(A_D) event is unused
```
Athanasius, line 1209: It's a Tk event I believe, so having the parameter is mandatory.

plugins/inara.py
A-UNDERSCORE-D: oh I didnt commit this, yeah I had done that. I blame auto-formatters

plugins/inara.py
A-UNDERSCORE-D: if its mandatory it can be replaced with _, so not to imply its useful

plugins/inara.py
A-UNDERSCORE-D: Whoops. yeah. maybe:
```py
if not to_set:
    if this.system_population is not None and this.system_population > 0:
        to_set = STATION_UNDOCKED
    else:
        to_set = ''
```

plugins/inara.py
A-UNDERSCORE-D: elif is not allowed in conditionals like that no -- will do as suggested

plugins/inara.py
A-UNDERSCORE-D: either way Ill remove the comment

PR #650, opened 2020-07-30
Ensure 'found' plugins have correct logging from core code

Athanasius:
1. Log messages propagate up Parent.Child chains, so we don't need a channel on the plugin logger.
1. But it still needs the filter to define qualname and class for formatting.

### DIFF
--- a/EDMCLogging.py
+++ b/EDMCLogging.py
@@ -241,7 +241,7 @@ class EDMCContextFilter(logging.Filter):
         return frame
 
     @classmethod
-    def munge_module_name(cls, frame_info: inspect.FrameInfo, module_name: str) -> str:
+    def munge_module_name(cls, frame_info: inspect.Traceback, module_name: str) -> str:
         """
         Adjust module_name based on the file path for the given frame.
 
--- a/EDMCLogging.py
+++ b/EDMCLogging.py
@@ -35,6 +35,8 @@ from typing import Tuple
 #
 #      14. Call from *package*
 
+_default_loglevel = logging.DEBUG
+
 
 class Logger:
     """
@@ -47,7 +49,7 @@ class Logger:
     Users of this class should then call getLogger() to get the
     logging.Logger instance.
     """
-    def __init__(self, logger_name: str, loglevel: int = logging.DEBUG):
+    def __init__(self, logger_name: str, loglevel: int = _default_loglevel):
         """
         Set up a `logging.Logger` with our preferred configuration.
         This includes using an EDMCContextFilter to add 'class' and 'qualname'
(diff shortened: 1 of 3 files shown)


Athanasius:
The internal inara plugin currently gives:

1. `2020-07-30 15:36:34.467 - EDMarketConnector - INFO - inara.call:871: <none>: queuing upload of 1 events`

Which of these would we prefer?

2. `2020-07-30 15:36:34.467 - EDMarketConnector.inara - INFO - inara.call:871: <none>: queuing upload of 1 events`

3. `2020-07-30 15:36:34.467 - EDMarketConnector.inara - INFO - call:871: <none>: queuing upload of 1 events`

4. `2020-07-30 15:36:34.467 - EDMarketConnector - INFO - plugins.inara.call:871: <none>: queuing upload of 1 events`

I don't like option 3, 'call:871' is plain misleading.

I think I favour option 4.

Athanasius:
Oh, and ignore the flake8 on this, I only did a little adding of type hints in bits I changed, so plug.py is still a mess.

A-UNDERSCORE-D:
4 Provides the most amount of info with the least amount of redundant info, but dropping the EDMarketConnector part entirely may not be a bad idea, its a "given" based on the fact that thats what we're executing.

EDMCLogging.py
```diff
@@ -78,6 +80,33 @@ def get_logger(self) -> logging.Logger:
...
 
+def get_plugin_logger(name: str, loglevel: int = _default_loglevel) -> logging.Logger:
+    """
+    'Found' plugins need their own logger to call out where the logging is
+    coming from, but we don't need to set up *everything* for them.
+
+    The name will be '{config.appname}.{plugin.name}', e.g.
+    'EDMarketConnector.miggytest'.  This means that any logging sent through
+    there *also* goes to the channels defined in the 'EDMarketConnector'
+    logger, so we can let that take care of the formatting.
```
A-UNDERSCORE-D, line 91: Will be or should be? We dont set the loggers name here, only use what we're given

Athanasius: We set it in plug.py.  Either it's EDMarketConnector.<something> from that else it's bare EDMarketConnector due to being nothing to do with plugins.  The docs also now say to achieve the former *now* in plugin code.

The point is that we don't need to add a channel there because the message will propagate upwards to the EDMarketConnector logger which has one.

If a plugin author sets up a Logger with a name NOT of that format they're to blame when it doesn't log how we expect.

A-UNDERSCORE-D: Ah okay. Was mostly looking for the commentary of the name format being described here when it isnt set or enforced here. But if you're okay with "go against this at your own peril" that works for me

Athanasius:
This *should* be good to go now.  NB: flake8-your-pr is only complaining about plug.py stuff.

PR #658, opened 2020-08-02
Cleanup eddn plugin

A-UNDERSCORE-D:


plugins/eddn.py
```diff
@@ -35,121 +46,147 @@
...
+        self.parent: tk.Tk = parent
         self.session = requests.Session()
-        self.replayfile = None	# For delayed messages
-        self.replaylog = []
+        self.replayfile: Optional[TextIO] = None  # For delayed messages
+        self.replaylog: List[str] = []
 
-    def load(self):
+    def load(self) -> bool:
         # Try to obtain exclusive access to the journal cache
```
Athanasius, line 76: Is "journal cache" correct?  I'd assume this will cache *all* EDDN messages, and some of that data is from CAPI.

plugins/eddn.py
```diff
@@ -35,121 +46,147 @@
...
+
         self.replayfile = None
 
-    def send(self, cmdr, msg):
-        uploaderID = cmdr
+    def send(self, cmdr: str, msg: Mapping[str, Any]):
+        uploader_id = cmdr
 
-        msg = OrderedDict([
+        to_send: OrderedDictT[str, str] = OrderedDict([
```
Athanasius, line 119: Is this a case where OrderedDict() is still necessary ?

plugins/eddn.py
```diff
@@ -35,121 +46,147 @@
...
             logger.debug(f'\n{self.replaylog[0]}\n', exc_info=e)
             # Discard and continue
             self.replaylog.pop(0)
+
         else:
             # Rewrite old schema name
             if msg['$schemaRef'].startswith('http://schemas.elite-markets.net/eddn/'):
-                msg['$schemaRef'] = 'https://eddn.edcd.io/schemas/' + msg['$schemaRef'][38:]
+                msg['$schemaRef'] = 'https://eddn.edcd.io/schemas/' + msg['$schemaRef'][38:]  # TODO: 38?!
+
```
Athanasius, line 167: Indeed, replace with something actually referencing the string used in the conditional.

Although... we can either lost this code entirely *or* keep it as a reference on how to migrate older schema URLs.  It seems unlikely it would change *away* from eddn.edcd.io at this point though.

plugins/eddn.py
```diff
@@ -35,121 +46,147 @@
 this.marketId = None
 this.commodities = this.outfitting = this.shipyard = None
 
+HORIZ_SKU = 'ELITE_HORIZONS_V_PLANETARY_LANDINGS'
+
 
-class EDDN(object):
+# TODO:
+# 1, this does a bunch of magic checks for whether or not the given commander has horizons. We can get that directly
```
Athanasius, line 51: I wouldn't just assume that the CAPI indication can be 100% relied upon.  Heck I see reports about the *game* sometimes being confused as to if a Cmdr has Horizons or not.  So I'd not change of these checks unless they're provably broken.

plugins/eddn.py
```diff
@@ -207,20 +268,30 @@ def export_outfitting(self, data, is_beta):
...
-        horizons = (any(economy['name'] == 'Colony' for economy in economies.values()) or
-                    any(module.get('sku') == 'ELITE_HORIZONS_V_PLANETARY_LANDINGS' for module in modules.values()) or
-                    any(ship.get('sku') == 'ELITE_HORIZONS_V_PLANETARY_LANDINGS' for ship in list((ships['shipyard_list'] or {}).values())))
-        shipyard = sorted([ship['name'].lower() for ship in list((ships['shipyard_list'] or {}).values()) + ships['unavailable_list']])
-        if shipyard and this.shipyard != (horizons, shipyard):	# Don't send empty ships list - shipyard data is only guaranteed present if user has visited the shipyard.
+    def export_shipyard(self, data: Dict[str, Any], is_beta: bool):
+        economies: Dict[str, Any] = data['lastStarport'].get('economies') or {}
+        modules: Dict[str, Any] = data['lastStarport'].get('modules') or {}
+        ships: Dict[str, Any] = data['lastStarport'].get('ships') or {'shipyard_list': {}, 'unavailable_list': []}
+        horizons: bool = (
```
Athanasius, line 276: This is copy+pasted code.  Refactor into a helper function ?

plugins/eddn.py
A-UNDERSCORE-D: It appears to be. The code path the `cmdr_data` callback takes does not touch any of the replaylog stuff

plugins/eddn.py
```diff
@@ -325,96 +408,159 @@ def canonicalise(self, item):
...
+            'FuelUsed',
+            'JumpDist',
+            'Latitude',
+            'Longitude',
+            'Wanted'
+        ):
             entry.pop(thing, None)
+
         if 'Factions' in entry:
             # Filter faction state. `entry` is a shallow copy so replace 'Factions' value rather than modify in-place.
```
Athanasius, line 555: Mention in the comment that this is to comply with schema restrictions not allowing personal data over EDDN.

plugins/eddn.py
```diff
@@ -424,51 +570,64 @@ def filter_localised(d):
         # add mandatory StarSystem, StarPos and SystemAddress properties to Scan events
         if 'StarSystem' not in entry:
             if not system:
-                return("system is None, can't add StarSystem")
+                return "system is None, can't add StarSystem"
```
Athanasius, line 573: Add logger.debug(...) for these cases ?

plugins/eddn.py
```diff
@@ -424,51 +570,64 @@ def filter_localised(d):
...
+            entry['event'] in ('Market', 'Outfitting', 'Shipyard')):
+
         try:
             if this.marketId != entry['MarketID']:
                 this.commodities = this.outfitting = this.shipyard = None
                 this.marketId = entry['MarketID']
 
-            with open(join(config.get('journaldir') or config.default_journal_dir, '%s.json' % entry['event']), 'rb') as h:
+            with open(
+                join(str(config.get('journaldir') or config.default_journal_dir), f'{entry["event"]}.json'), 'rb'
```
Athanasius, line 609: I'm surprised you're not pathlib'ing things like this.

A-UNDERSCORE-D: was trying not to make that sort of change, but if Im good to do that I very much will

plugins/eddn.py
A-UNDERSCORE-D: In most of the code almost none of it is necessary. For the most part in cleanups Ive been leaving them just in case we end up somewhere thats doing an isinstance. In #660 I pretty much just replaced a bunch as I was controlling the code path there

plugins/eddn.py
A-UNDERSCORE-D: Do we want to change behavior there? that return ends up in [EDMarketConnector.py](https://github.com/EDCD/EDMarketConnector/blob/develop/EDMarketConnector.py#L671-L680) and does some sound silliness

Athanasius: Keep the return.  My thinking is that this looks like a "should never happen" thing so needs logging.  The 'sound silliness' alerts the user who might then actually send us logs.

A-UNDERSCORE-D: possibly a warn then over a debug?

Athanasius: Sure.

PR #659, opened 2020-08-03
UNTESTED: Fix inara timer (part 2)

A-UNDERSCORE-D:
This adds an additional wait to the worker thread of the inara code. As the API key used and all other pertinent data is added to the queue in call() we can afford here to wait a while between sends without worrying about going over limits.

I left the original timer fix in place, as it has the added benefit of ensuring that 30s worth of events are collected _before_ they are queued. Without that the worker thread would probably end up being overwhelmed with a bunch of single events.

This also breaks the sending logic into multiple functions for clarity.

### DIFF
--- a/plugins/inara.py
+++ b/plugins/inara.py
@@ -59,7 +59,9 @@ this.shipswap: bool = False  # just swapped ship
 
 # last time we updated, if unset in config this is 0, which means an instant update
 LAST_UPDATE_CONF_KEY = 'inara_last_update'
-FLOOD_LIMIT_SECONDS = 30  # minimum time between sending events
+EVENT_COLLECT_TIME = 31  # Minimum time to take collecting events before requesting a send
+WORKER_WAIT_TIME = 35  # Minimum time for worker to wait between sends
+
 this.timer_run = True
 
 
@@ -1097,7 +1099,7 @@ def add_event(name: str, timestamp: str, data: Union[EVENT_DATA, Sequence[EVENT_
     ]))
 
 
-def call_timer(wait: int = FLOOD_LIMIT_SECONDS):
+def call_timer(wait: int = EVENT_COLLECT_TIME):
     """
     call_timer runs in its own thread polling out to INARA once every FLOOD_LIMIT_SECONDS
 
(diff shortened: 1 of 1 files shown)


A-UNDERSCORE-D:
It may be prudent to have the worker timer do some more logic to wait exactly 30s, taking wait time etc into account, but I really dont think thats needed as yet.

plugins/inara.py
```diff
@@ -1141,74 +1142,84 @@ def call(callback=None, force=False):
...
+        try_send_data(url, data)
+        time.sleep(FLOOD_LIMIT_SECONDS)
 
+
+def try_send_data(url: str, data: OrderedDictT[str, Any]):
+    for i in range(3):
+        logger.debug(f"sending data to API, retry #{i}")
+        try:
+            if send_data(url, data):
                 break
```
A-UNDERSCORE-D, line 1169: This should probably be a return

plugins/inara.py
```diff
@@ -1141,74 +1142,84 @@ def call(callback=None, force=False):
...
-                            this.lastship = reply_event.get('eventData', {})
-                            # calls update_ship in main thread
-                            this.system_link.event_generate('<<InaraShip>>', when="tail")
+        url, data, _ = to_send
+        try_send_data(url, data)
+        time.sleep(FLOOD_LIMIT_SECONDS)
 
+
+def try_send_data(url: str, data: OrderedDictT[str, Any]):
+    for i in range(3):
```
A-UNDERSCORE-D, line 1165: if the below-mentioned return is added, this can log at the end of the loop that sending failed etc.

Athanasius:
I also note that we were specifically told:

> Artie (Inara) 23/07/2020 : 2 requests per minute are perfectly fine and it's shouldn't cause any problem when it will differ by a few seconds

30 seconds technically allows for 3 requests per minute, at :00, :30, :60 (i.e. next minute).  If there's any wiggle in the time checks that could go over.  So for one thing let's make it 35 seconds.  It won't make much difference to the user, but makes sure we definitely can't try to send more than twice a minute.

plugins/inara.py
Athanasius: Agreed

A-UNDERSCORE-D:
that sounds good to me, will change

A-UNDERSCORE-D:
Per discussion in discord, this may need closing and replacing with a rewritten inara sending system

Athanasius:
This is moot now given https://github.com/EDCD/EDMarketConnector/pull/660 , yes ?

A-UNDERSCORE-D:
Correct, I apparently forgot to close this

PR #660, opened 2020-08-03
Rewrote inara queue system

A-UNDERSCORE-D:
This replaces the list+queue system that the inara plugin originally
used with a deque based one.

The main differences here are that the list the worker thread uses to
send to inara and the list that events are added to is the same, with
the worker thread making a duplicate and clearing the original each time
it sends events (losing events if it fails to upload three times).

The format of the data has changed as well, from simple tuples to
NamedTuple classes that provide some extra type safety and sanity when
accessing fields.

The event queue itself is actually multiple queues, one per
API/FID/CMDR_name triplicate, thus allowing multiple commander switches
while we're running without causing any weird issues

### DIFF
--- a/plugins/inara.py
+++ b/plugins/inara.py
@@ -2,17 +2,18 @@
 # Inara sync
 #
 
-from collections import OrderedDict
+from collections import OrderedDict, defaultdict
 import json
-from typing import Any, AnyStr, Dict, List, Mapping, Optional, OrderedDict as OrderedDictT, \
-    Sequence, TYPE_CHECKING, Union
+from typing import Any, AnyStr, Callable, Deque, Dict, List, Mapping, NamedTuple, Optional, OrderedDict as OrderedDictT, \
+    Sequence, TYPE_CHECKING, Tuple, Union
+import dataclasses
 
 import requests
 import sys
 import time
 from operator import itemgetter
 from queue import Queue
-from threading import Thread
+from threading import Lock, Thread
 import logging
 
 import tkinter as tk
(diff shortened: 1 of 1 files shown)


A-UNDERSCORE-D:
Okay Ive tested this though a playthough, no explosions other than the above fixed commit. ready for review (not merge yet though)

plugins/inara.py
```diff
@@ -74,6 +83,50 @@ def _(x):
 STATION_UNDOCKED: str = '×'  # "Station" name to display when not docked = U+00D7
 
 
+class Credentials(NamedTuple):
+    """
+    Credentials holds an inara API payload
```
Athanasius, line 88: That docstring doesn't seem strictly true.  This is the identifier for a queue that *holds* payloads.

plugins/inara.py
```diff
@@ -1078,137 +1128,146 @@ def make_loadout(state: Dict[str, Any]) -> OrderedDictT[str, Any]:
...
     """
-    worker is the main thread worker and backbone of the plugin.
+    attempt repeatedly to send the payload forward
 
-    As events are added to `this.queue`, the worker thread will push them to the API
+    :param url: target URL for the payload
+    :param data: the payload
     """
+    for i in range(3):
+        logger.debug(f"sending data to API, retry #{i}")
```
Athanasius, line 1214: Technically an attempt, not a retry ?

A-UNDERSCORE-D:
This looks to be good. Tested over a few hours and days

PR #672, opened 2020-08-25
Cleanup EDSM plugin

A-UNDERSCORE-D:


plugins/edsm.py
```diff
@@ -12,135 +12,191 @@
...
+        underline=True
+    )
 
-    nb.Label(frame).grid(sticky=tk.W)	# big spacer
-    this.label = HyperlinkLabel(frame, text=_('Elite Dangerous Star Map credentials'), background=nb.Label().cget('background'), url='https://www.edsm.net/settings/api', underline=True)	# Section heading in settings
     this.label.grid(columnspan=2, padx=PADX, sticky=tk.W)
 
-    this.cmdr_label = nb.Label(frame, text=_('Cmdr'))	# Main window
+    this.cmdr_label = nb.Label(frame, text=_('Cmdr'))  # Main window
     this.cmdr_label.grid(row=10, padx=PADX, sticky=tk.W)
```
Athanasius, line 190: Rather than magic row numbers change the code to have a row variable incremented after each row is added.  It avoids issues with forgetting to change row numbers if you insert or delete something.

plugins/edsm.py
```diff
@@ -389,50 +519,58 @@ def worker():
...
     if this.navbeaconscan:
         if entries and entries[-1]['event'] == 'Scan':
             this.navbeaconscan -= 1
             if this.navbeaconscan:
                 return False
+
         else:
             assert(False)
-            this.navbeaconscan = 0
+            this.navbeaconscan = 0  # TODO: Unreachable code?
```
Athanasius, line 532: Need to review the intent of this code and ensure it's actually fulfilled.  Either this is a brainfart and the lines need swapping, or the `assert(False)` was a leftover from development of this code and needs removing.

Athanasius:
This needs re-basing on now-`develop` for the recent plugin changes.

plugins/edsm.py
A-UNDERSCORE-D: Even if intentional that should `raise ValueError` rather than `assert False`, because as discussed asserts are stripped in -O3, and an assert False without even a message attached does nothing to explain why it exploded

Athanasius: I *think* the intent is that the code in the if part of the conditional this is the else of *should* count the Scan events from a NavBeaconScan correctly, decrementing as it finds them.  *That* should always be what reduces this.navbeaconscan from the incoming scan count to zero.  So, if this.navbeaconscan was still true (> 0) but the next event isn't a Scan event, that's an error.  It's assuming the game always stictly follows the NavBeaconScan event with all the Scan events, nothing inbetween I think?

So, yeah, the `assert(False)` was for debugging, to ensure things fell over if the accounting didn't work.  What it should be replaced by is a `logger.error(...)`.  As for the `this.navbeaconscan = 0`, that should be moot as the only way to hit this code is if it was zero anyway (given the value is initialised so is never None, and nowhere is setting it False).

I do wonder about the logic of testing only the current last member of the passed `entries`.  Because if it's more than one Scan event it should be reducing this.navbeaconscan by more than one, and if there are other types of events mixed in it breaks the basic assumption.

The fact we don't have reports of the program crashing all over on Nav Beacon scans probably means the assumptions are correct and that technically `entries[-1]` is just devolving into `entries[0]`.

A-UNDERSCORE-D: Replaced with a log.error call. Left the `this.navbeaconscan` assignment as its a handy "just in case"

plugins/edsm.py
```diff
@@ -331,87 +471,116 @@ def cmdr_data(data, is_beta):
 
 
 # Worker thread
-def worker():
+def worker() -> None:
+    """
+    Upload worker.
 
-    pending = []	# Unsent events
+    Works based the `this.queue` queue, loops infinitely until the item returned from the queue is None.
```
Athanasius, line 478: Processes `this.queue`, looping until the item is None.

plugins/edsm.py
```diff
@@ -331,87 +471,116 @@ def cmdr_data(data, is_beta):
...
                     pending.append(entry)
 
                 # Get list of events to discard
                 if not this.discardedEvents:
                     r = this.session.get('https://www.edsm.net/api-journal-v1/discard', timeout=_TIMEOUT)
                     r.raise_for_status()
                     this.discardedEvents = set(r.json())
-                    this.discardedEvents.discard('Docked')	# should_send() assumes that we send 'Docked' events
+                    this.discardedEvents.discard('Docked')  # should_send() assumes that we send 'Docked' events
                     assert this.discardedEvents			# wouldn't expect this to be empty
```
Athanasius, line 509: Do we REALLY want an assert for this?  What if EDSM returns something that can cause this to be None?  Sounds more like a logger.error() and from this code condition to me.

plugins/edsm.py
```diff
@@ -331,87 +471,116 @@ def cmdr_data(data, is_beta):
...
+                    # Filter out unwanted events
+                    pending = list(filter(lambda x: x['event'] not in this.discardedEvents, pending))
 
                 if should_send(pending):
-                    (username, apikey) = credentials(cmdr)
+                    creds = credentials(cmdr)  # TODO: possibly unbound
+                    if creds is None:
+                        raise ValueError("Unexpected lack of credentials")
+
+                    (username, apikey) = creds
```
Athanasius, line 518: () not required ?

plugins/edsm.py
```diff
@@ -331,87 +471,116 @@ def cmdr_data(data, is_beta):
...
                         'fromSoftware': applongname,
                         'fromSoftwareVersion': appversion,
                         'message': json.dumps(pending, ensure_ascii=False).encode('utf-8'),
                     }
+
                     r = this.session.post('https://www.edsm.net/api-journal-v1', data=data, timeout=_TIMEOUT)
                     r.raise_for_status()
                     reply = r.json()
-                    (msgnum, msg) = reply['msgnum'], reply['msg']
+                    (msg_num, msg) = reply['msgnum'], reply['msg']
```
Athanasius, line 530: () not required ?

plugins/edsm.py
```diff
@@ -331,87 +471,116 @@ def cmdr_data(data, is_beta):
...
             plug.show_error(_("Error: Can't connect to EDSM"))
 
         if closing:
             return
 
 
-# Whether any of the entries should be sent immediately
-def should_send(entries):
+def should_send(entries: List[Mapping[str, Any]]) -> bool:
+    """
```
Athanasius, line 566: The navbeaconscan is about sending immediately, but the rest of the code is whether the event is sent at all.

plugins/edsm.py
A-UNDERSCORE-D: Annoyingly that is what this does. It bails out because of the try/accept. But agreed. Replacing with error and `continue`

PR #684, opened 2020-09-06
Feature/logging select loglevel

Athanasius:
Adds a GUI component in Settings>Configuration to set LogLevel.  This is used at startup if present, defaulting to logging.INFO.

EDMC.py now also uses logging, with CL arg `--loglevel` to set the info (again defaulting to logging.INFO).  Error messages converted to use logging instead of plain prints to stderr.

### DIFF
--- a/EDMC.py
+++ b/EDMC.py
@@ -131,8 +131,8 @@ def main():
                         except Exception:
                             logger.debug(f'Invalid journal entry {line!r}')
 
-            except Exception as e:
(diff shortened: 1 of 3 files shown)


A-UNDERSCORE-D (approved):

Athanasius:
We'll ignore those failing flake8 checks, prefs.py hasn't been cleaned up yet ;).

PR #685, opened 2020-09-07
Enhancement/logging rotating archive

Athanasius:
Implements logging to a rotating set of files in %TEMP%/EDMarketConnector


EDMCLogging.py
```diff
@@ -79,6 +83,24 @@ def __init__(self, logger_name: str, loglevel: int = _default_loglevel):
         self.logger_channel.setFormatter(self.logger_formatter)
         self.logger.addHandler(self.logger_channel)
 
+        ## Rotating Handler in sub-directory
+        ## We want the files in %TEMP%\{appname}\ as {logger_name}.log and rotated versions
+        ## This is {logger_name} so that EDMC.py logs to a different file.
+        logfile_rotating = pathlib.Path(tempfile.gettempdir())
+        logfile_rotating = logfile_rotating / f'{appname}'
+        try:
```
A-UNDERSCORE-D, line 91: Could use the [pathlib method](https://docs.python.org/3/library/pathlib.html#pathlib.Path.mkdir), `logfile_rotating.mkdir(exist_ok=True)`

EDMarketConnector.py
```diff
@@ -1083,3 +1084,5 @@ def messagebox_not_py3():
 
     root.after(0, messagebox_not_py3)
     root.mainloop()
+
+    logger.info('Exiting')
```
A-UNDERSCORE-D, line 1088:
```suggestion
    logger.info('Exiting')

```

EDMC.py
```diff
@@ -99,14 +97,16 @@ def main():
                 print(f'{appversion} ({newversion.title!r} is available)')
             else:
                 print(appversion)
-            sys.exit(EXIT_SUCCESS)
+            return
```
A-UNDERSCORE-D, line 100:
```suggestion

            return
```
newline on scope change

A-UNDERSCORE-D (approved):

Issue #515, opened 2020-06-20
Possible to respect windows scaling settings or add scaling option


Sabrehawk:
As a boomer i have bad eyes and its a bit of a pain to handle that tiny window on a 4k /32"   200% scaling setting, could we have options to increase the size of the app appearance?

Athanasius:
I've no idea if 'Tk' (the library used for all the EDMC UI) allows for this.

If it does then I won't be looking at this until after the migration Python 3.7, which only has a tentative ETA of "maybe by 2 weeks from now".

Sabrehawk:
awesome

Holen Sie sich Outlook für Android<https://aka.ms/ghei36>

________________________________

Athanasius:
Has anyone tried adjusting the properties of EDMarketConnector.exe to set scaling there ?

Athanasius:
> Has anyone tried adjusting the properties of EDMarketConnector.exe to set scaling there ?

Well, that seems to do nothing on my 2560x1440 monitors.

However I've just discovered we can set a tk scaling option, so will play around to see if we can do that on the fly, or if it would require a restart of EDMC after changing it.

Athanasius:
Please try out the [Pre-Release 4.1.0-beta2](https://github.com/EDCD/EDMarketConnector/releases/tag/Release%2F4.1.0-beta2) as it has a new UI Scaling option.  Pay attention to the changelog notes about how it does, and doesn't, work.

Athanasius:
In looking up documentation for another issue I've come across the DPI awareness and GDI scaling options we can set in the application manifest.  It's possible some of these might allow for EDMC scaling to work better.  Specifically this would be making the actualy **OS** scaling better, i.e. how things look with EDMC's own option still at 100%.

Ref: https://docs.microsoft.com/en-us/windows/win32/sbscs/application-manifests#dpiaware

Athanasius:
Please try out https://github.com/EDCD/EDMarketConnector/releases/tag/Release%2F4.1.0-beta8 and report back.

PR #687, opened 2020-09-08
Enhancement/515 ui scaling

Athanasius:
Implement a UI scaling option to address #515 

It might not be possible to have this take effect on the fly, but it does take effect on the next start of the program.


Athanasius:
I'm going to merge this so I can include it in 4.1.0-beta2 for people to test.

PR #701, opened 2020-09-14
Fix/671 no modules

Athanasius:
Get rid of the:

> AttributeError: 'list' object has no attribute 'values'

noise from plugins/eddn.

### DIFF
--- a/plugins/eddn.py
+++ b/plugins/eddn.py
@@ -270,14 +270,22 @@ Msg:\n{msg}''')
         :param data: dict containing the outfitting data
         :param is_beta: whether or not we're currently in beta mode
         """
-        modules: Dict[str, Any] = data['lastStarport'].get('modules') or {}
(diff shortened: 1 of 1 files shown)


plugins/eddn.py
```diff
@@ -270,14 +270,22 @@ def export_outfitting(self, data: Mapping[str, Any], is_beta: bool) -> None:
         :param data: dict containing the outfitting data
         :param is_beta: whether or not we're currently in beta mode
         """
-        modules: Dict[str, Any] = data['lastStarport'].get('modules') or {}
+        modules: Dict[str, Any] = data['lastStarport'].get('modules')
+        if not modules:
+            logger.debug('modules was None')
```
A-UNDERSCORE-D, line 275: if you want to check for None, use `is None`, other things resolve to False, such as an empty dict

A-UNDERSCORE-D (approved):

PR #704, opened 2020-09-15
Cleanup/commodity.py

Athanasius:
An attempt to get companion.py passing all our flake8 tests.

Some of the docstrings could be better, but at least they now exist.


A-UNDERSCORE-D (approved):
Im assuming you have already tested this so it should be good to merge

Athanasius:
Some minimal testing, yes.

1. Authentication from scratch works.
1. Using 'Update' button then works (which should be the same as the auto update on dock, and hits all three endpoints: /profile, /shipyard /market).
1. New Access Token on startup works (I'd not run it yet today and it just worked on that branch).

Issue #711, opened 2020-09-20
4.1.0-beta5 can't start


arymkus:
Checked Known Issues and Troubleshooting.

**Please complete the following information:**
 - Version 4.1.0-beta5
 - OS: Windows 10 1903
 - antivirus - Kaspersky EP, was turned on and later off to ensure it's not involved.
 - OS Locale: Russian
 - Browser: Firefox
 - Log file.
[EDMarketConnector.log](https://github.com/EDCD/EDMarketConnector/files/5252518/EDMarketConnector.log)

**Describe the bug**
EDMC can't start after upgrade from 4.1.0-b4 to 4.1.0-b5.
Ship's loadout is generated properly

**To Reproduce**
1. While being docked to a Fleet Carrier saw the error about value isn't in list and decided to upgrade.
2. Closed EDMC app
3. Without exiting downloaded new msi
4. Installed new version and tired to start it

**Expected behavior**
App should start.

**Screenshots**
Can't make a screenshot of an app without any visible parts.

Athanasius:
Presumably that's the EDMarketConnector.log from `%TEMP%` ?  If so, could you look in `%TEMP%\EDMarketConnector\` (note, this is a folder) and attach the relevant `EDMarketConnector.log` file from in there please.  It's the newer log file we're adding in 4.1.0 which always contains all the DEBUG logging.

arymkus:
[EDMarketConnector.log](https://github.com/EDCD/EDMarketConnector/files/5252533/EDMarketConnector.log)

arymkus:
I also noted a difference in ship's loadout between what was created by 4.1.0-b4 and one by 4.1.0-b5:
![изображение](https://user-images.githubusercontent.com/19597706/93721444-13dc6380-fb99-11ea-931a-b3266686216e.png)

Athanasius:
Well, then, it seems it's falling over on `locale.setlocale(locale.LC_ALL, (locale_startup[0], 'utf8'))` on your system, but in such a manner as to not log *anything* in either file about why it's falling over.

As a last ditch, could you open a command prompt, `cmd.exe`, navigate to where you have EDMC installed, and run `EDMarketConnector.exe` manually from there?  All the output should be redirected into the traditional log file, but let's see if anything useful shows up.

arymkus:
Started as you requested, no difference.
Both log files are attached:
[...]

PR #712, opened 2020-09-21
Attempt to select UTF-8 code page in manifest.

Athanasius:
Given the workaround we found for #711 talks about applications that are NOT unicode aware I looked into how to tell Windows that you *are*.

This small change should do that.


Athanasius:
Apparently this tweak has worked, so merging.

PR #717, opened 2020-09-22
Change monitor.py to use logging, not `__debug__` & print

Athanasius:
Some useful output is hidden behind `__debug__`.  Also some of it should be INFO anyway.

Draft as I work on this.


Athanasius:
This should be ready to go. Tested with game offline, at menu and on logging in.

A-UNDERSCORE-D (approved):

PR #719, opened 2020-09-23
Implement a TRACE level of logging

Athanasius:
We'll sometimes want detailed logging to trace down a bug, but don't want to always spam DEBUG level.

1. Defines logging.TRACE as 5.
1. Adds the string "TRACE" to match it.
1. Adds a new trace() method to log at this level.
1. Adds a CL arg on EDMarketConnector to set this.

Need to address EDMC.py as well.

### DIFF
--- a/EDMC.py
+++ b/EDMC.py
@@ -12,6 +12,7 @@ from os.path import getmtime, join
 from time import sleep, time
 from typing import TYPE_CHECKING, Any, Optional
 
+# See EDMCLogging.py docs.
 # isort: off
 from EDMCLogging import edmclogger, logger, logging
 if TYPE_CHECKING:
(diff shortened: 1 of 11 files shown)


Athanasius:
I need to remember to see if I can make a shortcut to ship that doesn't specify a working directory, but then just runs EDMarketConnector.exe with `--trace` to make it easy to get people to use it.

EDMC.py
```diff
@@ -1,27 +1,30 @@
...
 import sys
 from os.path import getmtime, join
 from time import sleep, time
-from typing import Any, Optional
+from typing import TYPE_CHECKING, Any, Optional
+
+# isort: off
+from EDMCLogging import edmclogger, logger, logging
+if TYPE_CHECKING:
+    from logging import trace, TRACE  # type: ignore # noqa: F401
```
A-UNDERSCORE-D, line 18: add a `# We add this ourselves at runtime` to this, that way in a few weeks we're not wondering how and why

A-UNDERSCORE-D (approved):

PR #720, opened 2020-09-27
Add CAPIData object 

A-UNDERSCORE-D:
CAPIData is a dict that we control, it will be replaced later with a more specific object that can be type checked.

### DIFF
--- a/plugins/eddn.py
+++ b/plugins/eddn.py
@@ -17,7 +17,7 @@ from typing import Sequence, TextIO
 import requests
 
 import myNotebook as nb  # noqa: N813
-from companion import category_map
+from companion import CAPIData, category_map
 from config import applongname, appversion, config
 from EDMCLogging import get_main_logger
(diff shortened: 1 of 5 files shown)


plugins/eddn.py
```diff
@@ -331,19 +306,20 @@ def export_outfitting(self, data: Mapping[str, Any], is_beta: bool) -> None:
...
         """
-        modules, ships = self.safe_modules_and_ships(data)
+
+        ships = data['lastStarport']['ships']
 
         horizons: bool = is_horizons(
             data['lastStarport'].get('economies', {}),
-            modules,
+            data['lastStarport']['modules'],
             ships
```
Athanasius, line 323: If we don't need the temporary var for modules then we don't need it now for ships either.

A-UNDERSCORE-D: Is used below in the construction of the Shipyard list (L326-331)

PR #746, opened 2020-10-08
Cleanup of config.py

Athanasius:
The intent is to pass flake8 and mypy tests.

### DIFF
--- a/config.py
+++ b/config.py
@@ -12,7 +12,7 @@ import warnings
 from os import getenv, makedirs, mkdir, pardir
 from os.path import dirname, expanduser, isdir, join, normpath
 from sys import platform
-from typing import Union
+from typing import Optional, Union
 
 import semantic_version
 
@@ -102,7 +102,7 @@ elif platform == 'win32':
     RegDeleteValue.restype = LONG
     RegDeleteValue.argtypes = [HKEY, LPCWSTR]
 
-    def known_folder_path(guid):
+    def known_folder_path(guid: uuid.UUID) -> Optional[str]:
         """Look up a Windows GUID to actual folder path name."""
         buf = ctypes.c_wchar_p()
         if SHGetKnownFolderPath(ctypes.create_string_buffer(guid.bytes_le), 0, 0, ctypes.byref(buf)):
@@ -476,11 +476,11 @@ class Config():
             self.save()
             self.config = None
 
-        def _escape(self, val):
+        def _escape(self, val: str) -> str:
             """Escape a string for storage."""
             return str(val).replace('\\', '\\\\').replace('\n', '\\n').replace(';', '\\;')
 
-        def _unescape(self, val):
+        def _unescape(self, val: str) -> str:
             """Un-escape a string from storage."""
             chars = list(val)
             i = 0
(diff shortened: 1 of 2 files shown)


config.py
```diff
@@ -82,194 +102,283 @@
...
+
             else:
                 return None
 
-        def getint(self, key):
+        def getint(self, key: str) -> int:
+            """Look up an integer configuration value."""
             try:
-                return int(self.settings.get(key, 0))	# should already be int, but check by casting
-            except:
```
A-UNDERSCORE-D, line 163: `ValueError: invalid literal for int() with base 10: 'asd'` et al is the most likely, I believe.

config.py
```diff
@@ -284,72 +393,86 @@ def __init__(self):
...
                 self.set('outdir', expanduser('~'))
 
-        def get(self, key):
+        def get(self, key: str) -> Union[None, list, str]:
+            """Look up a string configuration value."""
             try:
                 val = self.config.get(self.SECTION, key)
-                if u'\n' in val:	# list
-                    # ConfigParser drops the last entry if blank, so we add a spurious ';' entry in set() and remove it here
+                if u'\n' in val:  # list
```
A-UNDERSCORE-D, line 421:
```suggestion
                if '\n' in val:  # list
```

config.py
```diff
@@ -284,72 +393,86 @@ def __init__(self):
...
+                    # ConfigParser drops the last entry if blank,
+                    # so we add a spurious ';' entry in set() and remove it here
                     assert val.split('\n')[-1] == ';', val.split('\n')
                     return [self._unescape(x) for x in val.split(u'\n')[:-1]]
```
A-UNDERSCORE-D, line 425:
```suggestion
                    return [self._unescape(x) for x in val.split('\n')[:-1]]
```

config.py
```diff
@@ -284,72 +393,86 @@ def __init__(self):
...
+            """Set value on the specified configuration key."""
             if isinstance(val, bool):
                 self.config.set(self.SECTION, key, val and '1' or '0')
+
             elif isinstance(val, str) or isinstance(val, numbers.Integral):
                 self.config.set(self.SECTION, key, self._escape(val))
-            elif hasattr(val, '__iter__'):	# iterable
+
+            elif isinstance(val, list):
                 self.config.set(self.SECTION, key, u'\n'.join([self._escape(x) for x in val] + [u';']))
```
A-UNDERSCORE-D, line 451:
```suggestion
                self.config.set(self.SECTION, key, '\n'.join([self._escape(x) for x in val] + [u';']))
```

Athanasius: And presumably `u';'` -> `';'` as well.

A-UNDERSCORE-D: Yeah, missed one :D

config.py
Athanasius: I'll add an except for ValueError, but keep the general one for now, just in case.

config.py
Athanasius: Doing a search on u' and changing them all in this file....

Athanasius:
I ran into trouble with my pre-commit not liking the f'{var=}' syntax, turns out it needed updating to use Python 3.8, and whilst I was at it I stopped pinning an old version of it, let it use latest.

config.py
```diff
@@ -82,194 +102,288 @@
...
         return retval
 
-
-elif platform=='linux':
+elif platform == 'linux':
     import codecs
     from configparser import RawConfigParser
 
 
 class Config(object):
```
A-UNDERSCORE-D, line 119:
```suggestion
class Config:
```

No need to subclass object as we dont support py2 anymore

A-UNDERSCORE-D (approved):

PR #754, opened 2020-10-12
Cleanup of EDMarketConnector.py

Athanasius:
The main aim here is to get it passing my pre-commit hooks, so flake8 and mypy primarily.

I'm leaving TODO's behind on some things for later cleanup.

### DIFF
--- a/EDMarketConnector.py
+++ b/EDMarketConnector.py
@@ -94,7 +94,7 @@ class AppWindow(object):
     EVENT_BUTTON = 4
     EVENT_VIRTUAL = 35
 
-    def __init__(self, master: tk):
+    def __init__(self, master: tk.Tk):  # noqa: C901 # TODO - can possibly factor something out
 
         self.holdofftime = config.getint('querytime') + companion.holdoff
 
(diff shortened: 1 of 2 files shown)


EDMarketConnector.py
```diff
@@ -730,27 +775,41 @@ def crewroletext(role):
...
+
         except Exception as e:
             logger.debug('Frontier CAPI Auth:', exc_info=e)
             self.status['text'] = str(e)
+
         self.cooldown()
 
-    # Handle Status event
-    def dashboard_event(self, event):
+    def dashboard_event(self, event) -> None:
```
A-UNDERSCORE-D, line 807: Can we assume that event is at least a `Mapping[str, Any]` here?

EDMarketConnector.py
```diff
@@ -730,27 +775,41 @@ def crewroletext(role):
...
+        """
+        Handle DashBoardEvent tk event.
+
+        Event is sent by code in dashboard.py.
+        """
         entry = dashboard.status
         if entry:
```
A-UNDERSCORE-D, line 814: Do we want to use a guard clause here instead?
```
if not entry:
    return

<rest of code>
```

Saves an indent level

EDMarketConnector.py
```diff
@@ -806,30 +870,38 @@ def cooldown(self):
...
         webbrowser.open('https://github.com/EDCD/EDMarketConnector/wiki/Privacy-Policy')
 
-    def help_releases(self, event=None):
+    def help_releases(self, event=None) -> None:
+        """Open Releases page in browser."""
         webbrowser.open('https://github.com/EDCD/EDMarketConnector/releases')
 
     class HelpAbout(tk.Toplevel):
+        """The applications Help > About popup."""
+
```
A-UNDERSCORE-D, line 898: extra newline?

EDMarketConnector.py
```diff
@@ -901,28 +973,35 @@ def __init__(self, parent):
...
 
-        def _destroy(self):
+        def _destroy(self) -> None:
+            """Set parent window's topmost appropriately as we close."""
             self.parent.wm_attributes('-topmost', config.getint('always_ontop') and 1 or 0)
             self.destroy()
             self.__class__.showing = False
 
-    def save_raw(self):
+    def save_raw(self) -> None:  # noqa: CCR001
```
A-UNDERSCORE-D, line 986:
```suggestion
    def save_raw(self) -> None:  # noqa: CCR001 # Not easily broken up
```

EDMarketConnector.py
```diff
@@ -256,7 +264,7 @@ def __init__(self, master):
                                       image=self.theme_icon, cursor='fleur',
                                       anchor=tk.W, compound=tk.LEFT)
             theme_titlebar.grid(columnspan=3, padx=2, sticky=tk.NSEW)
-            self.drag_offset = None
+            self.drag_offset: Tuple[Union[None, int], Union[None, int]] = (None, None)
```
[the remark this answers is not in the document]
A-UNDERSCORE-D: If you use this dont forget to import Optional

EDMarketConnector.py
Athanasius: That results in:

```py
EDMarketConnector.py:340: error: Argument 2 to "bind_all" of "Misc" has incompatible type "Callable[[Mapping[str, Any]], None]"; expected "Optional[Callable[[Event[Misc]], Optional[Literal['break']]]]"
```

A-UNDERSCORE-D: Oh, thats wonderful. Nevermind then

EDMarketConnector.py
Athanasius: Newline after *class* docstring is mandatory.

Annoyingly *not* having such after function docstring is mandatory.  It keeps tripping me up :P .

A-UNDERSCORE-D: Yeah it likes catching me too

A-UNDERSCORE-D (approved):

PR #757, opened 2020-10-13
Add detailed logging to shutdown code

Athanasius:
The branch name is because of ensuring we notify the user when shutting down.  This will also now change the status text to 'Shutdown down...' and *not* 'withdraw' the window in the meantime.

Also adding more logging to startup so we have at least one logged line recording thread id.

### DIFF
--- a/EDMarketConnector.py
+++ b/EDMarketConnector.py
@@ -1163,8 +1163,8 @@ def enforce_single_instance() -> None:  # noqa: CCR001
                     and GetProcessHandleFromHwnd(window_handle):
                 # If GetProcessHandleFromHwnd succeeds then the app is already running as this user
                 if len(sys.argv) > 1 and sys.argv[1].startswith(protocolhandler.redirect):
-                    logger.debug('Browser invoked us directly with auth response. '
-                                 'Forwarding the response to the other app instance.')
+                    # logger.debug('Browser invoked us directly with auth response. '
+                    #              'Forwarding the response to the other app instance.')
                     CoInitializeEx(0, COINIT_APARTMENTTHREADED | COINIT_DISABLE_OLE1DDE)
                     # Wait for it to be responsive to avoid ShellExecute recursing
                     ShowWindow(window_handle, SW_RESTORE)
@@ -1174,12 +1174,13 @@ def enforce_single_instance() -> None:  # noqa: CCR001
                     ShowWindowAsync(window_handle, SW_RESTORE)
                     SetForegroundWindow(window_handle)
 
-                logger.info(f'A running {applongname} process was found, exiting.')
-                sys.exit(0)
+                # logger.info(f'A running {applongname} process was found, exiting.')
+                return False
 
             return True
 
-        EnumWindows(enumwindowsproc, 0)
+        if not EnumWindows(enumwindowsproc, 0):
+            sys.exit(0)
 
 
 def test_logging() -> None:
--- a/EDMarketConnector.py
+++ b/EDMarketConnector.py
@@ -1149,6 +1149,7 @@ def enforce_single_instance() -> None:  # noqa: CCR001
                     ShowWindowAsync(window_handle, SW_RESTORE)
                     SetForegroundWindow(window_handle)
 
+                logger.info(f'A running {applongname} process was found, exiting.')
                 sys.exit(0)
 
             return True
--- a/EDMarketConnector.py
+++ b/EDMarketConnector.py
@@ -1139,7 +1139,8 @@ def enforce_single_instance() -> None:  # noqa: CCR001
                     and GetProcessHandleFromHwnd(window_handle):
                 # If GetProcessHandleFromHwnd succeeds then the app is already running as this user
                 if len(sys.argv) > 1 and sys.argv[1].startswith(protocolhandler.redirect):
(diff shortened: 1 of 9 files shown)


EDMarketConnector.py
```diff
@@ -1139,7 +1163,8 @@ def enumwindowsproc(window_handle, l_param):
                     and GetProcessHandleFromHwnd(window_handle):
                 # If GetProcessHandleFromHwnd succeeds then the app is already running as this user
                 if len(sys.argv) > 1 and sys.argv[1].startswith(protocolhandler.redirect):
-                    # Browser invoked us directly with auth response. Forward the response to the other app instance.
+                    logger.debug('Browser invoked us directly with auth response. '
```
A-UNDERSCORE-D, line 1166: Will this write to our normal log file? I think it will. Possibly stepping in the middle of a write?

dashboard.py
```diff
@@ -119,16 +144,20 @@ def process(self, logfile=None):
...
+                    if (
+                            timegm(time.strptime(entry['timestamp'], '%Y-%m-%dT%H:%M:%SZ')) >= self.session_start
+                            and self.status != entry
+                    ):
                         self.status = entry
                         self.root.event_generate('<<DashboardEvent>>', when="tail")
-        except:
-            if __debug__: print_exc()
+
+        except Exception as e:
```
A-UNDERSCORE-D, line 159:
```suggestion
        except Exception:
```
Not using e anyway

protocol.py
```diff
@@ -225,10 +226,20 @@ def start(self, master):
...
                 self.thread = None
+
                 if self.httpd:
+                    logger.info('Shutting down httpd')
                     self.httpd.shutdown()
-                thread.join()	# Wait for it to quit
+
+                logger.info('Joining thread')
+                thread.join()  # Wait for it to quit
+
```
A-UNDERSCORE-D, line 238:
```suggestion
logger.info('thread exited')

```

EDMarketConnector.py
Athanasius: The fun thing is in checking what happened with that ... now *this* code is:

```py
Exception ignored on calling ctypes callback function: <function enforce_single_instance.<locals>.enumwindowsproc at 0x04D2CBF8>
  File "C:/Users/Athan/Documents/Devel/EDMarketConnector/EDMarketConnector.py", line 1178, in enumwindowsproc
    sys.exit(0)
SystemExit: 0```
when it works fine with the Release/4.1.3 code.

Comment (Athanasius) [reply EDMarketConnector.py:1166]: Now, is there any chance that Python 3.8.6 changed behaviour on this from 3.7.9 ?  i.e. is the solution to 'return False' and check that ?

Comment (Athanasius) [reply EDMarketConnector.py:1166]: Well, that does seem to be working.

Comment (Athanasius) [reply EDMarketConnector.py:1166]: And now I checked what that did with log files... yes, any logging from that will cause overwrite of the plain log file, but unless we adjust things that will have happened by this point now anyway, as we do the redirect, which truncates, ASAP.

The debug log file appears to work OK.  This is with 4.1.3 running and using this branch's code as second instance:

```py
2020-10-14 19:38:46.043 - DEBUG - companion.Session.start:455: Starting session
2020-10-14 19:41:38.786 - INFO - 17404:1004:1004 EDMarketConnector.__main__:1227: Startup v4.2.0-beta1 : Running on Python v3.8.6 (tags/v3.8.6:db45529, Sep 23 2020, 15:37:30) [MSC v.1927 32 bit (Intel)]
2020-10-14 19:47:43.320 - DEBUG - 16912:7884:7884 hotkey.HotkeyMgr.unregister:320: No thread
2020-10-14 19:47:43.528 - DEBUG - 16912:7884:7884 hotkey.HotkeyMgr.unregister:322: Done.
2020-10-14 19:49:46.415 - INFO - <plugins>.plugintest.load.plugin_stop:105: Stopping
2020-10-14 19:49:46.438 - INFO - EDMarketConnector.__main__:1204: Exiting
```
Note the two hotkey lines in the middle with process/thread IDs from this branch.  Just the import of hotkey is enough to cause that thread to start, and then main thread exit causes it to exit and log that.

So it *looks* like debug log file contents will end up done line by line, no overwrite.

Athanasius: I can move the call to `enforce_single_instance()` to before even the redirect, and have no logging, *but* the point of this was to make it *obvious* when "installed new version, won't run" is because there's a lingering EDMarketConnector.exe process.

If we really don't want to overwrite EDMarketConnector.log in that case then I'd want to add a pop-up of some form in that case so the user is informed.  I'll take a look at the best way to do that tomorrow.

protocol.py
Athanasius: This is covered by the end of function 'Done.'.

EDMarketConnector.py
A-UNDERSCORE-D: Up to you if we overwrite, though I'd rather see about appending instead

Athanasius:
Tested with 4.1.4 already running (and when not).  Closing the new popup with 'OK' or the decoration 'X' causes process to exit as expected.  Passing all my pre-commit checks for EDMarketConnector.py.

The print to the plain log file happens in append mode, but will be overwritten by subsequent already-running process output, hence the popup to be sure the user is informed.

PR #758, opened 2020-10-13
Added killswitch library

A-UNDERSCORE-D:
Killswitch exists to allow us to disable features causing issues in
specific versions remotely.

The system is designed to support plugins, where they can request their
own killswitch list by querying any web file that follows our format
(docs to be added), and using the returned KillSwitchSet to check for
disabled features.

### DIFF
--- a/docs/Killswitches.md
+++ b/docs/Killswitches.md
@@ -54,3 +54,9 @@ The current recognised (to EDMC and its internal plugins) killswitch strings are
 | `plugins.(eddn|inara|edsm|eddb).journal`             | Disables all journal processing for EDDN/EDSM/INARA                                          |
 | `plugins.(edsm|inara).worker`                        | Disables the EDSM/INARA worker thread (effectively disables updates) (does not close thread) |
 | `plugins.(eddn|inara|edsm).journal.event.$eventname` | Specific events to disable processing for                                                    |
+
+## File location
+
+The main killswitch file is kept in the `releases` branch on the EDMC github repo. The file should NEVER be committed to
+any other repos. In the case that the killswitch file is found in other repos, the one in releases should always
+be taken as correct regardless of others.
--- a/docs/Killswitches.md
+++ b/docs/Killswitches.md
@@ -6,18 +6,18 @@ EDMarketConnector implements a Kill Switch system that allows us to disable feat
 
 Killswitches are stored in a JSON file that is queried by EDMC on startup. The format is as follows:
 
-|             Key |  Type   | Description                                                   |
-| --------------: | :-----: | :------------------------------------------------------------ |
-|       `version` | integer | the version of the Kill Switch JSON file, always 1            |
-|  `last_updated` | string  | When last the kill switches were updated (for human use only) |
-| `kill_switches` |  array  | The kill switches this file contains (expanded below)         |
+|             Key |   Type   | Description                                                   |
+| --------------: | :------: | :------------------------------------------------------------ |
+|       `version` |  `int`   | the version of the Kill Switch JSON file, always 1            |
+|  `last_updated` | `string` | When last the kill switches were updated (for human use only) |
+| `kill_switches` | `array`  | The kill switches this file contains (expanded below)         |
 
 The `kill_switches` array contains kill switch objects. Each contains two fields:
 
 |       Key |        Type        | Description                                                             |
 | --------: | :----------------: | :---------------------------------------------------------------------- |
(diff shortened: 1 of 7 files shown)


killswitch.py
```diff
@@ -0,0 +1,189 @@
...
+import requests
+import semantic_version
+
+import config
+import EDMCLogging
+
+logger = EDMCLogging.get_main_logger()
+
+# DEFAULT_KILLSWITCH_URL = 'https://github.com/EDCD/EDMarketConnector'
+DEFAULT_KILLSWITCH_URL = 'http://127.0.0.1:8080/killswitches.json'
```
Athanasius, line 13: We need to have a think about where to host the kill switches.  Given the appcast file is via raw github file I see no reason not to utilise the same for this file.  

However, what branch ?  The appcast is looked for in the `releases` branch.  Given the version/branch-agnostic nature of the killswitch file (pertinent application versioning is within it) it could also be in the `releases` branch.  We would just need to remember to take care with merges so as to not accidentally unintentionally change the appcast file.  In fact I think the defined procedure should be to simply commit only to the `releases` branch for edits to the killswitch file.

docs/Killswitches.md
```diff
@@ -0,0 +1,55 @@
...
+                "plugins.eddn.send": "some reason"
+            }
+        }
+    ]
+}
+```
+
+### Versions
+
+Versions are checked using equality checks on `semantic_version.Version` instances. Meaning that **all** fields are checked (ie, Major, Minor, Patch, Prerelease, and Build).
```
Athanasius, line 40: IIRC there's support in semantic_version for things like a <= comparison.  We *might* discover that something affects "every version before we fixed it" and wouldn't want to therefor list every version (since killswitch support was added) up to the last broken one.  It would be cleaner to be able to specify a version as "<= 4.2.6" if we fixed it in 4.2.7.

killswitch.py
A-UNDERSCORE-D: Could be in releases or in main, I've left it because Im not sure where exactly as of yet. Releases sounds like the better option as its where appcast stuff is done already, and having that all together seems like a better idea

docs/Killswitches.md
A-UNDERSCORE-D: Yeah I was poking at it a little, https://python-semanticversion.readthedocs.io/en/latest/reference.html#semantic_version.SimpleSpec looks to be what would work for it I think?

killswitch.py
```diff
@@ -0,0 +1,189 @@
...
+
+
+def get_reason(id: str, *, version=_current_version) -> str:
+    """Query the global KillSwitchSet#get_reason method."""
+    return active.get_reason(id, version=version)
+
+
+if __name__ == "__main__":
+    setup_main_list()
+    print(f'{_current_version=}')
```
Athanasius, line 187: This is either "prototyping" and needs removing, or should be converted to logging.

docs/Killswitches.md
Athanasius: Yup, exactly what I had in mind.  I'll leave you to figure out the most succinct way of making use of it :) .

killswitch.py
A-UNDERSCORE-D: Was prototyping, explicitly in the main guard for that reason, I'll remove it before I mark this ready to merge

plugins/eddn.py
```diff
@@ -596,6 +602,14 @@ def plugin_stop() -> None:
 def journal_entry(  # noqa: C901
     cmdr: str, is_beta: bool, system: str, station: str, entry: MutableMapping[str, Any], state: Mapping[str, Any]
 ) -> Optional[str]:
+    if (ks := killswitch.get_disabled("plugins.eddn.journal")).disabled:
+        logger.warning(f"EDDN journal handler has been disabled via killswitch: {ks.reason}")
+        plug.show_error("EDDN journal handler disabled. See Log.")
+        return None
+    elif (ks := killswitch.get_disabled(f'plugins.eddn.journal.event{entry["event"]}')).disabled:
```
Athanasius, line 609: Isn't this missing a `.`, i.e. it should be

`killswitch.get_disabled(f'plugins.eddn.journal.event.{entry["event"]}')).disabled`

?  I'm going to guess the other plugins will have the same issue.

A-UNDERSCORE-D: Whoops. Yes

A-UNDERSCORE-D: Fixed

Issue #755, opened 2020-10-12
Release 4.1.3 Silently failing


TonyFlury:
Having upgraded from the previous release to Release 4.1.3 on Windows 10, I am unable to run the new release.

There is no error message, and even running the executable from the command line provides no console based errors either.

I have done an entirely clean installation as well and still had no joy.

I don't know how to help diagnose this issue. For the time being I will revert.

Athanasius:
Can I ask what path you used to open this issue?  I thought I'd found all the references to the no-template issue form, replacing them with the Bug Report specific one.

Athanasius:
As you didn't use the Bug Report template could you please fill it in in a comment here:

**Please check the [Known Issues](https://github.com/EDCD/EDMarketConnector/issues/618) in case this has already been reported.**

**Please also check if the issue is covered in our [Troubleshooting Guide](https://github.com/EDCD/EDMarketConnector/wiki/Troubleshooting).** It might be something with a known work around, or where a third party (such as EDSM) is causing logging that is harmless.

**Please complete the following information:**
 - Version [e.g. 4.0.6 - See 'Help > About E:D Market Connector'.  If running from source using git then please paste the output of `git log --decorate=full | head -1`]
 - OS: [e.g. Windows 10, Linux Debian 10.6, etc.]
 - OS Locale: [e.g. English, French, Serbian...]
 - If applicable: Browser [e.g. chrome, safari]
 - Please attach log files:
     1. Always attach `%TEMP%\EDMarketConnector.log` from *immediately* after the bug occurs (re-running the application overwrites this file).
     1. If running 4.1.0 (including betas) or later: `%TEMP%\EDMarketConnector\EDMarketConnector-debug.log`.  See [Debug Log File](https://github.com/EDCD/EDMarketConnector/wiki/Troubleshooting#debug-log-files).
   
   Use the Icon that looks like `_`, `|` and `^` all in one character towards the right of the toolbar above.

**Describe the bug**
A clear and concise description of what the bug is.

**To Reproduce**
Steps to reproduce the behavior:
1. Go to '...'
2. Click on '....'
3. Scroll down to '....'
4. See error

**Expected behavior**
A clear and concise description of what you expected to happen.

**Screenshots**
If applicable, add screenshots to help explain your problem.

**Additional context**
Add any other context about the problem here.

NoFoolLikeOne:
Hey Tony. What plugins are you running?

TonyFlury:
[...]

PR #760, opened 2020-10-14
Locale: Be even more paranoid about setting UTF-8 encoding

Athanasius:
We have at least one instance of not starting on Windows 7 still, see #755


Athanasius:
Mostly this needs a spot check on Linux to be sure that conditional order doesn't cause any "windows_ver ... None has no member" or whatever errors.

A-UNDERSCORE-D:
Launches fine for me on 3.8

A-UNDERSCORE-D (approved):

PR #769, opened 2020-10-26
Cleanup l10n file

A-UNDERSCORE-D:


PLUGINS.md
```diff
@@ -225,7 +225,7 @@ to retrieve your plugin's settings in a platform-independent way.
 **Be sure to use a unique prefix for any settings you save so as not to clash
 with core EDMC or other plugins.**
 
-Use `numberFromString()` from EDMC's `l10n.Locale` object to parse input
+Use `number_from_string()` from EDMC's `l10n.Locale` object to parse input
```
Athanasius, line 228: Urgh, I know it's the standard, but this counts as changing a published API.  That means bumping major version on the first release with this in.

On those grounds it would be better to keep the old name for the time being.

l10n.py
```diff
@@ -1,45 +1,60 @@
...
-from traceback import print_exc
-import builtins
+from typing import TYPE_CHECKING, Dict, Iterable, Optional, Set, TextIO, Union, cast
+
+if TYPE_CHECKING:
+    def _(x: str) -> str: ...
 
-import locale
 try:
     locale.setlocale(locale.LC_ALL, '')
```
Athanasius, line 22: It bothers me slightly that we're now doing this both here, and later (given the import is before) in EDMarketConnector.py's `__main__`.  But, both are doing it when they are for good reasons, so I guess I can live with it.  Maybe add a comment to each site so it's clear that changing it 'here' won't necessarily change overall behaviour.

l10n.py
```diff
@@ -48,235 +63,325 @@
...
+        Install a dummy translation function.
 
-    def install_dummy(self):
-        # For when translation is not desired or not available
-        self.translations = { None: {} }
-        builtins.__dict__['_'] = lambda x: str(x).replace(r'\"', u'"').replace(u'{CR}', u'\n')	# Promote strings to Unicode for consistency
+        Use when translation is not desired or not available
+        """
+        self.translations = {None: {}}
+        # Promote strings to Unicode for consistency
```
Athanasius, line 84: Misleading comment, as we're no longer doing that.  What the next line does is some replacements on quotes and the like.

l10n.py
```diff
@@ -48,235 +63,325 @@
...
+                except UnicodeDecodeError as e:
+                    logger.warning(f'Malformed file {lang}.strings in plugin {plugin}: {e}')
+
+                except Exception:
+                    logger.exception(f'Exception occurred while parsing {lang}.strings in plugin {plugin}')
+
+        builtins.__dict__['_'] = self.translate
+
+    def contents(self, lang: str, plugin_path: Optional[str] = None) -> Dict[str, str]:
+        """Get all the translations from a translation file."""
```
Athanasius, line 131: s/Get/Load/

l10n.py
```diff
@@ -48,235 +63,325 @@
...
+        """
+        Open the given lang file for reading.
+
+        :param lang: The lang file to open (just the name of the lang)
+        :param plugin_path: path to plugins dir, to check for plugin based lang files, defaults to None
+        :return: the opened file (Note: This should be closed when done)
+        """
         if plugin_path:
-            f = join(plugin_path, '%s.strings' % lang)
+            f = join(plugin_path, f'{lang}.strings')
```
Athanasius, line 222: Not using pathlib.Path() here ?

l10n.py
```diff
@@ -48,235 +63,325 @@
...
             self.float_formatter.setMinimumFractionDigits_(5)
             self.float_formatter.setMaximumFractionDigits_(5)
 
-    def stringFromNumber(self, number, decimals=None):
-        # Uses the current system locale, irrespective of language choice.
-        # Unless `decimals` is specified, the number will be formatted with 5 decimal
-        # places if the input is a float, or none if the input is an int.
+    def stringFromNumber(self, number: Union[float, int], decimals: int = None) -> str:  # noqa: N802
+        warnings.warn(DeprecationWarning('use _Locale.string_from_number instead.'))
+        return self.stringFromNumber(number, decimals)
```
Athanasius, line 252: Shouldn't this be `self.string_from_number(...)` ?

l10n.py
```diff
@@ -48,235 +63,325 @@
...
+
+    def numberFromString(self, string: str) -> Union[int, float, None]:  # noqa: N802
+        warnings.warn(DeprecationWarning('use _Locale.number_from_string instead.'))
+        return self.numberFromString(string)
```
Athanasius, line 256: Again with what looks like infinite recursion to me.

l10n.py
```diff
@@ -48,235 +63,325 @@
...
+
+    def preferredLanguages(self) -> Iterable[str]:  # noqa: N802
+        warnings.warn(DeprecationWarning('use _Locale.preferred_languages instead.'))
+        return self.preferred_languages()
```
Athanasius, line 260: Again with what looks like infinite recursion to me.

PLUGINS.md
Athanasius: Ah, you attempted to just make the old names deprecated, but see comments on those bits, it looks like you brainfarted and have introduced infinite recursion if the old names are used.

l10n.py
```diff
@@ -48,235 +63,325 @@
...
+        """
+        Convert a number to a string.
+
+        Uses the current system locale, irrespective of language choice.
+
+        :param number: The number to stringify
+        :param decimals: The number of decimals to return, defaults to 5 if the given number is a float, otherwise None
+        :return: the stringified number
+        """
+        decimals = decimals if decimals is not None else 5
```
Athanasius, line 272: I'm struggling to see how this is different from just having `decimals: int =  5` in the function signature.

1. decimals not passed, so it defaults to None, and ends up set to 5 here.
2. decimals passed as None, it will be set to 5 here.
3. decimals passed as a valid integer, it will be set to that here.

I guess this means that if you literally want to have no decimals on what *is* a floating point number you'd need to pass `0`.  Indeed the next bit of code handles that.

l10n.py
A-UNDERSCORE-D: Whoops! yeah it should be

l10n.py
A-UNDERSCORE-D: Yeah I think that was a "I translated from or-magic to a ternary without looking to closely". Fixed now

l10n.py
A-UNDERSCORE-D: This and the other one have been fixed

l10n.py
A-UNDERSCORE-D: Switched.

l10n.py
A-UNDERSCORE-D: Removed the comment, what the below does is reasonably self explanitory

l10n.py
A-UNDERSCORE-D: Done

l10n.py
A-UNDERSCORE-D: Yeah Im not sure if we want to remove it here or not. For now I added the suggested comment

PLUGINS.md
A-UNDERSCORE-D: Yep! fixed now

l10n.py
```diff
@@ -48,235 +63,325 @@
...
+
 
 # singletons
-Locale = Locale()
-Translations = Translations()
+Locale = _Locale()
+Translations = _Translations()
 
 
 # generate template strings file - like xgettext
```
Athanasius, line 354: OH, that's what this code is doing.

This makes me nervous, because I doubt this is going to handle where we split the arguments to `_(...)` calls over multiple lines.  If anyone runs this in the future it will generate an incomplete file.

At the very least I think we need to change the template file to have a `.new` extension so you can compare it to the current one before deciding to use it or not.

PLUGINS.md
Athanasius: Maybe explicitly document here that using the old name will cause deprecation logging *but continue to work for the time being*.  Just so no-one screams about the change, despite it only causing log noise at this point.

l10n.py
Athanasius: For completeness sake this PR also needs to add the equivalent comment in EDMarketConnector.py

Athanasius (changes requested):

l10n.py
A-UNDERSCORE-D: Yeah thats not a bad idea. Honestly using the python AST to find these is PROBABLY a better idea, if I can figure out how. This isnt exactly the smartest method to doing this.

A-UNDERSCORE-D: Okay Ive reimplemented this (see: https://gist.github.com/A-UNDERSCORE-D/8bb1d769c51295e827ec92951c464d94 ) And will add a new PR to replace this code once this PR is merged. In the interim, I've updated the old code to create the file with a .new as suggested.

l10n.py
A-UNDERSCORE-D: Done in both now

Athanasius (approved):

PR #770, opened 2020-11-04
Add multi platform transparency

A-UNDERSCORE-D:
This is distinct from the windows only transparent theme, I wanted to enable that on other platforms too but could not find a way.

The way this works is it changes the transparency of the entire EDMC window. Tested on windows and linux, works as expected for both.
Screenshot from 2020-11-04 17-57-19

### DIFF
--- a/EDMarketConnector - reset-ui.bat
+++ b/EDMarketConnector - reset-ui.bat
@@ -1,0 +1,1 @@
+EDMarketConnector.exe --reset-ui
--- a/EDMarketConnector.py
+++ b/EDMarketConnector.py
@@ -1406,6 +1406,13 @@ sys.path: {sys.path}'''
             )
             config.set('plugins_not_py3_last', int(time()))
 
+    # UI Transparency
+    ui_transparency = config.getint('ui_transparency')
+    if ui_transparency == 0:
+        ui_transparency = 100
+
+    root.wm_attributes('-alpha', ui_transparency / 100)
+
     root.after(0, messagebox_not_py3)
     root.mainloop()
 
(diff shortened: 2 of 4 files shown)


A-UNDERSCORE-D:
I probably should figure out how to do the translations with this before its merged

Athanasius:
1. Basic functionality checks out on my Windows 10 install (2004).
1. I didn't actually see any issue with using this with the Transparent theme, despite the warning.
1. Unless *that* is what causes this new transparency to *not* be applied when you switch back to Dark theme.  This also happens if you start on the Default theme, change transparency, exit settings, go back in, change to Dark theme.  A restart of the app does apply the selected transparency to either Default or Dark theme as expected.

A-UNDERSCORE-D:
Yeah I spotted that bug as well. I think its just something weird about how the transparency is done for the theme. The warning is there just to try to differentiate the two, they will work together, but it has "unexpected" results, causes the text to be less opaque which (I would assume) is not what users want

Athanasius:
> Yeah I spotted that bug as well. I think its just something weird about how the transparency is done for the theme.

So, did you figure out the cause of that bug ?  I have a vague memory of trying to track it down, but it's a royal PITA due to how the whole thing works, i.e. it's difficult to set a breakpoint and step through such that you can *see* the effect of each line because you don't see it at all until a forced UI update call.

A-UNDERSCORE-D:
> > Yeah I spotted that bug as well. I think its just something weird about how the transparency is done for the theme.
> 
> So, did you figure out the cause of that bug ? I have a vague memory of trying to track it down, but it's a royal PITA due to how the whole thing works, i.e. it's difficult to set a breakpoint and step through such that you can _see_ the effect of each line because you don't see it at all until a forced UI update call.

I did not, I feel like its corner case enough to go under "its okay"? but it IS annoying. Possibly a `restart may be required for changes to take effect`?

Athanasius:
OK, let's merge this and see if we can get people to test it to see if anything else shakes out.

PR #776, opened 2020-11-23
Reimplement config code (but not actual storage)

A-UNDERSCORE-D:
This started as a quick cleanup but ended up as a rewrite.

The rewrite is 100% compatible with the old implementation. But also raises more warnings and exceptions on bad state.

Additionally:
- `get` has been deprecated in preference to `get_$type` methods, which provide more type safety.
- If a type is supported, it is supported everywhere. No more "bools are broken on windows".
- The re-implementation is based on an abstract base class. This ensures implementation at least follows the same set of signatures everywhere. 
- There is a complete test suite added, using pytest
  - Tests exist both for both the new config standalone, and the new config reading data from the old config


A-UNDERSCORE-D:
This should be good now. No deprecation warnings are raised on start (related to this, anyway) and there do not appear to be any references to `.get()` or `.getint()` left in our code

EDMarketConnector.py
```diff
@@ -209,7 +209,7 @@ class AppWindow(object):
 
     def __init__(self, master: tk.Tk):  # noqa: C901, CCR001 # TODO - can possibly factor something out
 
-        self.holdofftime = config.getint('querytime') + companion.holdoff
+        self.holdofftime = cast(int, config.get_int('querytime', 0)) + companion.holdoff
```
Athanasius, line 212: Enlighten me as to why this cast() is needed, given the function signature of `config.get_int()` has `-> int` ?

PLUGINS.md
```diff
@@ -219,8 +219,8 @@ title will be the value that you returned from `plugin_start3`. Use widgets
 from EDMC's myNotebook.py for the correct look-and-feel. You can be notified
 when the settings dialog is closed so you can save your settings.
 
-You can use `set()`, `get()` and `getint()` from EDMC's `config.config` object
-to retrieve your plugin's settings in a platform-independent way.
+You can use `set()` and `get_$type()` (where type is one of: `int`, `bool`, `str`, `list`) from EDMC's
```
Athanasius, line 222: Add something to be explicit about what the old methods were and that they're deprecated.  People won't necessarily notice the code logged deprecation warnings.

EDMarketConnector.py
A-UNDERSCORE-D: Oh whoops. I thought I got all of those. At one point it was `-> Optional[int]`. I decided to drop that because it really wasnt worth the hassle and its explained in the docstring what happens on a default

config.py
```diff
@@ -31,23 +36,29 @@
 update_feed = 'https://raw.githubusercontent.com/EDCD/EDMarketConnector/releases/edmarketconnector.xml'
 update_interval = 8*60*60
 
-if getenv("EDMC_NO_UI"):
+if os.getenv("EDMC_NO_UI"):
```
Athanasius, line 39: This should just be able to use EDMCLogging.get_main_logger(), rather than repeating the if/else code.

config.py
```diff
@@ -511,5 +263,547 @@ def delete_password(self, account: str) -> None:
...
+        try:
+            self.config.read(self.filename)
+        except Exception as e:
+            logger.debug(f'Error occurred while reading in file. Assuming that we are creating a new one: {e}')
+            self.config.add_section(self.SECTION)
+
+        if (outdir := self.get_str('outdir')) is None or not pathlib.Path(outdir).is_dir():
+            self.set('outdir', str(self.home))
+
+    def __escape(self, s: str) -> str:
```
Athanasius, line 669: No handy-dandy module for this sort of thing?

Are strings 'bare' in the file, or do we need to handle the relevant quote marks ?

loadout.py
```diff
@@ -22,15 +22,15 @@ def export(data, filename=None):
...
-    oldfiles = sorted([x for x in os.listdir(config.get('outdir')) if regexp.match(x)])
+    oldfiles = sorted([x for x in os.listdir(config.get_str('outdir')) if regexp.match(x)])
     if oldfiles:
-        with open(join(config.get('outdir'), oldfiles[-1]), 'rU') as h:
+        with open(join(config.get_str('outdir'), oldfiles[-1]), 'rU') as h:
             if h.read() == string:
                 return	# same as last time - don't write
 
-    querytime = config.getint('querytime') or int(time.time())
+    querytime = config.get_int('querytime') or int(time.time())
```
Athanasius, line 31: Use the default in the get_int() call here ?

outfitting.py
```diff
@@ -509,7 +509,7 @@ def lookup(module, ship_map, entitled=False):
 
 def export(data, filename):
 
-    querytime = config.getint('querytime') or int(time.time())
+    querytime = config.get_int('querytime') or int(time.time())
```
Athanasius, line 512: Use default in get_int()

plugins/eddn.py
```diff
@@ -487,7 +487,7 @@ def export_journal_entry(self, cmdr: str, is_beta: bool, entry: Mapping[str, Any
 
             if (
                 entry['event'] == 'Docked' or (entry['event'] == 'Location' and entry['Docked']) or not
-                (config.getint('output') & config.OUT_SYS_DELAY)
+                (config.get_int('output') & config.OUT_SYS_DELAY)
```
Athanasius, line 490: Hmmm, now I wonder if this is the time to abstract these flags.  So that rather than storing them all in one bit-packed DWORD they're actually separate.

But perhaps that's better saved for the great storage method unification.

plugins/eddn.py
```diff
@@ -714,7 +714,7 @@ def filter_localised(d: Mapping[str, Any]) -> OrderedDictT[str, Any]:
                 this.commodities = this.outfitting = this.shipyard = None
                 this.marketId = entry['MarketID']
 
-            path = pathlib.Path(str(config.get('journaldir') or config.default_journal_dir)) / f'{entry["event"]}.json'
+            path = pathlib.Path(str(config.get_str('journaldir') or config.default_journal_dir)) / f'{entry["event"]}.json'
```
Athanasius, line 717: again, pass default to get_str() rather than the 'or'.

plugins/edsm.py
```diff
@@ -124,20 +124,20 @@ def plugin_start3(plugin_dir: str) -> str:
     this._IMG_ERROR = tk.PhotoImage(data=IMG_ERR_B64)  # BBC Mode 5 '?'
 
     # Migrate old settings
-    if not config.get('edsm_cmdrs'):
-        if isinstance(config.get('cmdrs'), list) and config.get('edsm_usernames') and config.get('edsm_apikeys'):
+    if not config.get_list('edsm_cmdrs'):
+        if isinstance(config.get_list('cmdrs'), list) and config.get_list('edsm_usernames') and config.get_list('edsm_apikeys'):
```
Athanasius, line 128: Technically there's no way config.get_list() can return other than `list` now, right ?  Actually, given this is some very old settings migration code, perhaps that first get_list() should still be get() ?  I think before 2.34 it was a simple string ?

prefs.py
```diff
@@ -407,7 +407,7 @@ def __setup_config_tab(self, notebook: Notebook) -> None:
         row = AutoInc(start=1)
 
         self.logdir = tk.StringVar()
-        self.logdir.set(str(config.get('journaldir') or config.default_journal_dir or ''))
+        self.logdir.set(str(config.get_str('journaldir') or config.default_journal_dir or ''))
```
Athanasius, line 410: `config.get_str('journaldir' default=config.default_journal_dir or '')`  I think ?

Athanasius:
So a whole bunch of "use `default=...` instead of `or ...`", plus specific comments.

My eyes glazed over on some of the large bits of diff (the guts of e.g. the new Windows code), but I'm trusting it's a copy/paste of the old, just now encapsulated in classes and abstraction.

Likewise I've not looked too closely at the tests stuff yet (I'll run it when this is ready to merge to see if anything drops out).

config.py
A-UNDERSCORE-D: I was going to use regexp's escape, but there are some things it does that are weird on the outside, and yeah other than that they're bare

A-UNDERSCORE-D:
the windows stuff is about 50/50 implementation and copy, however it all passes my tests

plugins/edsm.py
A-UNDERSCORE-D: It'll throw an error if its not a list from the config side itself, otherwise the only other thing it can return in that form is `None` if the field does not exist

plugins/eddn.py
A-UNDERSCORE-D: Updated

plugins/eddn.py
A-UNDERSCORE-D: Whoops didnt mean to resolve that. Yeah I think waiting is best. No need to implement double migration code

outfitting.py
A-UNDERSCORE-D: Updated

loadout.py
A-UNDERSCORE-D: Yep!

config.py
A-UNDERSCORE-D: Do you want me to change that in this PR?

EDMarketConnector.py
A-UNDERSCORE-D: All of these should be fixed

config.py
Athanasius: Yes please, it should be a minimal change.

EDMarketConnector.py
```diff
@@ -351,13 +351,13 @@ def __init__(self, master: tk.Tk):  # noqa: C901, CCR001 # TODO - can possibly f
             self.menubar.add_cascade(menu=self.help_menu)
             if platform == 'win32':
                 # Must be added after at least one "real" menu entry
-                self.always_ontop = tk.BooleanVar(value=cast(int, config.get_int('always_ontop')))
-                self.system_menu = tk.Menu(self.menubar, name='system', tearoff=tk.FALSE)  # type: ignore # https://github.com/python/typeshed/issues/4658 # noqa: E501
+                self.always_ontop = tk.BooleanVar(value=config.get_int('always_ontop')))
+                self.system_menu=tk.Menu(self.menubar, name = 'system', tearoff = tk.FALSE)  # type: ignore # https://github.com/python/typeshed/issues/4658 # noqa: E501
```
Athanasius, line 355: I thought the proper way was "foo=val" in arguments, and "foo = val" in assignments otherwise?  You're flipping a load to the opposite of that.

plugins/eddn.py
```diff
@@ -714,7 +714,7 @@ def filter_localised(d: Mapping[str, Any]) -> OrderedDictT[str, Any]:
                 this.commodities = this.outfitting = this.shipyard = None
                 this.marketId = entry['MarketID']
 
-            path = pathlib.Path(str(config.get_str('journaldir') or config.default_journal_dir)) / f'{entry["event"]}.json'
+            path = pathlib.Path(config.get_str('journaldir', str(config.default_journal_dir))) / f'{entry["event"]}.json'
```
Athanasius, line 717: default=

EDMarketConnector.py
```diff
@@ -925,25 +925,25 @@ def dashboard_event(self, event) -> None:
...
         """Despatch a ship URL to the configured handler."""
-        if not bool(config.getint("use_alt_shipyard_open")):
-            return plug.invoke(config.get('shipyard_provider'), 'EDSY', 'shipyard_url', monitor.ship(), monitor.is_beta)
+        if not bool(config.get_int("use_alt_shipyard_open")):
+            return plug.invoke(config.get_str('shipyard_provider'), 'EDSY', 'shipyard_url', monitor.ship(), monitor.is_beta)
 
         # Avoid file length limits if possible
-        provider = config.get('shipyard_provider') or 'EDSY'
-        target = plug.invoke(config.get('shipyard_provider'), 'EDSY', 'shipyard_url', monitor.ship(), monitor.is_beta)
+        provider = config.get_str('shipyard_provider', 'EDSY')
```
Athanasius, line 945: default=

Athanasius:
I think those comments cover any missing `default=` where a default is specified (I'm assuming where we're not specifying a default in the call we're happy with the, er, default default).  As per:

`egrep -n 'config\.get_[^\)]+\([^\)]+,[^\)]+\)' *.py plugins/*.py | fgrep -v default`

EDMarketConnector.py
A-UNDERSCORE-D: I think theres an extra paren in there or something that caused the formatter to do that. Will fix when Im more awake tomorrow

config.py
A-UNDERSCORE-D: Creates a circular import (which is probably why config doesn't import anything else 

```
  File "./EDMarketConnector.py", line 19, in <module>
    from config import applongname, appname, appversion, appversion_nobuild, config, copyright
  File "/home/ad/development/python/EDMarketConnector/config.py", line 24, in <module>
    import EDMCLogging
  File "/home/ad/development/python/EDMarketConnector/EDMCLogging.py", line 49, in <module>
    from config import appcmdname, appname, config
ImportError: cannot import name 'appcmdname' from partially initialized module 'config' (most likely due to a circular import) (/home/ad/development/python/EDMarketConnector/config.py)
```

Athanasius: I had an after-comment suspicion it might have been such a special case.  We should add a comment to that effect.

PR #791, opened 2020-12-08
Cherry-pick in some key commits for upcoming 4.1.5

Athanasius:
* Change to show 'Shutting down' on exit, rather than withdrawing the UI straight away.  Should make it more obvious if the process hangs during shutdown.
* Likewise add a pop-up if the program is detected as already running at startup, instead of exiting silently.
* Fix for mangled names support in our logging.
* py2exe now from PyPi (paving the way for GitHub builds).


A-UNDERSCORE-D (approved):
LGTM

PR #799, opened 2020-12-09
Added auto-annotation to both flake8 checks

A-UNDERSCORE-D:
This works based on telling github how to understand the flake8 output


.github/workflows/push-checks.yml
```diff
@@ -10,7 +10,7 @@ name: Push-Checks
 on:
   push:
     branches: [ develop ]
-
+
```
Athanasius, line 13: Woo, whitespace-only change!

A-UNDERSCORE-D: Removed :D

PR #803, opened 2020-12-15
EDMCLogging: Catch inspect.getframeinfo(frame) not working

Athanasius:
Maybe with this we'll see something 'here' or later to give a clue as to why this doesn't work on at least two systems.

### DIFF
--- a/EDMCLogging.py
+++ b/EDMCLogging.py
@@ -286,7 +286,7 @@ class EDMCContextFilter(logging.Filter):
             # <https://stackoverflow.com/questions/2203424/python-how-to-retrieve-class-information-from-a-frame-object#2220759>
             try:
                 frame_info = inspect.getframeinfo(frame)
-                raise(IndexError)  # TODO: Remove, only for testing
+                # raise(IndexError)  # TODO: Remove, only for testing
 
             except Exception:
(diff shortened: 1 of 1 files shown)


Athanasius:
This is just the *minimal* try/catch/print something in order to try and get more information about root causes.

Down the line this could become more refined, making best efforts to at least, e.g. return immediate class name as happens without our EDMContextFilter.

EDMCLogging.py
```diff
@@ -284,7 +284,25 @@ class if relevant.
...
+                # We want to *attempt* to show something about the nature of 'frame',
+                # but at this point we can't trust it will work.
+                try:
+                    print(f'frame: {frame}')
+
+                except Exception:
+                    pass
+
+                # We've given up, so just return all '??' to signal we couldn't get the info
+                return '??', '??', '??'
```
A-UNDERSCORE-D, line 304: Somewhere in here we should dump the (textual) traceback as well, if we can

Athanasius: As discussed on Discord, that's something for a later, 'proper' version of this. For now I want the minimum to keep the program running as long as possible, rather than throwing a fresh, un-caught, exception for other reasons in the logging code itself.

PR #812, opened 2020-12-21
Added support for logging from properties

### LINKED ISSUE #808: RecursionError when a property does logging
- Version: Release/4.1.5-4-g4883c1606, i.e. 4883c16 ("stable")
 - OS: Ubuntu 18.04
 - OS Locale: en_GB.UTF-8

[EDMarketConnector-debug.log](https://github.com/EDCD/EDMarketConnector/files/5721107/EDMarketConnector-debug.log)

the interesting bit:

```pytb
  File "/home/ash/.local/share/EDMarketConnector/plugins/edrodent/load.py", line 38, in ready_to_rat
    logger.debug("ready_to_rat: %r %r %r %r %r", self.open, self.low_fuel, self.hud_in_analysis_mode, self.has_fuel_transfer_limpet_controller, self.limpets)
  File "/usr/local/lib/python3.7/logging/__init__.py", line 1366, in debug
    self._log(DEBUG, msg, args, **kwargs)
  File "/usr/local/lib/python3.7/logging/__init__.py", line 1514, in _log
    self.handle(record)
  File "/usr/local/lib/python3.7/logging/__init__.py", line 1523, in handle
    if (not self.disabled) and self.filter(record):
  File "/usr/local/lib/python3.7/logging/__init__.py", line 751, in filter
    result = f.filter(record)
  File "/home/ash/src/EDMarketConnector/EDMCLogging.py", line 255, in filter
    (class_name, qualname, module_name) = self.caller_attributes(module_name=getattr(record, 'module'))
  File "/home/ash/src/EDMarketConnector/EDMCLogging.py", line 317, in caller_attributes
    fn = getattr(frame_class, name, None)
  File "/home/ash/.local/share/EDMarketConnector/plugins/edrodent/load.py", line 38, in ready_to_rat
    logger.debug("ready_to_rat: %r %r %r %r %r", self.open, self.low_fuel, self.hud_in_analysis_mode, self.has_fuel_transfer_limpet_controller, self.limpets)
  File "/usr/local/lib/python3.7/logging/__init__.py", line 1366, in debug
    self._log(DEBUG, msg, args, **kwargs)
  File "/usr/local/lib/python3.7/logging/__init__.py", line 1513, in _log
    exc_info, func, extra, sinfo)
  File "/usr/local/lib/python3.7/logging/__init__.py", line 1483, in makeRecord
    sinfo)
  File "/usr/local/lib/python3.7/logging/__init__.py", line 316, in __init__
    self.filename = os.path.basename(pathname)
  File "/usr/local/lib/python3.7/posixpath.py", line 147, in basename
    sep = _get_sep(p)
  File "/usr/local/lib/python3.7/posixpath.py", line 42, in _get_sep
    if isinstance(path, bytes):
RecursionError: maximum recursion depth exceeded while calling a Python object
```

**Describe the bug**
If a property attempts to log anything, the EDMC logging infrastructure tries to find the property's `__qualname__`, but inadvertently ends up evaluating the property instead. If re-executing the property means another logging call, then this will end up in a RecursionError. If not, EDMC will compute the incorrect `__qualname__` or will fail to determine it at all (depending on what the property returns).

**To Reproduce**
Something like this:

```python
class C:
    @property
    def name(self):
        logger.info("getting plugin name")  # <-- recursion happens here
        return "foo"
def plugin_start3(plugin_dir):
    return C().name
```

**Expected behavior**
Logging inside a property should ideally do something sensible, or the fact that it won't work should be documented.

[`inspect.getattr_static`](https://docs.python.org/3/library/inspect.html#fetching-attributes-statically) looks like it would be useful here.

A-UNDERSCORE-D:
This is some best effort support for using logging in properties.

This works by using the (as suggested by reporter) inspect
`getattr_static` method, and failing that (as it can possibly fail),
wrapping a `getattr` in a try/catch for a RecursionError--don't want to
catch other things, probably best if that explodes on its own.

From there as the `property` object will not have location information,
we rebuild as best we can to an approximation of what the path would be.
With a healthy dash of defensive programming "Just in case".

I don't think that this will have any adverse effects to other logging
methods, as all the new code should only be touched if we hit a property
object.

Closes #808

### DIFF
--- a/EDMCLogging.py
+++ b/EDMCLogging.py
@@ -296,9 +296,32 @@ class EDMCContextFilter(logging.Filter):
                         name = f'_{frame_class.__class__.__name__}{frame_info.function}'
 
                     # Find __qualname__ of the caller
-                    fn = getattr(frame_class, name, None)
-                    if fn and fn.__qualname__:
-                        caller_qualname = fn.__qualname__
+                    fn = inspect.getattr_static(frame_class, name, None)
+                    if fn is None:
+                        # For some reason getattr_static cant grab this. Try and grab it with getattr, bail out
+                        # if we get a RecursionError indicating a property
+                        try:
+                            fn = getattr(frame_class, name, None)
+                        except RecursionError:
+                            print(
+                                "EDMCLogging:EDMCContextFilter:caller_attributes():"
+                                "Failed to get attribute for function info. Bailing out"
+                            )
+                            return "??", "??", "??"
+
+                    if fn is not None:
+                        if isinstance(fn, property):
+                            class_name = str(frame_class)
+                            # If somehow you make your __class__ or __class__.__qualname__ recursive, I'll be impressed.
+                            if hasattr(frame_class, '__class__') and hasattr(frame_class.__class__, "__qualname__"):
+                                class_name = frame_class.__class__.__qualname__
+                                caller_qualname = f"{class_name}.{name}(property)"
+
+                            else:
+                                caller_qualname = f"<property {name} on {class_name}>"
+
+                        elif fn.__qualname__:
+                            caller_qualname = fn.__qualname__
 
                     # Find containing class name(s) of caller, if any
                     if frame_class.__class__ and frame_class.__class__.__qualname__:
--- a/EDMarketConnector.py
+++ b/EDMarketConnector.py
@@ -1238,7 +1238,6 @@ executable: {sys.executable}
 sys.path: {sys.path}'''
                  )
 
-
     # We prefer a UTF-8 encoding gets set, but older Windows versions have
     # issues with this.  From Windows 10 1903 onwards we can rely on the
     # manifest ActiveCodePage to set this, but that is silently ignored on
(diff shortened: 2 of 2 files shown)


EDMCLogging.py
```diff
@@ -296,9 +296,32 @@ class if relevant.
...
+                                "EDMCLogging:EDMCContextFilter:caller_attributes():"
+                                "Failed to get attribute for function info. Bailing out"
+                            )
+                            return "??", "??", "??"
+
+                    if fn is not None:
+                        if isinstance(fn, property):
+                            class_name = str(frame_class)
+                            # If somehow you make your __class__ or __class__.__qualname__ recursive, I'll be impressed.
+                            if hasattr(frame_class, '__class__') and hasattr(frame_class.__class__, "__qualname__"):
```
Athanasius, line 316: If you feel the need to check for both here, then presumably one of them could be set but not the other?  In which case separate the conditionals so we *get* the one we can.  Or are we just being extra paranoid with the conditional ?

A-UNDERSCORE-D: Extra paranoid, if `__class__` exists it *probably* has a `__qualname__`

PR #819, opened 2020-12-30
Fix multiple missions with the same cargo type

### LINKED ISSUE #817: Unique cargo of the same commodity type is only shown once
**Please check the [Known Issues](https://github.com/EDCD/EDMarketConnector/issues/618) in case this has already been reported.**

**Please also check if the issue is covered in our [Troubleshooting Guide](https://github.com/EDCD/EDMarketConnector/wiki/Troubleshooting).** It might be something with a known work around, or where a third party (such as EDSM) is causing logging that is harmless.

*It does not appear to be in either*

 - Version 4.1.5
 - OS: Windows 10
 - OS Locale: English
 - Please attach **BOTH** log files as follows:
     1. `%TEMP%\EDMarketConnector.log` from *immediately* after the bug occurs (re-running the application overwrites this file).  [EDMarketConnector.log](https://github.com/EDCD/EDMarketConnector/files/5745960/EDMarketConnector.log)
     1. %TEMP%\EDMarketConnector\EDMarketConnector-debug.log`.  See [Debug Log File](https://github.com/EDCD/EDMarketConnector/wiki/Troubleshooting#debug-log-files).  [EDMarketConnector-debug.log](https://github.com/EDCD/EDMarketConnector/files/5745981/EDMarketConnector-debug.log)

**Describe the bug**
When carrying unique cargo for multiple missions of the same commodity type (i.e. two missions require delivering Food Cartridges) only one of these will be reported.

**To Reproduce**
1. Accept a mission requiring delivering unique cargo, pick up the cargo
2. Accept a second mission requiring the same type of unique cargo, pick it up too
3. Only one of these groups of cargo will be reported in EDMC

**Expected behavior**
Both unique cargo are recorded, either in bulk or as separate instances

**Screenshots**
EDMC Bug
EDMC Bug Transactions
EDMC Bug Inventory

**Additional context**
This was found while using the Cargo Manifest plugin, which I modified to output the cargo dictionary in to the log files.  I did restart between noticing the bug and reporting it, but the same behaviour was observed before the restart, and in another case with different missions/cargo.

A-UNDERSCORE-D:
A user can collect multiple missions with the same mission cargo, which
is added to the cargo.json as distinct entries. Previously we would take
the last number as the total, leading to invalid counts (and possibly
negative counts).

This adds a method to sum cargo entries based on the cargo name. It also
adds a field on status to access the original cargo JSON data.

Fixes #817

### DIFF
--- a/PLUGINS.md
+++ b/PLUGINS.md
@@ -412,6 +412,7 @@ Content of `state` (updated to the current journal entry):
 | `ModulesValue` |            `int`            | Value of the current ship's modules                                                                             |
 | `Rebuy`        |            `int`            | Current ship's rebuy cost                                                                                       |
 | `Modules`      |           `dict`            | Currently fitted modules                                                                                        |
+| `CargoJSON`    |           `dict`            | content of cargo.json as of last read.                                                                          |
 
 A special "StartUp" entry is sent if EDMC is started while the game is already
 running. In this case you won't receive initial events such as "LoadGame",
--- a/PLUGINS.md
+++ b/PLUGINS.md
@@ -392,7 +392,8 @@ Content of `state` (updated to the current journal entry):
 | Field          |            Type             | Description                                                                                                     |
 | :------------- | :-------------------------: | :-------------------------------------------------------------------------------------------------------------- |
 | `Captian`      |       `Optional[str]`       | Name of the commander who's crew you're on, if any                                                              |
-| `Cargo`        |           `dict`            | Current cargo                                                                                                   |
+| `Cargo`        |           `dict`            | Current cargo. Note that this will be totals, and any mission specific duplicates will be counted together      |
+| `CargoJSON`    |           `dict`            | content of cargo.json as of last read.                                                                          |
 | `Credits`      |            `int`            | Current credits balance                                                                                         |
 | `FID`          |            `str`            | Frontier commander ID                                                                                           |
 | `Loan`         |       `Optional[int]`       | Current loan amount, if any                                                                                     |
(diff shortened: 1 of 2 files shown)


Athanasius:
Needs a PLUGINS.md update as well.  Both to explicitly call out that state['Cargo'] will only hold *totals*, and that there's the new key to get at the last known detail *from Cargo.json*.

A-UNDERSCORE-D:
Updated with some docs. Think I should add a section stating what `state` contains in its entirety? could be helpful in future

Athanasius:
> Updated with some docs. Think I should add a section stating what `state` contains in its entirety? could be helpful in future

There is such in PLUGINS.md already, and it's what I was thinking would be updated/added to.

A-UNDERSCORE-D:
Added

PLUGINS.md
```diff
@@ -412,6 +412,7 @@ Content of `state` (updated to the current journal entry):
 | `ModulesValue` |            `int`            | Value of the current ship's modules                                                                             |
 | `Rebuy`        |            `int`            | Current ship's rebuy cost                                                                                       |
 | `Modules`      |           `dict`            | Currently fitted modules                                                                                        |
+| `CargoJSON`    |           `dict`            | content of cargo.json as of last read.                                                                          |
```
Athanasius, line 415: Shift this up to just under the 'Cargo' one, and call out in the 'Cargo' one that it's the *totals*.

Athanasius:
This looks good, but whilst needs to be mergeable into the next release shouldn't actually be merged into release-4.1.5.  So I'll pull it locally and do a local merge against main (rather than wait for our sleep patterns to line up enough again).

Athanasius:
Merged into `main` manually.

PR #831, opened 2021-01-06
Fix logging from mangled methods where class name starts with "_"

### LINKED ISSUE #764: Logging throws an exception if the caller is a name-mangled method
 - Version: 4.1.3 or so, `commit f3e8dbf7c1239c334dfff5bb74240e219776c258 (HEAD -> refs/heads/stable, refs/remotes/origin/stable, refs/remotes/origin/main, refs/remotes/origin/HEAD)`
 - OS: Linux Ubuntu 18.04
 - OS Locale: en_GB.UTF-8

**Describe the bug**
If you call e.g. `logger.info` from within a function with a mangled name (a class method starting with two underscores and ending with zero or one underscores, e.g. `__server`), EDMC's logging infrastructure tries to look up the function on the class and fails:

```pytb
Exception in thread Thread-2:
  File "/usr/local/lib/python3.7/threading.py", line 926, in _bootstrap_inner
    self.run()
  File "/usr/local/lib/python3.7/threading.py", line 870, in run
  File "/home/ash/.local/share/EDMarketConnector/plugins/edmcoverlay/_edmcoverlay.py", line 94, in __server
    logger.info("edmcoverlay2: server running")
  File "/usr/local/lib/python3.7/logging/__init__.py", line 1378, in info
    self._log(INFO, msg, args, **kwargs)
  File "/usr/local/lib/python3.7/logging/__init__.py", line 1514, in _log
    self.handle(record)
  File "/usr/local/lib/python3.7/logging/__init__.py", line 1523, in handle
    if (not self.disabled) and self.filter(record):
  File "/usr/local/lib/python3.7/logging/__init__.py", line 751, in filter
    result = f.filter(record)
  File "/home/ash/src/EDMarketConnector/EDMCLogging.py", line 255, in filter
    (class_name, qualname, module_name) = self.caller_attributes(module_name=getattr(record, 'module'))
  File "/home/ash/src/EDMarketConnector/EDMCLogging.py", line 294, in caller_attributes
    fn = getattr(frame_class, frame_info.function)
AttributeError: 'Overlay' object has no attribute '__server'
```

**To Reproduce**
1. Install [edmcoverlay2](https://github.com/sersorrel/edmcoverlay2)
2. Run EDMC

**Expected behavior**
EDMC should support logging from class methods with mangled names.

sersorrel:
#766 didn't completely fix #764, unfortunately: logging from mangled methods inside classes whose names started with an underscore is still broken (it no longer crashes, just fails to determine the name). This PR fixes that.

Python's name-mangling rules are _weird_:

```python
# test.py

class Foo:
    def method(self): pass
    def __mangled_method(self): pass

class _Foo:
    def method(self): pass
    def __mangled_method(self): pass

class ___:
    def method(self): pass
    def __mangled_method(self): pass

for cls in [Foo, _Foo, ___]:
    print(cls.__name__, [attr for attr in dir(cls()) if "method" in attr])
```

```console
$ python3 test.py
Foo ['_Foo__mangled_method', 'method']
_Foo ['_Foo__mangled_method', 'method']
___ ['__mangled_method', 'method']
```

This PR is still arguably incomplete, because it incorrectly applies mangling to names with dots in, but that seems unlikely to be an issue for EDMC – as noted in [`_Py_Mangle`](https://github.com/python/cpython/blob/ee9f98d9f4b881ee15868a836a2b99271df1bc0e/Python/compile.c#L255), "The only time a name with a dot can occur is when we are compiling an import statement that has a package name", and this doesn't seem like something that the EDMC logging infrastructure will have to deal with.

### DIFF
--- a/EDMCLogging.py
+++ b/EDMCLogging.py
@@ -322,9 +322,11 @@ class EDMCContextFilter(logging.Filter):
 
                 if frame_class:
                     # See https://en.wikipedia.org/wiki/Name_mangling#Python for how name mangling works.
+                    # For more detail, see _Py_Mangle in CPython's Python/compile.c.
                     name = frame_info.function
-                    if name.startswith("__") and not name.endswith("__"):
-                        name = f'_{frame_class.__class__.__name__}{frame_info.function}'
+                    class_name = frame_class.__class__.__name__.lstrip("_")
+                    if name.startswith("__") and not name.endswith("__") and class_name:
+                        name = f'_{class_name}{frame_info.function}'
 
                     # Find __qualname__ of the caller
                     fn = inspect.getattr_static(frame_class, name, None)


A-UNDERSCORE-D:
And here I thought it'd be a simple fix.
At this point I feel like we may want to pull this out to a separate function operating on the class object itself, and throw a wall of tests at it. That said, this looks good to me. Flake8 fails are unrelated.

A-UNDERSCORE-D (approved):
Forgot the magic green check

Athanasius:
Checking that cpython code, the check for a dot in the name looks easy enough to add ?  Extra paranoia.

Mostly, having reviewed `pydoc inspect` I'm now annoyed that we're having to get at the function object *via* the textual name, rather than some more robust method.  At this point we don't actually 'care' about the function name other than using it in `fn = inspect.getattr_static(frame_class, name, None)` (it is used later on in the "why do properties have to be awkward?" code).

A-UNDERSCORE-D:
> Checking that cpython code, the check for a dot in the name looks easy enough to add ? Extra paranoia.
> 
> Mostly, having reviewed `pydoc inspect` I'm now annoyed that we're having to get at the function object _via_ the textual name, rather than some more robust method. At this point we don't actually 'care' about the function name other than using it in `fn = inspect.getattr_static(frame_class, name, None)` (it is used later on in the "why do properties have to be awkward?" code).

Well technically we dont have to at all, we could iterate the `__dict__` and do comparison checking, I'd assume. but that falls afoul of anything that uses a `__getattr__` or `__getattribute__` override. So looking up the stack could work for the function object, maybe. I'd have to check, but there be dragons so I'm not sure we want to go that way.

Issue #834, opened 2021-01-08
Cleanup monitor shutdown code


Athanasius:
As per https://github.com/EDCD/EDMarketConnector/issues/678#issuecomment-756463931 we at least have some code running when it shouldn't during shutdown.

1. Move the
```
            # Check whether we're still supposed to be running
            if threading.current_thread() != self.thread:
                logger.info("We're not meant to be running, exiting...")
                return  # Terminate
```
to the top of the `while True` it's in.  Although it being directly after the sleep() **should** mean this is moot.  We might also want to be more paranoid in the main logfile reading loop.  That *should* exit when it hits end of file, then cause the sleep and checks and going around the loop again.  But what if we trigger shutdown whilst it's in the middle of reading a new set of lines ?

2. As monitor.stop() does things like `self.currentdir = None` we should check for that kind of state in any other monitor related code that could be triggered outside of the monitor.worker() thread.  In theory we should be able to trigger off of `self.thread is None`.  Candidates are `EDMarketConnector.AppWindow.journal_event()` and `monitor.get_entry()`.
3. Check for any old style debug printing and convert to logging. e.g. in monitor.stop()
```
        if __debug__:
            print('Stopping monitoring Journal')
```

A-UNDERSCORE-D:
Possibly do 1 in both places? rather than hope we catch both sides?

Athanasius:
4. We might want to re-consider the order of shutdown calls in EDMarketConnector.py.  Mostly, do we perhaps want to shutdown plugins *first* ?  Doing them first risks something in their shutdown code hanging, but it's not like the core EDMC code will be needing to save any persistent state.  Doing them later risks them depending on something we've already shutdown.

A-UNDERSCORE-D:
plugin teardown being first makes sense logically, obviously we can have the hang as you mentioned. But it leaves us an opening to do other teardown that depends on plugins not existing (Not that we have anything as of yet, but it means we can make assumptions about messing with `monitor`, for example).

Honestly what we could also do, to be "sure" is to start up a shutdown thread that forces us to die (in some way, I havent really thought about how :D perhaps asking the OS to kill us?) if we're still hanging around after x seconds

Athanasius:
> Honestly what we could also do, to be "sure" is to start up a shutdown thread that forces us to die (in some way, I havent really thought about how :D perhaps asking the OS to kill us?) if we're still hanging around after x seconds

[...]

PR #835, opened 2021-01-08
Bail out of Journal monitoring during shutdown

Athanasius:
See #834 - we need to be better about ceasing journal parsing during application shutdown.

This might go on to also re-ordering the shutdown sequence.

### DIFF
--- a/dashboard.py
+++ b/dashboard.py
@@ -1,16 +1,13 @@
 import json
 from calendar import timegm
-from operator import itemgetter
-from os import listdir
 from os.path import isdir, isfile, join, getsize
 from sys import platform
 import time
 
-if __debug__:
-    from traceback import print_exc
-
 from config import config
+from EDMCLogging import get_main_logger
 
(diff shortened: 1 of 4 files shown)


dashboard.py
```diff
@@ -67,8 +64,10 @@ def start(self, root, started):
         if not self.observed and not polling:
             self.observed = self.observer.schedule(self, self.currentdir)
 
-        if __debug__:
-            print('%s Dashboard "%s"' % (polling and 'Polling' or 'Monitoring', self.currentdir))
+        if polling:
+            logger.debug(f'Polling Dashboard "{self.currentdir}"')
+        else:
+            logger.debug(f'Monitoring Dashboard "{self.currentdir}"')
```
A-UNDERSCORE-D, line 70: Do you want these explicitly to be double quoted? using `!r` would quote them and also immediately indicate if something isnt a string

A-UNDERSCORE-D (approved):

PR #839, opened 2021-01-11
Renamed various config values to be backwards compatible

A-UNDERSCORE-D:


config.py
```diff
@@ -341,7 +341,7 @@ def __init__(self, do_winsparkle=True) -> None:
         self.identifier = applongname
         if (outdir_str := self.get_str('outdir')) is None or not pathlib.Path(outdir_str).is_dir():
             docs = known_folder_path(FOLDERID_Documents)
-            self.set('outdir',  docs if docs is not None else str(self.home))
+            self.set('outdir',  docs if docs is not None else str(self.home_path))
```
Athanasius, line 344: This could just be `self.home` now, no `str(...)`, yes ?

config.py
```diff
@@ -702,7 +702,7 @@ def __init__(self, filename: Optional[str] = None) -> None:
             self.config.add_section(self.SECTION)
 
         if (outdir := self.get_str('outdir')) is None or not pathlib.Path(outdir).is_dir():
-            self.set('outdir', str(self.home))
+            self.set('outdir', str(self.home_path))
```
Athanasius, line 705: Again just `self.home`

Athanasius:
Just those two `self.home` bits then.

A-UNDERSCORE-D:
Whoops.
Both of those resolved

PR #840, opened 2021-01-11
Avoid Tk event_generate() calls during shutdown.

Athanasius:
1. Adds a Config class member and helper functions to set and check for shutdown.
2. Sets this flag at the top of AppWindow.onexit().
3. Checks the flag so far only in plug.py show_error().

Still needs a further look into where else calls event_generate() and should have the check.

### DIFF
--- a/config.py
+++ b/config.py
@@ -117,7 +117,7 @@ class Config(object):
     if platform=='darwin':
 
         def __init__(self):
-            self.in_shutdown = False  # Is the application currently shutting down ?
+            self.__in_shutdown = False  # Is the application currently shutting down ?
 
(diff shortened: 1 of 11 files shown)


config.py
```diff
@@ -177,9 +179,16 @@ def close(self):
             self.save()
             self.defaults = None
 
+        def set_shutdown(self):
+            self.__in_shutdown = True
+
+        def shutting_down(self) -> bool:
+            return self.__in_shutdown
+
```
A-UNDERSCORE-D, line 187: May as well make this a property so access isnt `config.shutting_down()`

Athanasius: Done

A-UNDERSCORE-D (approved):
Works for me on Linux, and I assume your testing was on windows

PR #845, opened 2021-01-13
Fix console redirect to work with config.py now using logging.

Athanasius:
1. We were now importing from config before the console redirect code
   runs.
2. That means that config.py's `logger = logging.getLogger(appname)` caused the logger to be set up whilst stdout/err were still pointing at the console.
3. So the redirect then had no effect on logging output.

This commit moves the import from config (and also killswitch, because it also imports logging) to after the *moved* console redirect code.

NB: This needs the "detected a process already running" instance checking.

NB: There's still a `True or ` for testing purposes EDMarketConnector.py:25

### DIFF
--- a/EDMarketConnector.py
+++ b/EDMarketConnector.py
@@ -133,7 +133,7 @@ if __name__ == '__main__':
 
     # Keep this as the very first code run to be as sure as possible of no
     # output until after this redirect is done, if needed.
-    if True or getattr(sys, 'frozen', False):
+    if getattr(sys, 'frozen', False):
         # By default py2exe tries to write log to dirname(sys.executable) which fails when installed
         import tempfile
 
--- a/EDMarketConnector.py
+++ b/EDMarketConnector.py
@@ -16,11 +16,10 @@ from sys import platform
 from time import localtime, strftime, time
 from typing import TYPE_CHECKING, Any, Mapping, Optional, Tuple, cast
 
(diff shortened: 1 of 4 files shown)


EDMarketConnector.py
```diff
@@ -12,12 +12,28 @@
...
 from typing import TYPE_CHECKING, Any, Mapping, Optional, Tuple, cast
 
-from config import applongname, appname, appversion, appversion_nobuild, config, copyright
+# config will now cause an appname logger to be set up, so we need the
+# console redirect before this
+appname = 'EDMarketConnector'  # TODO: Must match config.appname, add test
+if __name__ == '__main__':
+    # Keep this as the very first code run to be as sure as possible of no
+    # output until after this redirect is done, if needed.
+    if True or getattr(sys, 'frozen', False):
```
Athanasius, line 25: This `True or ` needs removing before this is merged.

A-UNDERSCORE-D (approved):
Other than removing the `True or`, looks good

EDMarketConnector.py
```diff
@@ -132,14 +133,20 @@ def already_running_popup():
 
     # Keep this as the very first code run to be as sure as possible of no
     # output until after this redirect is done, if needed.
-    if getattr(sys, 'frozen', False):
+    if True or getattr(sys, 'frozen', False):
```
Athanasius, line 136: ` True or` to be removed

Issue #809, opened 2020-12-20
Investigate why the multi-process check stops `runas` from working.


Athanasius:
We document that using `runas` to run EDMarketConnector.exe as another user is the way to run more than one instance in a sane manner.

I tried it, the second instance detected the first and refused to fully run.

The wording of 25e5f27c3a4c7b537cd08b9d0fa52d6add2facf5 implies it's meant to only be checking within the current user, does it ?  If so, why isn't that working ?

Athanasius:
Some more notes.

If I first run EDMarketConnector.exe as my normal Limited user, and then:

runas /profile /user:<an admin user> <path to>\EDMarketConnector.exe`

from that Limited logged in user.  It doesn't work, I get the popup complaining that the program is already running.

But if I actually login separately as the Admin user (whilst my Limited user is still logged in and running EDMarketConnector.exe) then I can run another instance of EDMarketConnector.exe as that other user.

Athanasius:
Dropping the `/profile` doesn't work, nor does changing it to `/noprofile`.

Athanasius:
I've only glanced at the code, but it seems to me that it's checking if any window, in the currently logged in user's desktop environment, has a handle matching what we expect EDMarketConnector.exe to have.  i.e. it's NOT actually "check if there's already an EDMarketConnector.exe process for this user".

Perhaps we should see if there's a better way of performing that check.

Athanasius:
https://github.com/EDCD/EDMarketConnector/releases/tag/Release%2F4.1.7-rc1 contains the changes to this if anyone wants to test.

Athanasius:
This is further addressed by code I just merged into `main`, which is intended for a `4.2.0-beta1` in the next few days (possibly even today, pending a few other small bits of work).

Athanasius:
There's now https://github.com/EDCD/EDMarketConnector/releases/tag/Release%2F4.2.0-beta3 which I've confirmed has this fully working.

PR #852, opened 2021-01-18
Switch to using a journals folder lockfile to enforce single process

Athanasius:
So as to actually allow `runas` this changes the win32 "Is there any window with the expected handle already?" check into a lock on a file in the `journals_dir` instead.

This also means that we can actually enforce this on all platforms.  win32 doesn't have fcntl so we use msvcrt there, but we assume (with a try/except) that all other platforms will have fcntl.  Worst case scenario they're in the same situation as before, no check at all for more than one process.

NB: This is going to be tricky when forward-ported into `develop` due to absolutely needing to use config.get() to obtain the `journals_dir`.  In `develop` that will cause logging to be set up before it should.  Might have to add something to explicitly close all logging so that setting it up again later correctly follows the stdout/err redirect.

### DIFF
--- a/EDMarketConnector.py
+++ b/EDMarketConnector.py
@@ -15,11 +15,25 @@ from sys import platform
 from time import localtime, strftime, time
 from typing import TYPE_CHECKING
 
-from config import appversion, appversion_nobuild, config, copyright
 from constants import applongname, appname, protocolhandler_redirect
 
 # config will now cause an appname logger to be set up, so we need the
 # console redirect before this
+if __name__ == '__main__':
+    # Keep this as the very first code run to be as sure as possible of no
+    # output until after this redirect is done, if needed.
+    if getattr(sys, 'frozen', False):
+        # By default py2exe tries to write log to dirname(sys.executable) which fails when installed
+        import tempfile
+
+        # unbuffered not allowed for text in python3, so use `1 for line buffering
+        sys.stdout = sys.stderr = open(join(tempfile.gettempdir(), f'{appname}.log'), mode='a', buffering=1)
+    # TODO: Test: Make *sure* this redirect is working, else py2exe is going to cause an exit popup
+
+# After the redirect in case config does logging setup
+from config import appversion, appversion_nobuild, config, copyright
+from EDMCLogging import edmclogger, logger, logging
+
 if __name__ == '__main__':
     # Command-line arguments
     parser = argparse.ArgumentParser(
@@ -165,16 +179,9 @@ if __name__ == '__main__':
         root.mainloop()
 
     if not no_other_instance_running():
-        # There's a copy already running.  We want to inform the user by
-        # **appending** to the log file, not truncating it.
-        if getattr(sys, 'frozen', False):
-            # By default py2exe tries to write log to dirname(sys.executable) which fails when installed
-            import tempfile
+        # There's a copy already running.
 
-            # unbuffered not allowed for text in python3, so use `1 for line buffering
-            sys.stdout = sys.stderr = open(join(tempfile.gettempdir(), f'{appname}.log'), mode='a', buffering=1)
-
-        # Logging isn't set up yet.
+        # Logging isn't set up yet. # TODO: Actually it will be by now in `develop`
         # We'll keep this print, but it will be over-written by any subsequent
         # write by the already-running process.
         print("An EDMarketConnector.exe process was already running, exiting.")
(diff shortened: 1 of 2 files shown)


Athanasius:
NB: I specifically chose *not* to make any attempt to unlock the file or close that file handle *so that* the process exiting takes care of it naturally.  As well as being less code this also demonstrates that Windows isn't going to keep the lock there and trip over it if the application is killed, or even just exits naturally.

Athanasius:
This will address #809

EDMarketConnector.py
```diff
@@ -18,74 +18,69 @@
...
+                return False
+
+            journal_dir_lockfile.write(f"Path: {journal_dir}\nPID: {os_getpid()}\n")
+
+        else:
+            print('no_other_instance_running(): NOT win32, using fcntl')
+            try:
+                import fcntl
+
+            except Exception:
```
A-UNDERSCORE-D, line 63:
```suggestion
            except ImportError:
```
Rather only catch the "It doesnt exist" case

EDMarketConnector.py
```diff
@@ -18,74 +18,69 @@
...
+                print(f"BlockingIOError: Couldn't lock journal directory \"{journal_dir}\", assuming another process running\n{e}")
+                return False
+
+            except OSError as e:
+                print(f"OSError: Couldn't lock journal directory \"{journal_dir}\", assuming another process running\n{e}")
+                return False
+
+            except Exception as e:
+                print(f"other Exception: Couldn't lock journal directory \"{journal_dir}\", assuming another process running\n{e}")
+                return False
```
A-UNDERSCORE-D, line 79:
```suggestion
            except (BlockingIOError, OSError, Exception) as e:
                print(f'{type(e).__name__}: Couldn't lock journal directory "{journal_dir}", assuming another process running\n{e}')
                return False
```

EDMarketConnector.py
```diff
@@ -18,74 +18,69 @@
 from config import applongname, appname, appversion, appversion_nobuild, config, copyright
 
 if __name__ == "__main__":
+    journal_dir: str = config.get('journaldir') or config.default_journal_dir
+    # This must be at top level to guarantee the file handle doesn't go out
+    # of scope and get cleaned up, removing the lock with it.
+    journal_dir_lockfile = open(join(journal_dir, 'edmc-journal-lock.txt'), mode='w+', encoding='utf-8')
+
     def no_other_instance_running() -> bool:  # noqa: CCR001
```
A-UNDERSCORE-D, line 26: Does this still need to be as a closure? also that noqa can probably be removed now

A-UNDERSCORE-D (commented):
Few style / cleaner options

EDMarketConnector.py
A-UNDERSCORE-D: Whoops, not a closure, just in the if, but still, probably doesnt need to be in there anymore

EDMarketConnector.py
Athanasius: This was paranoia brought on by the initial problems I had getting it working at all.  Given *every* other platform will use fcntl version, and I have no way to test it on e.g. MacOS, I wanted to leave this in so it goes from more specific to less to be sure we print *something* for when it doesn't work as expected.

EDMarketConnector.py
Athanasius: I'm not sure how this *is* a closure.

However, it is true that the function definitions could probably be taken out of this `__main__` clause at least.  On the other hand they *are* only used within it.

EDMarketConnector.py
A-UNDERSCORE-D: Yeah, I just thought that three different except branches that all do the same thing is silly. Its still catching all three, and still prints the exception name as it goes, just all in one rather than as three

EDMarketConnector.py
```diff
@@ -41,74 +57,115 @@
...
-                    else:
-                        ShowWindowAsync(window_handle, SW_RESTORE)
-                        SetForegroundWindow(window_handle)
-
-                    return False
-
-                return True
-
-            return EnumWindows(enumwindowsproc, 0)
+            print('no_other_instance_running(): win32, using msvcrt')
```
A-UNDERSCORE-D, line 74: All of these prints can probably be switched to trace logs

A-UNDERSCORE-D (approved):
LGTM, other than the aforementioned prints

Athanasius:
I just had one horrible thought.  Someone might be sharing their Journals read-only to where they run EDMC.  We'll fail to even create the lockfile in that case.

EDMarketConnector.py
```diff
@@ -41,74 +57,115 @@
 
     args = parser.parse_args()
 
+    journal_dir: str = config.get('journaldir') or config.default_journal_dir
+    # This must be at top level to guarantee the file handle doesn't go out
+    # of scope and get cleaned up, removing the lock with it.
+    journal_dir_lockfile = open(join(journal_dir, 'edmc-journal-lock.txt'), mode='w+', encoding='utf-8')
```
Athanasius, line 63: Need a try/except here for the case of Journal files being shared read-only.  In that case we just assume we're the only process running and keep going.  That will be no different to running on !win32 in current `main`.

EDMarketConnector.py
```diff
@@ -123,6 +123,19 @@ def window_title(h):
...
+                    Determine if any window for the Application exists.
+
+                    Called for each found window by EnumWindows().
+
+                    When a match is found we check if we're being invoked as the
+                    edmc://auth handler.  If so we send the message to the existing
+                    process/window.  If not we'll raise that existing window to the
+                    foreground.
+                    :param window_handle: Window to check.
+                    :param l_param: ???
```
A-UNDERSCORE-D, line 137:
```
Type: LPARAM

An application-defined value to be passed to the callback function.
```
https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-enumwindows
May as well note both of those

A-UNDERSCORE-D (approved):

PR #858, opened 2021-01-22
Port in the journals_dir locking and other changes

Athanasius:
This ports in various recent changes to `main`.  `develop` has diverged too much for this to be easy via a merge or rebase.

* Use a lock file in the journals_dir location to prevent more than one instance running against the same journals.  We no longer check just for a Windows handle.  So this is more correct on win32 *and* is now a thing on all other platforms.
* Adds `--suppress-dupe-process-popup` CL arg to suppress "we're a dupe!" popup to aid those using batch files to launch EDMC alongside the game.
* Two minor fixups of typos in PLUGINS.md.
* Misc noqa comments and other flake8 fixups.  We're now only missing type annotations in EDMarketConnector.py.

### DIFF
--- a/EDMarketConnector.py
+++ b/EDMarketConnector.py
@@ -68,6 +68,11 @@ if __name__ == '__main__':  # noqa: C901
                         action='store_true'
                         )
 
+    parser.add_argument('--force-localserver-for-auth',
+                        help='Force EDMC to use a localhost webserver for Frontier Auth callback',
+                        action='store_true'
+                        )
+
     args = parser.parse_args()
 
     if args.trace:
@@ -75,6 +80,9 @@ if __name__ == '__main__':  # noqa: C901
         edmclogger.set_channels_loglevel(logging.TRACE)
     else:
         edmclogger.set_channels_loglevel(logging.DEBUG)
+
+    if args.force_localserver_for_auth:
+        config.set_auth_force_localserver()
 
     def no_other_instance_running() -> bool:  # noqa: CCR001
         """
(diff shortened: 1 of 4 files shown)


Athanasius:
Once https://github.com/EDCD/EDMarketConnector/pull/857 is approved I'll add those changes to this PR as well.

Athanasius:
This now needs to wait on fixing up the `main` version to handle when the user changes the journals directory.

Athanasius:
> This now needs to wait on fixing up the `main` version to handle when the user changes the journals directory.

Actually, as I seem to have gotten distracted with other things, this should just get merged now.

EDMarketConnector.py
```diff
@@ -11,91 +11,218 @@
...
+
+        # unbuffered not allowed for text in python3, so use `1 for line buffering
+        sys.stdout = sys.stderr = open(join(tempfile.gettempdir(), f'{appname}.log'), mode='wt', buffering=1)
+    # TODO: Test: Make *sure* this redirect is working, else py2exe is going to cause an exit popup
+
+# These need to be after the stdout/err redirect because they will cause
+# logging to be set up.
+# isort: off
+import killswitch
+from config import appversion, appversion_nobuild, config, copyright
```
A-UNDERSCORE-D, line 39: Do we want to switch some of these to be `constants` imports?

A-UNDERSCORE-D (approved):

EDMarketConnector.py
Athanasius: At some point, yes.  But that would be for another PR.

In general this PR is just to get things in sync with what's in `main` for these changes.  I'd prefer your review be limited to finding any actual mistakes ;) .

PR #859, opened 2021-01-23
Update journal lock if user changes journal dir

Athanasius:
The user can change the location of the Journal files (mostly for if they're running EDMC on another host).

We need to update the lockfile if they do this.


Athanasius:
This work is now being done in https://github.com/EDCD/EDMarketConnector/compare/enhancement/859-journal-lock-on-change

PR #896, opened 2021-03-09
Removed remaining star imports

A-UNDERSCORE-D:
fixes #984

### DIFF
--- a/theme.py
+++ b/theme.py
@@ -6,26 +6,25 @@
 #
 
 import os
-from sys import platform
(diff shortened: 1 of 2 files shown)


Athanasius:
Running on Windows under PyCharm:

1. Opening a ship build still works, no errors logged.
1. I can switch to transparent theme, which uses the EUROCAPS font, the loading of which is what the theme.py code seems to be doing.

Approved.

PR #897, opened 2021-03-10
Add support for NavRoute event and file contents.

Athanasius:
Journal doc v28 is incorrect about the word for the event and file being `Route`, it is `NavRoute` (tested in-game).

We augment the `NavRoute` event with the data from the file, so plugins can just look at the event rather than needing to check monitor.state[`NavRoute`], although the data is also there for later reference.

There is **no indication** if the player clears a route (not in journal or a Status.json flag).

The format of the `Route` array data is also different from the v28 doc.

### DIFF
--- a/monitor.py
+++ b/monitor.py
@@ -126,7 +126,8 @@ class EDLogs(FileSystemEventHandler):  # type: ignore # See below
             'ModulesValue': None,
             'Rebuy':        None,
             'Modules':      None,
-            'CargoJSON':   None,  # The raw data from the last time cargo.json was read
+            'CargoJSON':    None,  # The raw data from the last time cargo.json was read
+            'Route':        None,  # Last plotted route from Route.json file
         }
 
     def start(self, root: 'tkinter.Tk'):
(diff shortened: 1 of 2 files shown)


monitor.py
```diff
@@ -665,6 +667,12 @@ def parse_entry(self, line: str):
 
                 self.state['Cargo'].update({self.canonicalise(x['Name']): x['Count'] for x in clean})
 
+            elif event_type == 'NavRoute':
+                # Added in ED 3.7 - multi-hop route details in Route.json
+                with open(join(self.currentdir, 'NavRoute.json'), 'rb') as rf:  # type: ignore
+                    entry = json.load(rf)
```
A-UNDERSCORE-D, line 673: Dash of defensive programming, just in case.
```suggestion
                    try:
                        entry = json.load(rf)
                    except JsonDecodeError as e:
                        self.state['Route'] = None
                        logger.exception(f'Failed to read JSON from NavRoute.json: {e}')
                    else:
                        self.state['Route'] = entry
```

A-UNDERSCORE-D (approved):

PR #899, opened 2021-03-10
monitor.py: General cleanup work on flake8 & mypy output

Athanasius:
Without making major code changes the aim is to get nothing flagged by flake8 or mypy.

### DIFF
--- a/monitor.py
+++ b/monitor.py
@@ -439,6 +439,8 @@ class EDLogs(FileSystemEventHandler):  # type: ignore # See below
         This augments some events, sets internal state in reaction to many and
         loads some extra files, e.g. Cargo.json, as necessary.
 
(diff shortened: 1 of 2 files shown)


monitor.py
```diff
@@ -37,6 +39,8 @@ def _(x: str) -> str:
     from ctypes.wintypes import BOOL, HWND, LPARAM, LPWSTR
 
     from watchdog.events import FileSystemEventHandler
+    if TYPE_CHECKING:
+        from watchdog.events import FileCreatedEvent
```
Athanasius, line 43: Yes, this is in the win32 section, so type checking will likely need tweaks on MacOS.

There's not even proper Linux stuff for the observer stuff, so it's a no-op there.

A-UNDERSCORE-D (approved):

PR #900, opened 2021-03-11
Add more logging to getint and get methods

A-UNDERSCORE-D:
Also a tiny comment only fix to make mypy happy


config.py
```diff
@@ -253,6 +256,8 @@ def getint(self, key: str, *, default: int = 0) -> int:
         :raises OSError: on windows, if a registry error occurs.
         """
         warnings.warn(DeprecationWarning('getint is Deprecated. Use get_int instead'))
+        logger.debug('Attempt to use Deprecated get_int() method\n' + ''.join(traceback.format_stack()))
```
Athanasius, line 259: This is getint() not get_int().

A-UNDERSCORE-D: whoops

PR #903, opened 2021-03-11
Fixed invalid type annotation

A-UNDERSCORE-D:


Athanasius:
I'm slightly confused as to why GH thinks this has no conflicts when it's a repeat of a commit you put into the branch I just merged.  "Same change as the last commit on `develop` so that's fine" ?

PR #905, opened 2021-03-13
Get MacOS working as well as we can.

Athanasius:
1. pyobjc in requirements.txt
2. Bug with `darwin` config.delete() not respecting `suppress` argument fixed.
3. Only actually JournalLock.release_lock() if we actually hold a lock.  Enables changing Journals location from non-existent to another without a fatal exception.

There's still an issue with the whole hotkey code not working because the application needs a MacOS permission to generally monitor keyboard input.  The old way to grant this has changed in Catalina.  There's "Input Monitoring" but I can't find any way to add an arbitrary app to it.  We might need to find how to get the EDMC code to explicitly make the request so that the user can grant it.

And even then I'm not sure the current way of doing hotkeys on `darwin` is the recommended way on Catalina.  It's going to take a lot of study of the code to discern what is just the tk input monitoring and what is specifici MacOS calls.

### DIFF
--- a/config.py
+++ b/config.py
@@ -581,6 +581,12 @@ class MacConfig(AbstractConfig):
 
     def __raw_get(self, key: str) -> Union[None, list, str, int]:
         res = self._settings.get(key)
+        # On MacOS Catalina, with python.org python 3.9.2 any 'list'
+        # has type __NSCFArray so a simple `isinstance(res, list)` is
+        # False.  So, check it's not-None, and not the other types.
+        #
+        # If we can find where to import the definition of NSCFArray
+        # then we could possibly test against that.
         if res and not isinstance(res, str) and not isinstance(res, int):
             return list(res)
 
--- a/config.py
+++ b/config.py
@@ -587,7 +587,7 @@ class MacConfig(AbstractConfig):
         #
         # If we can find where to import the definition of NSCFArray
         # then we could possibly test against that.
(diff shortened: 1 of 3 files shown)


Athanasius:
NB: There was also something weird with a loooooooooong delay before it popped up Safari for the Frontier Authorisation after I'd changed the Journals location.  That may well not be MacOS-specific though.

config.py
```diff
@@ -581,7 +581,13 @@ def __init__(self) -> None:
...
     def __raw_get(self, key: str) -> Union[None, list, str, int]:
         res = self._settings.get(key)
-        if isinstance(res, list):
+        # On MacOS Catalina, with python.org python 3.9.2 any 'list'
+        # has type __NSCFArray so a simple `isinstance(res, list)` is
+        # False.  So, check it's not-None, and not the other types.
+        #
+        # If we can find where to import the definition of NSCFArray
+        # then we could possibly test against that.
+        if res and not isinstance(res, str) and not isinstance(res, int):
```
A-UNDERSCORE-D, line 590:
```suggestion
        if res is not None and not isinstance(res, str) and not isinstance(res, int):
```
`if res` will be false if res is `[]`

A-UNDERSCORE-D (changes requested):

A-UNDERSCORE-D (approved):

Issue #907, opened 2021-03-13
Modules is a list not a dict on damaged stations


A-UNDERSCORE-D:
This breaks some of our horizons checks. lastStarport dump:

```json
{
  "commodities": [
    {
      "buyPrice": 0,
      "categoryname": "Chemicals",
      "demand": 17157650,
      "demandBracket": 3,
      "id": 128673851,
      "legality": "",
      "locName": "Liquid oxygen",
      "meanPrice": 1412,
      "name": "LiquidOxygen",
      "sellPrice": 3714,
      "statusFlags": [],
      "stock": 0,
      "stockBracket": 0
    },
    {
      "buyPrice": 0,
      "categoryname": "Chemicals",
      "demand": 424288680,
      "demandBracket": 3,
      "id": 128049166,
      "legality": "",
      "locName": "Water",
      "meanPrice": 266,
      "name": "Water",
      "sellPrice": 1574,
      "statusFlags": [],
      "stock": 0,
      "stockBracket": 0
    },
    {
      "buyPrice": 0,
      "categoryname": "Consumer Items",
      "demand": 113832818,
      "demandBracket": 3,
      "id": 128049241,
      "legality": "",
      "locName": "Clothing",
      "meanPrice": 459,
      "name": "Clothing",
      "sellPrice": 79,
      "statusFlags": [],
      "stock": 0,
      "stockBracket": 0
    },
    {
      "buyPrice": 0,
      "categoryname": "Consumer Items",
      "demand": 315378350,
      "demandBracket": 3,
      "id": 128672314,
      "legality": "",
      "locName": "Evacuation Shelter",
      "meanPrice": 535,
      "name": "EvacuationShelter",
      "sellPrice": 3425,
      "statusFlags": [],
      "stock": 0,
      "stockBracket": 0
    },
    {
      "buyPrice": 0,
      "categoryname": "Medicines",
      "demand": 2147483647,
      "demandBracket": 3,
      "id": 128049210,
      "legality": "",
      "locName": "Basic Medicines",
      "meanPrice": 436,
      "name": "BasicMedicines",
      "sellPrice": 4298,
      "statusFlags": [],
      "stock": 0,
      "stockBracket": 0
    },
    {
      "buyPrice": 0,
      "categoryname": "Technology",
      "demand": 531085250,
      "demandBracket": 3,
      "id": 128049226,
      "legality": "",
      "locName": "H.E. Suits",
      "meanPrice": 572,
[...]

PR #916, opened 2021-03-15
plugins/eddn: Add paranoia about data in is_horizons()

Athanasius:
A damaged station has `"modules": []`, so trips over modules.values().

Being paranoid about the other data in there as well.


A-UNDERSCORE-D:
Fixes #907

A-UNDERSCORE-D (approved):

PR #923, opened 2021-03-19
Add tests for the journal_lock.py code

Athanasius:
As I didn't want `tests/` becoming a mess I've instituted a policy of having a sub-dir in there named after the file the tests are against.

So, this adds `tests/journal_lock.py/` and files therein, and has moved the config.py test files into `tests/config.py/`.


Athanasius:
I'm still working through the journal_lock.py code adding tests, so this is draft for now.

tests/journal_lock.py/test_journal_lock.py
```diff
@@ -0,0 +1,194 @@
...
+
+    def test_obtain_lock_already_locked(self, mock_journaldir: py_path_local_LocalPath):
+        """Test JournalLock.obtain_lock() with tmpdir."""
+        jlock = JournalLock()
+
+        locked = jlock.obtain_lock()
+        assert locked == JournalLockResult.LOCKED
+        # Now attempt to lock again, but only that.
+        second_attempt = jlock._obtain_lock()
+        assert second_attempt == JournalLockResult.ALREADY_LOCKED
```
Athanasius, line 194: Fails on Linux because flock(2) (which is what fcntl.flock() uses) is a *per process* lock.  As the test is still in the same process, it can lock again because it has the lock already.

Will need to look into multiprocessing to properly test this.

Athanasius:
All tests currently pass for me on both Windows 10 and Linux.  pytest coverage is a bit of a pain, it *will* still report one, or the other, branch of `if platform == 'win32'` code, depending on your platform.

Contributing.md
```diff
@@ -157,10 +160,33 @@ Note that if your PR does not cleanly (or mostly cleanly) pass a linting scan, y
...
 ### Unit testing
 
-Where possible please write unit tests for your PRs, especially in the case of bug fixes, having regression tests help
-ensure that we don't accidentally re-introduce a bug down the line.
+Where possible please write unit tests for your PRs, especially in the case of
+bug fixes, having regression tests help ensure that we don't accidentally
+re-introduce a bug down the line.
 
-We use the python stdlib library `unittest` for unit testing.
+We use the `pytest` for unit testing.
```
A-UNDERSCORE-D, line 167:
```suggestion
We use [`pytest`](https://docs.pytest.org/en/stable/) for unit testing.
```

Contributing.md
```diff
@@ -157,10 +160,33 @@ Note that if your PR does not cleanly (or mostly cleanly) pass a linting scan, y
...
+containing test data.
+
+Invoking just a bare `pytest` command will run all tests.
+
+To run only a sub-set of tests you can use, e.g. `pytest -k journal_lock`. You
+might want to use `pytest -rA -k journal_lock` if you have any debug print()
+statements within the test code itself, so you can see the output even when the
+tests all succeed.
+
+Adding `--trace` to a `pytest` invocation causes it to drop into `pdb` prompt
```
A-UNDERSCORE-D, line 185:
```suggestion
Adding `--trace` to a `pytest` invocation causes it to drop into a [`pdb`](https://docs.python.org/3/library/pdb.html) prompt
```

Contributing.md
```diff
@@ -157,10 +160,33 @@ Note that if your PR does not cleanly (or mostly cleanly) pass a linting scan, y
...
+The `test_` prefix on `test_journal_lock.py` is necessary in order for `pytest`
+to recognise the file as containing tests to be run.
+The sub-directory avoids having a mess of files in `tests`, particularly when
+there might be supporting files, e.g. `tests/config.py/_old_config.py` or files
+containing test data.
+
+Invoking just a bare `pytest` command will run all tests.
+
+To run only a sub-set of tests you can use, e.g. `pytest -k journal_lock`. You
+might want to use `pytest -rA -k journal_lock` if you have any debug print()
```
A-UNDERSCORE-D, line 181:
```suggestion
might want to use `pytest -rA -k journal_lock` if you have any debug `print()`
```

config.py
```diff
@@ -29,7 +29,7 @@
 # appversion **MUST** follow Semantic Versioning rules:
 # <https://semver.org/#semantic-versioning-specification-semver>
 # Major.Minor.Patch(-prerelease)(+buildmetadata)
-appversion = '5.0.0-beta1'  # -rc1+a872b5f'
+appversion = '4.2.0-beta1'  # -rc1+a872b5f'
```
A-UNDERSCORE-D, line 32: I think this might be a holdover we don't want

tests/journal_lock.py/test_journal_lock.py
```diff
@@ -0,0 +1,349 @@
...
+#
+#  - Does release_lock() work?  Easier to test, if it's worked....
+#      1. return True if not locked.
+#      2. return True if locked, but successful unlock.
+#      3. return False otherwise.
+#
+#  - Can any string to pathlib.Path result in an invalid path for other
+#    operations?
+#
+#  - Not sure about testing JournalAlreadyLocked class.
```
A-UNDERSCORE-D, line 23: May want to throw this into the module docstring?

tests/journal_lock.py/test_journal_lock.py
```diff
@@ -0,0 +1,349 @@
...
+
+        with monkeypatch.context() as m:
+            m.setattr(config, "get_str", get_str)
+            yield tmpdir
+
+    @pytest.fixture
+    def mock_journaldir_changing(
+            self,
+            monkeypatch: _pytest_monkeypatch,
+            tmpdir_factory: _pytest_tmpdir.TempdirFactory) -> py_path_local_LocalPath:
```
A-UNDERSCORE-D, line 118:
```suggestion
            tmpdir_factory: _pytest_tmpdir.TempdirFactory
        ) -> py_path_local_LocalPath:
```
Not 100% on this one, I just didnt at a glance see where the function declaration ended and the body begun

config.py
Athanasius: The one time I picked the wrong option resolving a rebase, thanks.

tests/journal_lock.py/test_journal_lock.py
Athanasius: You're right, once we start line-separating function definition args like that we should separate the return type as well.

tests/journal_lock.py/test_journal_lock.py
Athanasius: I found the bit that I'd edited out in an earlier commit, so restored this and made it part of the module docstring.

A-UNDERSCORE-D (approved):

PR #931, opened 2021-03-23
    Fix spurious modules or ships was None

A-UNDERSCORE-D:
Nothing other than /shipyard actually returns the modules and ships data, so checking for it anywhere other than that automatically is just asking for extra log noise and nothing else.

### DIFF
--- a/companion.py
+++ b/companion.py
@@ -634,8 +634,6 @@ class Session(object):
             else:
                 data['lastStarport'].update(shipdata)
 
-        data.check_modules_ships()
-
         return data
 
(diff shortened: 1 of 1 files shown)


companion.py
```diff
@@ -628,6 +634,8 @@ def station(self) -> CAPIData:
             else:
                 data['lastStarport'].update(shipdata)
 
+        data.check_modules_ships()
```
Athanasius, line 637: Doesn't this mean the checks get run twice, and thus double-logging ?

PR #940, opened 2021-03-25
Show git HEAD short hash as part of version

Athanasius:
The idea here is that we can be sure that a version number such as '5.0.0-beta1' is either an actual release with that version or instead is someone running from source where appversion hasn't been changed.

### DIFF
--- a/config.py
+++ b/config.py
@@ -126,6 +126,11 @@ def git_shorthash_from_head() -> str:
 
 
 def appversion() -> str:
+    """
+    Determine app version including git short hash if possible.
+
+    :return: str - The augmented app version.
+    """
     if getattr(sys, 'frozen', False):
         # Running frozen, so we should have a .gitversion file
         with open(GITVERSION_FILE, 'r', encoding='utf-8') as gitv:
@@ -143,6 +148,14 @@ def appversion() -> str:
 
 
 def appversion_nobuild() -> str:
+    """
+    Determine app version without *any* build meta data.
+
+    This will not only strip any added git short hash, but also any trailing
+    '+<string>' in _static_appversion.
+
+    :return: str - App version without any build meta data.
+    """
     return str(semantic_version.Version(appversion()).truncate('prerelease'))
 ###########################################################################
 
--- a/config.py
+++ b/config.py
@@ -13,6 +13,8 @@ import functools
 import logging
 import os
 import pathlib
+import re
+import subprocess
 import sys
 import traceback
 import warnings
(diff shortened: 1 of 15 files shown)


setup.py
```diff
@@ -20,11 +20,26 @@
 
 import semantic_version
 
-from config import appcmdname, applongname, appname, appversion, copyright, update_feed, update_interval
+from config import (
+    appcmdname, applongname, appname, appversion, copyright, git_shorthash_from_head, update_feed, update_interval
+)
```
Athanasius, line 25: This is >< close to needing to be multi-line due to line length, but isort complains if *this* is multi-line.

config.py
```diff
@@ -89,6 +91,34 @@ def known_folder_path(guid: uuid.UUID) -> Optional[str]:
...
 
 
+###########################################################################
+def git_shorthash_from_head() -> str:
+    """
+    Determine short hash for current git HEAD.
+
+    :return: str - None if we couldn't determine the short hash.
+    """
+    shorthash: str = None  # type: ignore
```
Athanasius, line 101: If we try Optional[str] throughout here then the file write() call in setup.py complains because it wants only `str`.  So hide the issue here.

A-UNDERSCORE-D (approved):

Athanasius:
I think this is ready now.

1. Running from source with working `git` and a `.git/` directory will pick up the short hash from HEAD.
2. Running from source with a working `git`, but no `.git/` directory will cause, e.g. `ERROR - 2480:14844:14844 config.git_shorthash_from_head:123: 'fatal: not a git repository (or any of the parent directories): .git' doesn't look like a valid git short hash, forcing to None `.  This would be someone with a git command who is running from a source zip/tar archive.  They'll get `UNKNOWN` showing up as the short hash.
3. Running from source without a working `git` (whether you have a `.git/` or not) -> `ERROR - 15184:2968:2968 config.git_shorthash_from_head:118: Couldn't run git command for short hash: FileNotFoundError(2, 'The system cannot find the file specified', None, 2, None)`.  They'll also get `UNKNOWN`.
4. Running from an installed version will pick up what's in the .gitversion file (which the installer supplied) and show that.

Everything but EDDN gets the full version, including the short hash as 'build meta' information, as in `5.0.0-beta1+a900a59`.  EDDN gets the version *without* the build meta so as not to spam it with lots of different versions during development testing.

The one adjustment I could make is to have config.py:appversion() store what it gets on first run and not bother trying to run the command after that so as to slightly reduce log output.  But for case #3 above someone could potentially install git during a run of EDMC ....

A-UNDERSCORE-D (approved):

PR #943, opened 2021-03-27
fix missing screening in docs/Killswitches.md

norohind:


A-UNDERSCORE-D:
Whoops. That's what I get for not looking.

PR #947, opened 2021-03-28
config.py: Compatibility layer for newer API calls.

Athanasius:
* Also updates coriolis-data to what I pulled for 5.0.0 so `git status` isn't always noting it.  I have *not* generated new .p files from it for this branch.

### DIFF
--- a/config.py
+++ b/config.py
@@ -160,11 +160,23 @@ class Config(object):
             else:
                 return None
 
+        def get_str(self, key):
+            return self.get(key)
+
+        def get_list(self, key):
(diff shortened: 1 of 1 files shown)


A-UNDERSCORE-D (approved):
This appears to need a rebase, or stable does

Athanasius:
> This appears to need a rebase, or stable does

It's the damned coriolis-data sub-module again. But updating that in `stable` is fine, I said somewhere I'd done that but wasn't re-generating the .p files built from the data.

PR #957, opened 2021-03-30
Attempt to get on-foot detection working, so EDDN data sends

Athanasius:
This introduces a new monitor state variable `on_foot`.

Currently this will only be true when you login at a station.  There's only `Disembark` when you get off a shuttle at one, with no new `Location`/`Docked` event to fill in the necessary data to check against CAPI data.

### DIFF
--- a/monitor.py
+++ b/monitor.py
@@ -16,7 +16,7 @@ from typing import Tuple
 if TYPE_CHECKING:
     import tkinter
 
-from companion import ship_file_name
+import util_ships
 from config import config
 from EDMCLogging import get_main_logger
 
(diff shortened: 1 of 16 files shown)


Athanasius:
This should be ready to merge now, although could probably use some non-Alpha testing to ensure I didn't introduce regressions.

Athanasius:
I'll s/on_foot/OnFoot/g in this tomorrow, given the rest of monitor.state is CamelCase.

monitor.py
```diff
@@ -67,6 +67,7 @@ class EDLogs(FileSystemEventHandler):  # type: ignore # See below
     _RE_CANONICALISE = re.compile(r'\$(.+)_name;')
     _RE_CATEGORY = re.compile(r'\$MICRORESOURCE_CATEGORY_(.+);')
     _RE_LOGFILE = re.compile(r'^Journal(Alpha|Beta)?\.[0-9]{12}\.[0-9]{2}\.log$')
+    _RE_SHIP_ONFOOT = re.compile(r'^(FlightSuit|UtilitySuit_Class.)$')
```
A-UNDERSCORE-D, line 70: Is it confirmed whether or not `UtilitySuit_Class` is non-decimal? if its always a decimal this should be \d

Athanasius: It's `"name": "UtilitySuit_Class1"` for the Maverick suit, but I didn't want to assume it'd just be digits.

What we need is a new Journal doc specifying what those can be, or to be far enough into the Alpha to try out other suits.  Do NOT assume Frontier are going to be consistent on such.  Even 'single character' might turn out to not be enough.

So it'll do for now.  Change it as we learn more.

A-UNDERSCORE-D: sounds good

A-UNDERSCORE-D (approved):

PR #969, opened 2021-04-03
Add new data.py file to hold all static data.

Athanasius:
We need the config.py read of .gitversion to be from the right place. The aim here is to avoid static data in modules that contain code.

Why? So that we can avoid some potential import loops when the 'inner' module only wants static data but is already imported by the 'outer' module for code purposes.


outfitting.py
```diff
@@ -1,348 +1,35 @@
...
+from edmc_data import outfitting_planet_rating_map as planet_rating_map
+from edmc_data import outfitting_rating_map as rating_map
+from edmc_data import outfitting_standard_map as standard_map
+from edmc_data import outfitting_utility_map as utility_map
+from edmc_data import outfitting_weapon_map as weapon_map
+from edmc_data import outfitting_weaponclass_map as weaponclass_map
+from edmc_data import outfitting_weaponmount_map as weaponmount_map
+from edmc_data import outfitting_weaponoldvariant_map as weaponoldvariant_map
+from edmc_data import outfitting_weaponrating_map as weaponrating_map
+from edmc_data import ship_name_map
```
A-UNDERSCORE-D, line 24:
```suggestion
from edmc_data import (
    outfitting_armour_map as armour_map,
    outfitting_cabin_map as cabin_map,
    outfitting_corrosion_rating_map as corrosion_rating_map,
    outfitting_countermeasure_map as countermeasure_map,
    outfitting_fighter_rating_map as fighter_rating_map,
    outfitting_internal_map as internal_map,
    outfitting_misc_internal_map as misc_internal_map,
    outfitting_missiletype_map as missiletype_map,
    outfitting_planet_rating_map as planet_rating_map,
    outfitting_rating_map as rating_map,
    outfitting_standard_map as standard_map,
    outfitting_utility_map as utility_map,
    outfitting_weapon_map as weapon_map,
    outfitting_weaponclass_map as weaponclass_map,
    outfitting_weaponmount_map as weaponmount_map,
    outfitting_weaponoldvariant_map as weaponoldvariant_map,
    outfitting_weaponrating_map as weaponrating_map,
    ship_name_map,
)
```

Athanasius: isort no likey that style

PR #976, opened 2021-04-06
plugins/inara.py: Clean up per flake8 and mypy output

Athanasius:


Athanasius:
Remaining:

```
plugins\inara.py:785: error: No overload variant of "__getitem__" of "Sequence" matches argument type "str"
plugins\inara.py:785: note: Possible overload variants:
plugins\inara.py:785: note:     def __getitem__(self, int) -> Mapping[Any, Any]
plugins\inara.py:785: note:     def __getitem__(self, slice) -> Sequence[Mapping[Any, Any]]
plugins\inara.py:785: error: Value of type "Optional[OrderedDict[str, Any]]" is not indexable
```

A-UNDERSCORE-D (approved):

PR #978, opened 2021-04-06
collate.py: Cleanup flake8 and mypy

Athanasius:


Athanasius:
Tested with a fresh live 'Save Raw Data'.

PR #980, opened 2021-04-07
commodity.py: Fix flake8 lints

Athanasius:


commodity.py
```diff
@@ -39,27 +44,30 @@ def export(data, kind=COMMODITY_DEFAULT, filename=None):
...
+                bracketmap[commodity['stockBracket']]
+            ))
+
+            if kind == COMMODITY_DEFAULT:
+                line = sep.join((
+                    line,
+                    str(int(commodity['meanPrice'])),
+                    str(commodity['id']),
+                    data['timestamp'] + '\n')
+                )
```
A-UNDERSCORE-D, line 68:
```suggestion
                line = sep.join(
                    (
                        line,
                        str(int(commodity['meanPrice'])
                    ),
                    str(commodity['id']),
                    data['timestamp'] + '\n')
                )
```
Or similar, to better show that the tuple doesnt contain everything here. Though Im not sure if this is a bug or not.

Athanasius: It's actually:

```python
                line = sep.join(
                    (
                        line,
                        str(int(commodity['meanPrice'])),
                        str(commodity['id']),
                        data['timestamp'] + '\n'
                    )
                )
```
I'll move that 2nd-last `)` down a line to make it more obvious

Athanasius: Actually, I'll just split it as above comment.

A-UNDERSCORE-D (approved):

PR #983, opened 2021-04-07
edshipyard.py: linting

Athanasius:


edshipyard.py
```diff
@@ -124,40 +140,43 @@ def class_rating(module: __Module):
...
     assert ship_name_map[data['ship']['name'].lower()] in ships, ship_name_map[data['ship']['name'].lower()]
 
     try:
         mass += ships[ship_name_map[data['ship']['name'].lower()]]['hullMass']
-        string += 'Mass  : {:.2f} T empty\n        {:.2f} T full\n'.format(mass, mass + fuel + cargo)
+        string += f'Mass  : {mass:.2f} T empty\n        {mass + fuel + cargo:.2f} T full\n'
 
-        multiplier = pow(min(fuel, fsd['maxfuel']) / fsd['fuelmul'], 1.0 / fsd['fuelpower']) * fsd['optmass']
+        multiplier = pow(min(fuel, fsd['maxfuel']) / fsd['fuelmul'], 1.0  # type: ignore
+                         / fsd['fuelpower']) * fsd['optmass']  # type: ignore
```
Athanasius, line 173: The types get all sorts of confused here, I think due to `__Module`.  Things that are clearly `float` in a debugger are `Union[str, List[str]]`.

A-UNDERSCORE-D: Looks to be because fsd comes from __Module as you said. Not sure how you want to go about fixing that?

Athanasius: For now ?  The ignores :D

Athanasius: It checks out in the debugger... and if we wanted to be paranoid about the input data then that would get done long before this point.

edshipyard.py
```diff
@@ -174,14 +193,13 @@ def class_rating(module: __Module):
     regexp = re.compile(re.escape(ship) + r'\.\d{4}-\d\d-\d\dT\d\d\.\d\d\.\d\d\.txt')
     oldfiles = sorted([x for x in os.listdir(config.get_str('outdir')) if regexp.match(x)])
     if oldfiles:
-        with open(join(config.get_str('outdir'), oldfiles[-1]), 'rU') as h:
+        with open(pathlib.Path(config.get_str('outdir')) / oldfiles[-1], 'rU') as h:
```
A-UNDERSCORE-D, line 196:
```suggestion
        with (pathlib.Path(config.get_str('outdir')) / oldfiles[-1]).open('rU') as h:
```

edshipyard.py
```diff
@@ -21,29 +24,44 @@
...
-def export(data, filename=None):
-    def class_rating(module: __Module):
+        :param module: Module data dict.
+        :return: Rating of the module.
+        """
         mod_class = module['class']
         mod_rating = module['rating']
         mod_mount = module.get('mount')
-        mod_guidance = module.get('guidance')
+        mod_guidance: str = module.get('guidance')  # type: ignore
```
A-UNDERSCORE-D, line 47:
```suggestion
        mod_guidance: str = str(module.get('guidance'))
```
Just to force it and drop the need for the ignore

A-UNDERSCORE-D (approved):

PR #988, opened 2021-04-08
Any further fixups to on-foot support now we have our own ships

Athanasius:
1. monitor: Only forget station on Disembark if not from own ship


Athanasius:
As journal isn't flushing to disk with every write this sort of thing is a PITA to test.

I'm confident this single commit is correct, certainly more correct than without it.

A-UNDERSCORE-D (approved):

Issue #996, opened 2021-04-11
Make sure *all* EDLogs 'self' members are initialised


Athanasius:
```python
2021-04-10 23:49:05.193 - DEBUG - 154716:153680:153680 monitor.EDLogs.worker:321: Now at end of latest file.
Exception in thread Journal worker:
Traceback (most recent call last):
  File "threading.pyc", line 954, in _bootstrap_inner
  File "threading.pyc", line 892, in run
  File "monitor.pyc", line 334, in worker
AttributeError: 'EDLogs' object has no attribute 'systempopulation'
```

EDLogs.systempopulation is only set in one place and never properly initialised.

Athanasius:
It turns out this was literally only self.systempopulation

Athanasius:
Closed by #997

PR #997, opened 2021-04-12
monitor.py: Initialise EDLogs self.systempopulation

Athanasius:
This is the *one* field that wasn't always initialised.


Athanasius:
Will close #996

A-UNDERSCORE-D (approved):

PR #1004, opened 2021-04-13
Add support for Odyssey Suit/Loadout data

Athanasius:
There's no way to get any of this from the Journal currently, so it comes from CAPI.

We should be able to use some Journal events to then update the current suit in UI though.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 .gitversion
 .DS_Store
 build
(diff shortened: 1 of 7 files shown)


Athanasius:
I think this is as much as we can do with the current state of observed Journal events.

A-UNDERSCORE-D (approved):

PR #1007, opened 2021-04-16
CAPI loadouts is a sparse list, added index function

A-UNDERSCORE-D:


Athanasius:
For me they're still a dict.

There's some FDev shenanigans afoot that we need to wait to settle.

A-UNDERSCORE-D:
as per discussion, added a util function to index sparse lists

PR #1011, opened 2021-04-25
EDMarketConnector: Added minimize to tray on close functionality

### LINKED ISSUE #126: Minimize to system tray option (Windows)
Would love to see the window management options extend to allow minimizing the primary EDMC window to the System Tray (systray) instead of the Task Bar, please. This would also, naturally, remove it from the alt-tab listing since the program can already be interacted with via your hotkey feature.

SayakMukhopadhyay:
Closes #126 

This PR adds a preference option and implementation using `infi.systray` package. There are a few hacks here and there and translation for the new option is not done.

Starting with the PR to get feedback.

### DIFF
--- a/EDMarketConnector.py
+++ b/EDMarketConnector.py
@@ -264,6 +264,7 @@ if __name__ == '__main__':  # noqa: C901
 if TYPE_CHECKING:
     from logging import trace, TRACE  # type: ignore # noqa: F401
     import update
+    from infi.systray import SysTrayIcon
 # isort: on
 
     def _(x: str) -> str:
@@ -334,7 +335,7 @@ class AppWindow(object):
         if platform == 'win32':
             from infi.systray import SysTrayIcon
 
-            def open_window(systray) -> None:
+            def open_window(systray: 'SysTrayIcon') -> None:
                 self.w.deiconify()
 
             menu_options = (("Open", None, open_window),)
@@ -1383,7 +1384,7 @@ class AppWindow(object):
             logger.debug('"other" exception', exc_info=e)
             self.status['text'] = str(e)
 
-    def exit_tray(self, systray) -> None:
+    def exit_tray(self, systray: 'SysTrayIcon') -> None:
         """Tray icon is shutting down."""
         exit_thread = threading.Thread(target=self.onexit)
         exit_thread.setDaemon(True)
--- a/EDMarketConnector.py
+++ b/EDMarketConnector.py
@@ -9,6 +9,7 @@ import locale
 import pathlib
 import re
 import sys
+import threading
 import webbrowser
 from builtins import object, str
 from os import chdir, environ
@@ -331,28 +332,15 @@ class AppWindow(object):
         self.prefsdialog = None
 
         if platform == 'win32':
-            import threading
             from infi.systray import SysTrayIcon
 
-            self.only_tray_close = 0  # This is kind of a hack.
-            # When the tray icon is double click to reopen the EDMC window, I want the tray icon to disappear. To do
-            # that I have to call the `shutdown()` method of the systray object. Calling the shutdown method triggers
-            # the method associated with `on_quit` twice, once for WM_DESTROY and again for WM_CLOSE. This is not the
-            # case when the application is exited by clicking on `Quit` in the tray menu. So, to handle this, we are
-            # creating this class variable and setting it to 2 when `Open` is called from the tray menu (either by
-            # clicking open or by double clicking the icon) and decrementing it by 1 when the tray shutdown is triggered
-
             def open_window(systray) -> None:
-                self.only_tray_close = 2
-                # Shutdown needs to happen in a separate thread to prevent joining with itself
-                shutdown_thread = threading.Thread(target=systray.shutdown)
-                shutdown_thread.setDaemon(True)
-                shutdown_thread.start()
                 self.w.deiconify()
 
             menu_options = (("Open", None, open_window),)
             # Method associated with on_quit is called whenever the systray is closing
             self.systray = SysTrayIcon("EDMarketConnector.ico", applongname, menu_options, on_quit=self.exit_tray)
+            self.systray.start()
 
         plug.load_plugins(master)
 
@@ -1395,35 +1383,20 @@ class AppWindow(object):
             logger.debug('"other" exception', exc_info=e)
             self.status['text'] = str(e)
 
-    def onexit(self, event=None) -> None:
-        """Application shutdown procedure."""
-        if platform == 'win32':
-            value = bool(config.get_int('close_system_tray'))
-            # When exit is called, either exit application or minimize to system tray
-            if value:
-                self.w.withdraw()
-                self.systray.start()
-
-            else:
-                self.exit()
-
-        else:
-            self.exit()
-
     def exit_tray(self, systray) -> None:
         """Tray icon is shutting down."""
-        import threading
-        # Hack to see if the tray shutdown has been called by calling quit or calling shutdown
-        if self.only_tray_close > 0:
-            self.only_tray_close -= 1
+        exit_thread = threading.Thread(target=self.onexit)
+        exit_thread.setDaemon(True)
+        exit_thread.start()
 
-        else:
-            exit_thread = threading.Thread(target=self.exit)
-            exit_thread.setDaemon(True)
-            exit_thread.start()
+    def onexit(self, event=None) -> None:
+        """Application shutdown procedure."""
+        value = config.get_bool('minimize_system_tray')
+        if platform == 'win32' and value is not None and value:
+            shutdown_thread = threading.Thread(target=self.systray.shutdown)
+            shutdown_thread.setDaemon(True)
+            shutdown_thread.start()
 
-    def exit(self) -> None:
-        """Actual application shutdown."""
         config.set_shutdown()  # Signal we're in shutdown now.
 
         # http://core.tcl.tk/tk/tktview/c84f660833546b1b84e7
@@ -1490,12 +1463,17 @@ class AppWindow(object):
         self.drag_offset = (None, None)
 
     def oniconify(self, event=None) -> None:
-        """Handle iconification of the application."""
-        self.w.overrideredirect(0)  # Can't iconize while overrideredirect
-        self.w.iconify()
-        self.w.update_idletasks()  # Size and windows styles get recalculated here
-        self.w.wait_visibility()  # Need main window to be re-created before returning
-        theme.active = None  # So theme will be re-applied on map
+        """Handle minimization of the application."""
+        value = config.get_bool('minimize_system_tray')
+        if platform == 'win32' and value is not None and value:
+            self.w.withdraw()
+
+        else:
+            self.w.overrideredirect(0)  # Can't iconize while overrideredirect
+            self.w.iconify()
+            self.w.update_idletasks()  # Size and windows styles get recalculated here
+            self.w.wait_visibility()  # Need main window to be re-created before returning
+            theme.active = None  # So theme will be re-applied on map
 
     # TODO: Confirm this is unused and remove.
     def onmap(self, event=None) -> None:
--- a/EDMarketConnector.py
+++ b/EDMarketConnector.py
@@ -9,6 +9,7 @@ import locale
 import pathlib
 import re
 import sys
+import threading
 import webbrowser
 from builtins import object, str
 from os import chdir, environ
@@ -16,6 +17,7 @@ from os.path import dirname, join
 from sys import platform
 from time import localtime, strftime, time
 from typing import TYPE_CHECKING, Optional, Tuple
+from infi.systray import SysTrayIcon
 
 # Have this as early as possible for people running EDMarketConnector.exe
 # from cmd.exe or a bat file or similar.  Else they might not be in the correct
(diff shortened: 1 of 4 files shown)


EDMarketConnector.py
```diff
@@ -1373,6 +1387,23 @@ def save_raw(self) -> None:  # noqa: CCR001 # Not easily broken up.
 
     def onexit(self, event=None) -> None:
         """Application shutdown procedure."""
+        value = bool(config.get_int('close_system_tray'))
```
Athanasius, line 1390: `config.get_bool()` exists, use it.

Athanasius:
In general you seem to be doing *CLOSE* to systray, i.e. top-right red 'X' or File > Exit will, when the option is active, cause instead the window to be hidden and the systray active.

That's not the same as *minimise* to systray, which is what people have been asking for.

Athanasius:
> That's not the same as _minimise_ to systray, which is what people have been asking for.

Changing code in `def oniconify()` instead should achieve this.

Technically I think that "Close to systray" should actually *exit* the main program, and re-run it for 'restore', but that's a whole can of worms with needing the systray to be a separate process.

Athanasius:
Also, we use flake8 and mypy, they're in requirements-dev.txt for that reason.  The check failed on the PR due to necessary changes.  So please check the flake8 output and make corrections.

SayakMukhopadhyay:
Yeah, I am planning to add additional commits to add some comments at the least and maybe some tests if possible. I will lint then.

SayakMukhopadhyay:
> That's not the same as *minimise* to systray, which is what people have been asking for.

The reason I went for close to system tray as I find that too be a more popular option. Discord does it, teams, Outlook, Steam, etc. I realize that the issue was specifically for minimize but I implemented what I thought to be more popular. As far as the user is concerned, their experience should not change.

But I am open to changing this or even adding a seperate option to handle minizing. I had even thought of an additional button on the toolbar, next to the minimize icon which would always minimize to system tray. That would take away the need to have a app preference.

requirements.txt
```diff
@@ -1,6 +1,7 @@
 certifi==2020.12.5
 requests==2.25.1
 watchdog==2.0.3
+infi.systray==0.1.12
```
Athanasius, line 4: This will need a `; sys_platform == 'win32'` appending to the line, as this is literally only for on windows.

Athanasius:
All of this is Windows-only, so will need gating in `if sys.platform == 'win32':` conditionals.

EDMarketConnector.py
```diff
@@ -1373,6 +1387,23 @@ def save_raw(self) -> None:  # noqa: CCR001 # Not easily broken up.
...
+
+        if value:
+            self.w.withdraw()
+            self.systray.start()
+        else:
```
A-UNDERSCORE-D, line 1395:
```suggestion

        else:
```

EDMarketConnector.py
```diff
@@ -1373,6 +1387,23 @@ def save_raw(self) -> None:  # noqa: CCR001 # Not easily broken up.
...
+            self.exit()
+
+    def exit_tray(self, systray):
+        if self.only_tray_close > 0:
+            self.only_tray_close -= 1
+        else:
```
A-UNDERSCORE-D, line 1401:
```suggestion

        else:
```

prefs.py
```diff
@@ -636,6 +636,7 @@ def __setup_appearance_tab(self, notebook: Notebook) -> None:
         # Appearance theme and language setting
         self.lang = tk.StringVar(value=self.languages.get(config.get_str('language'), _('Default')))
         self.always_ontop = tk.BooleanVar(value=bool(config.get_int('always_ontop')))
+        self.close_system_tray = tk.BooleanVar(value=bool(config.get_int('close_system_tray')))
```
A-UNDERSCORE-D, line 639:
```suggestion
        self.close_system_tray = tk.BooleanVar(value=config.get_bool('close_system_tray'))
```

EDMarketConnector.py
```diff
@@ -330,6 +332,18 @@ def __init__(self, master: tk.Tk):  # noqa: C901, CCR001 # TODO - can possibly f
 
         self.prefsdialog = None
 
+        self.only_tray_close = 0
+
+        def open_window(systray):
```
A-UNDERSCORE-D, line 337: type annotations please

EDMarketConnector.py
```diff
@@ -1373,6 +1387,23 @@ def save_raw(self) -> None:  # noqa: CCR001 # Not easily broken up.
...
+
+    def exit_tray(self, systray):
+        if self.only_tray_close > 0:
+            self.only_tray_close -= 1
+        else:
+            exit_thread = threading.Thread(target=self.exit)
+            exit_thread.setDaemon(True)
+            exit_thread.start()
+
+    def exit(self):
```
A-UNDERSCORE-D, line 1406:
```suggestion
    def exit(self) -> None:
```

EDMarketConnector.py
```diff
@@ -330,6 +332,18 @@ def __init__(self, master: tk.Tk):  # noqa: C901, CCR001 # TODO - can possibly f
 
         self.prefsdialog = None
 
+        self.only_tray_close = 0
```
A-UNDERSCORE-D, line 335: Please clarify how this is used in a comment (I saw you're planning some commits for that), as its unclear on first look over.

A-UNDERSCORE-D (changes requested):

EDMarketConnector.py
SayakMukhopadhyay: Done

requirements.txt
SayakMukhopadhyay: Done

EDMarketConnector.py
SayakMukhopadhyay: Done

EDMarketConnector.py
SayakMukhopadhyay: Done

prefs.py
SayakMukhopadhyay: Done

EDMarketConnector.py
SayakMukhopadhyay: Done

EDMarketConnector.py
SayakMukhopadhyay: Done

EDMarketConnector.py
SayakMukhopadhyay: Done. Probably a bit too much but still.

SayakMukhopadhyay:
I have added commits with the requested changes. Please review and let me know if there are any further changes required.
Also, are there any comments regarding https://github.com/EDCD/EDMarketConnector/pull/1011#issuecomment-826396993

EDMarketConnector.py
```diff
@@ -330,6 +330,30 @@ def __init__(self, master: tk.Tk):  # noqa: C901, CCR001 # TODO - can possibly f
 
         self.prefsdialog = None
 
+        if platform == 'win32':
+            import threading
+            from infi.systray import SysTrayIcon
+
+            self.only_tray_close = 0  # This is kind of a hack.
```
Athanasius, line 337: I'm still not seeing why this is an `int` instead of a `bool`.  The config.py code will be doing the conversion to/from `int` for Windows Registry purposes.

Athanasius: And now I read the comment about the hack... which I'm going to have to have a think about.

EDMarketConnector.py
```diff
@@ -330,6 +330,30 @@ def __init__(self, master: tk.Tk):  # noqa: C901, CCR001 # TODO - can possibly f
...
+            # When the tray icon is double click to reopen the EDMC window, I want the tray icon to disappear. To do
+            # that I have to call the `shutdown()` method of the systray object. Calling the shutdown method triggers
+            # the method associated with `on_quit` twice, once for WM_DESTROY and again for WM_CLOSE. This is not the
+            # case when the application is exited by clicking on `Quit` in the tray menu. So, to handle this, we are
+            # creating this class variable and setting it to 2 when `Open` is called from the tray menu (either by
+            # clicking open or by double clicking the icon) and decrementing it by 1 when the tray shutdown is triggered
+
+            def open_window(systray) -> None:
+                self.only_tray_close = 2
```
Athanasius, line 346: At the very least, if we agree to keep this hacky behaviour then the possible values of `self.only_tray_close` need to be defined in an Enum and used as such.  No magic numbers in actual code.

Athanasius:
> > That's not the same as _minimise_ to systray, which is what people have been asking for.
> 
> The reason I went for close to system tray as I find that too be a more popular option. Discord does it, teams, Outlook, Steam, etc. I realize that the issue was specifically for minimize but I implemented what I thought to be more popular. As far as the user is concerned, their experience should not change.
> 
> But I am open to changing this or even adding a seperate option to handle minizing. I had even thought of an additional button on the toolbar, next to the minimize icon which would always minimize to system tray. That would take away the need to have a app preference.

The actual reason I saw for wanting the feature at all was from people who don't want EDMC to be in the alt-tab options when minimised.  So, at the very least, this PR needs to also catch `oniconify` and have that go to systray as well, with the `withdraw` taking care of hiding from alt-tab.

EDMarketConnector.py
SayakMukhopadhyay: Did you mark the wrong piece of code? Cause the marked code has nothing to do with config.

Athanasius: Ah yes, but my point about using an Enum still stands.

SayakMukhopadhyay:
Changed the logic. Now, the application is minimized to system tray on clicking the minimize icon. This also makes the hack unnecessary.

EDMarketConnector.py
SayakMukhopadhyay: Hacky behaviour no longer there

Athanasius:
With the latest code:

1. Run under PyCharm debugger
2. Wait for EDMarketConnector to fully start (detects I'm active in game and docked, so does a CAPI pull).
3. Hit red top-right 'X' to exit EDMC.
4. All the EDMc shutdown runs but...
5. The systray doesn't exit and the process remains.

You now have a stuck process where systray 'Open' does nothing:
```python
Exception ignored on calling ctypes callback function: <bound method SysTrayIcon.WndProc of <infi.systray.traybar.SysTrayIcon object at 0x0675B2B0>>
  File "C:\Users\Athan\Documents\Devel\EDMarketConnector\venv\lib\site-packages\infi\systray\traybar.py", line 79, in WndProc
    self._message_dict[msg](hwnd, msg, wparam.value, lparam.value)
  File "C:\Users\Athan\Documents\Devel\EDMarketConnector\venv\lib\site-packages\infi\systray\traybar.py", line 278, in _command
    self._execute_menu_option(id)
  File "C:\Users\Athan\Documents\Devel\EDMarketConnector\venv\lib\site-packages\infi\systray\traybar.py", line 285, in _execute_menu_option
    menu_action(self)
  File "C:/Users/Athan/Documents/Devel/EDMarketConnector/EDMarketConnector.py", line 339, in open_window
    self.w.deiconify()
  File "C:\Program Files (x86)\Python39-32\lib\tkinter\__init__.py", line 2013, in wm_deiconify
    return self.tk.call('wm', 'deiconify', self._w)
RuntimeError: main thread is not in main loop```
and you need to systray 'Quit' to clean things up.

**Review by A-UNDERSCORE-D — APPROVED** [review]
I have _not_ tested this as I am not on windows, but the code looks fine.

**Comment by SayakMukhopadhyay** [comment]
> 
> 
> With the latest code:
> 
>     1. Run under PyCharm debugger
> 
>     2. Wait for EDMarketConnector to fully start (detects I'm active in game and docked, so does a CAPI pull).
> 
>     3. Hit red top-right 'X' to exit EDMC.
> 
>     4. All the EDMc shutdown runs but...
> 
>     5. The systray doesn't exit and the process remains.
> 
> 
> You now have a stuck process where systray 'Open' does nothing:
> 
> ```python
> Exception ignored on calling ctypes callback function: <bound method SysTrayIcon.WndProc of <infi.systray.traybar.SysTrayIcon object at 0x0675B2B0>>
> Traceback (most recent call last):
>   File "C:\Users\Athan\Documents\Devel\EDMarketConnector\venv\lib\site-packages\infi\systray\traybar.py", line 79, in WndProc
>     self._message_dict[msg](hwnd, msg, wparam.value, lparam.value)
>   File "C:\Users\Athan\Documents\Devel\EDMarketConnector\venv\lib\site-packages\infi\systray\traybar.py", line 278, in _command
>     self._execute_menu_option(id)
>   File "C:\Users\Athan\Documents\Devel\EDMarketConnector\venv\lib\site-packages\infi\systray\traybar.py", line 285, in _execute_menu_option
>     menu_action(self)
>   File "C:/Users/Athan/Documents/Devel/EDMarketConnector/EDMarketConnector.py", line 339, in open_window
>     self.w.deiconify()
>   File "C:\Program Files (x86)\Python39-32\lib\tkinter\__init__.py", line 2013, in wm_deiconify
>     return self.tk.call('wm', 'deiconify', self._w)
> RuntimeError: main thread is not in main loop```
> and you need to systray 'Quit' to clean things up.
> ```

I am unable to reproduce this. Are there any other steps ?

**Comment by SayakMukhopadhyay** [comment]
Yep, I was able to reproduce the issue.

**Comment by SayakMukhopadhyay** [comment]
> 
> 
> > I am unable to reproduce this. Are there any other steps ?
> 
> Do **not** have the 'Minimise to system tray' option active.

This is fixed.

**Comment by Athanasius** [comment]
Yup, all looks good now.  I'll give the diff one last look over this evening and approve/merge if I don't spot anything else.

File: EDMarketConnector.py [hunk 11]
```diff
@@ -263,6 +264,7 @@ def already_running_popup():
 if TYPE_CHECKING:
     from logging import trace, TRACE  # type: ignore # noqa: F401
     import update
+    from infi.systray import SysTrayIcon
```
Athanasius, line 267: This needs a platform win32 guard.

Athanasius: Never mind, mypy is a weirdo and doesn't care.

PR #1013, opened 2021-04-29
Odyssey Alpha4 extra suits-related support

Athanasius:
Alpha4 has augmented Suit related events with extra data, so use it.

### DIFF
--- a/monitor.py
+++ b/monitor.py
@@ -973,10 +973,22 @@ class EDLogs(FileSystemEventHandler):  # type: ignore # See below
                 #     • Loadoutname
                 pass
 
-            # `BuySuit` has no useful info as of 4.0.0.13
             elif event_type == 'BuySuit':
-                # alpha4 - should have 'SuitID'
+                # alpha4 :
+                # { "timestamp":"2021-04-29T09:03:37Z", "event":"BuySuit", "Name":"UtilitySuit_Class1",
+                # "Name_Localised":"Maverick Suit", "Price":150000, "SuitID":1698364934364699 }
+                self.state['Suits'].update(
+                    {entry['SuitID']: {
+                        'name': entry['Name'],
+                        'locName': entry.get('Name_Localised', entry['Name']),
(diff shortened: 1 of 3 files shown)


monitor.py
```diff
@@ -971,13 +997,38 @@ def parse_entry(self, line: bytes) -> MutableMapping[str, Any]:  # noqa: C901, C
...
+                    'id': None,  # Is this an FDev ID for suit type ?
+                    'suitId':    entry['SuitID'],
+                    'slots':     [],
+                }
+                if self.state['Suits'] is None:
+                    self.state['Suits'] = {entry['SuitID']: new_suit}
+
+                else:
+                    self.state['Suits'].update(
+                        {entry['SuitID']: new_suit}
```
A-UNDERSCORE-D, line 1027: Does this really end up doing anything different here?
If suits is None: Add the suit under the ID
otherwise: Add the suit under the ID

All that changes is that update is used where on both it could be if suits is None make it an empty dict, then regardless set the key for this suits id

monitor.py
```diff
@@ -971,13 +997,38 @@ def parse_entry(self, line: bytes) -> MutableMapping[str, Any]:  # noqa: C901, C
...
+                    )
+
                 # update credits
-                pass
+                self.state['Credits'] -= entry.get('Price', 0)
```
A-UNDERSCORE-D, line 1031: we may want to log here if the price didnt exist? Not sure.

monitor.py
Athanasius: Yeah, I'll change this.  It's a product of the direction I came at it, and which bit got written first.

monitor.py
Athanasius: Yes, as this shouldn't actually ever happen, so no log noise, but we would want to know if someone complained Credits wasn't tracking correctly, I'll change all these to:

```python
                if price := entry.get('Price') is None:
                    logger.error(f"SellSuit didn't contain Price: {entry}")

                else:
                    self.state['Credits'] += price
```

PR #1017, opened 2021-05-01
Misc fixes for Odyssey

Athanasius:
1. Docking in a taxi is before CAPI location is updated, but it is for Disembark.


A-UNDERSCORE-D (approved):
looks alright to me

PR #1027, opened 2021-05-05
Be paranoid about presence of 'Cost' on some Journal events

### LINKED ISSUE #1026: Review Module store/retrieve 'cost'
The Journal document says that events like `ModuleRetrieve` have a 'Cost' on them, but it seems they don't always.  Check also `ModuleStore`.  Let's put a guard on the use of 'Cost' to update `monitor.store['Credits']`.

Athanasius:
This addresses `ModuleRetrieve` and `ModuleStore`, but I'm reviewing other events said to have 'Cost' as well.

Closes #1026

### DIFF
--- a/monitor.py
+++ b/monitor.py
@@ -607,7 +607,7 @@ class EDLogs(FileSystemEventHandler):  # type: ignore # See below
                 self.state['Credits'] -= entry['BuyPrice']
 
             elif event_type == 'ModuleRetrieve':
-                self.state['Credits'] -= entry['Cost']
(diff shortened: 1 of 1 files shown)


Athanasius:
I checked others where we attempt to reference `entry['Cost']`:

1. `BuyTradeData` does have it.
2. I can't trigger `BuyExplorationData`.  Odyssey Alphg doesn't seem to have any way to do this, and on live I either have the data, the icon is there but says 'No Exploration data to buy', or the icon isn't present at all.

All the others should be OK.

A-UNDERSCORE-D (approved):

Issue #1038, opened 2021-05-08
Don't assume MissionCompleted always has Reward


Athanasius:
```python
2021-05-08 16:01:43.954 - DEBUG - 332384:334436:334436 monitor.EDLogs.parse_entry:1510: Invalid journal entry:
b'{ "timestamp":"2021-05-08T13:01:43Z", "event":"MissionCompleted", "Faction":"Ngalinn Jet Natural Incorporated", "Name":"Mission_AltruismCredits_CivilUnrest_name", "MissionID":764123061, "Donation":"450000", "Donated":450000, "FactionEffects":[ { "Faction":"Ngalinn Jet Natural Incorporated", "Effects":[ { "Effect":"$MISSIONUTIL_Interaction_Summary_SP_up;", "Effect_Localised":"The security status of $#MinorFaction; has improved in the $#System; system.", "Trend":"UpGood" } ], "Influence":[ { "SystemAddress":3107509342922, "Trend":"UpGood", "Influence":"++" } ], "ReputationTrend":"UpGood", "Reputation":"++" } ] }\r\n'
Traceback (most recent call last):
  File "monitor.pyc", line 1351, in parse_entry
KeyError: 'Reward'
```

Athanasius:
I'll just go apply the `entry.get('thing', 0)` paranoia to *all* of the Credits updating code, rather than continue with it piece-meal as we find the exceptional instances of events.

PR #1040, opened 2021-05-08
Credits: Be paranoid about all related Journal events

Athanasius:
MissionCompleted doesn't contain 'Reward' if you pick an option without a credits amount (rather than being there but with zero value).

Let's not wait to find out about other exceptions, just utilise entry.get(<field>, 0) on them all.


Athanasius:
Closes #1038

A-UNDERSCORE-D (approved):

Issue #1054, opened 2021-05-12
EDDN: Send new 'odyssey' flag on all messages


Athanasius:
As per https://github.com/EDCD/EDDN/pull/115 we need to start sending an 'odyssey' flag when we detect we're running under Odyssey.

PR #1055, opened 2021-05-12
plugins/eddn: Send bool 'odyssey' in all event messages.

Athanasius:
This is so listeners can make decision on if/how/when to store data that might be from base/horizons or Odyssey.

### DIFF
--- a/monitor.py
+++ b/monitor.py
@@ -119,6 +119,7 @@ class EDLogs(FileSystemEventHandler):  # type: ignore # See below
             'Credits':            None,
             'FID':                None,  # Frontier Cmdr ID
             'Horizons':           None,  # Does this user have Horizons?
+            'Odyssey':            False,  # Have we detected we're running under Odyssey?
             'Loan':               None,
             'Raw':                defaultdict(int),
             'Manufactured':       defaultdict(int),
@@ -520,6 +521,7 @@ class EDLogs(FileSystemEventHandler):  # type: ignore # See below
                     'Credits':    entry['Credits'],
                     'FID':        entry.get('FID'),   # From 3.3
                     'Horizons':   entry['Horizons'],  # From 3.0
+                    'Odyssey':    entry.get('Odyssey', False),  # From 4.0 Odyssey
                     'Loan':       entry['Loan'],
                     'Engineers':  {},
                     'Rank':       {},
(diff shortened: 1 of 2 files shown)


Athanasius:
I think I got all that updated correctly, but until the live EDDN gateway schemas are updated I can't fully test it to be sure.  Hence draft for now.

Athanasius:
Will close #1054

Athanasius:
Testing with this branch today (EDDN Gateway has updated schemas now), it looks like outfitting, probably from CAPI, isn't yet sending the flag.

Athanasius:
> Testing with this branch today (EDDN Gateway has updated schemas now), it looks like outfitting, probably from CAPI, isn't yet sending the flag.

As per Discord, despite seeing https://ross.eddb.io/eddn/log/747006774 without the flag present, I can't see in the code why it wouldn't work, and during testing https://ross.eddb.io/eddn/log/747037393 has the flag (from CAPI data), and tracing through into the requests .post() call showed it in the text generated from the json for journal outfitting.

Either I'm not seeing how this could happen... or someone else running 5.0.0 (we don't send build hash over EDDN) on Windows happened to also dock there.

Other possibility is I'm in the habit of doing a quick relog at the end of a session so that Inara gets sent the new credits total.  Maybe that quick relog can trigger this?  But all the same monitor and plugins/eddn code would be run....

A-UNDERSCORE-D (approved):

Athanasius:
Right, the odyssey flag we're adding.  Out of paranoia I asked PyCharm what all the usages of plugins/eddn.py EDDN.send() were.  All but one, export_journal_entry() are explicitly setting it at the call to EDDN.send(), and the one use of export_journal_entry() is setting it early on in its body, and I couldn't immediately spot any way it would get unset/over-written, but let me double check that ....

That caller being the plugin journal_entry() itself.

And, yup, no other assignment/overwrite to entry up to the call to export_journal_entry().  Closest thing is the entry['Factions'] overwrite, but it's only for that key.

hmmm, actually, what does filer_localised() do ...

What you'd expect, only operating on keys ending with _Localised so it's not going to interfere.

Ah, and definitely at least one other Commander docked at the same station that wasn't me, because it was 3 hours ago.   So I'm going to conclude the "Ack, 5.0.0 message without odyssey flag!" was a false positive, merge the branch and move on.

At some point I think we need:

1. a CL arg to change the destination for EDDN messages.
2. A simple listener to just log all the messages.

so I don't have to rely on messages making it through the EDDN relay (duplication detection) and know it's MY messages.

PR #1066, opened 2021-05-17
Add the ability to copy entries from the status page

### LINKED ISSUE #1060: make values from File -> Status copyable
**Describe the enhancement**

In `Status` window (Main EDMC window -> `File` -> `Status`) I want to be able to copy values, as example: system, where a ship stored, balance.

**Expected behavior**

Can pick out and copy to clipboard by ctrl+c any values from any tab from `Status` window.

**Considered workarounds**

Retype needed value manually. It is slow and I don't like it.

**Additional context**

None

A-UNDERSCORE-D:
Closes #1060 

Currently theres no way to copy entire lines or pages as that would require more storage than currently exists.

### DIFF
--- a/stats.py
+++ b/stats.py
@@ -426,7 +426,9 @@ class StatsResults(tk.Toplevel):
         """Add a spacer to the page."""
         self.addpagerow(parent, [''])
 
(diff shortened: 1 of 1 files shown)


Athanasius:
A quick test and a glance over the code and this looks good.

But I'm paranoid, this can wait for post-Odyssey-launch.

PR #1084, opened 2021-05-20
EngineerProgress: Extra validation paranoia

Athanasius:
This is an example of just how icky it gets validating an event by hand.

Technically we need a function, many would probably be lengthier andmore complex than this one, for **every   single   journal   event  type**.

### DIFF
--- a/monitor.py
+++ b/monitor.py
@@ -1668,14 +1668,14 @@ class EDLogs(FileSystemEventHandler):  # type: ignore # See below
             for e in entry['Engineers']:
                 for f in ('Engineer', 'EngineerID', 'Rank', 'Progress', 'RankProgress'):
                     if f not in e:
-                        logger.warning(f"Engineer entry without '{f=}' key: {e=}")
+                        logger.warning(f"Engineer entry without '{f}' key: {e=}")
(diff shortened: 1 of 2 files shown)


A-UNDERSCORE-D (approved):
Looks okay, though chunks of that nested if can probably be converted to ands

Athanasius:
> Looks okay, though chunks of that nested if can probably be converted to ands

I like such nested conditionals for extra clarity, indent makes it very obvious that the second part depends on the first.

In the long run I suspect I'm going to have to knuckle down and write all those JSON schemas to validate against.  This was bad enough with the Rank/Progress caveats.  It'll long-term be cleaner to just have a "journal-schemas/2 directory with a file per event type to throw through the available JSON validator module.

A-UNDERSCORE-D:
ah okay, I was poking at doing something for that but if a validator exists it'll be far better than anything that I can come up with _anyway_

Athanasius:
> ah okay, I was poking at doing something for that but if a validator exists it'll be far better than anything that I can come up with _anyway_

See WIP https://github.com/Athanasius/ed-journal-verify

PR #1086, opened 2021-05-20
Fix BackPack event support to actually work.

Athanasius:
* Surprise!  The new event is `BackPack`, not `Backpack`, although the filename *is* `Backpack.json`.
* Store the loaded JSON dict in `monitor.state['BackpackJSON']`.  That `p` is lower case to match with the filename, not the event name.
* Document this in PLUGINS.md.

Unless EDSM is telling us to discard this we should now be sending it.

### DIFF
--- a/monitor.py
+++ b/monitor.py
@@ -151,6 +151,7 @@ class EDLogs(FileSystemEventHandler):  # type: ignore # See below
                 'Item':           defaultdict(int),    # BackPack Items
                 'Data':           defaultdict(int),  # Backpack Data
             },
+            'BackpackJSON':       None,  # Raw JSON from `Backpack.json` file, if available
             'SuitCurrent':        None,
             'Suits':              {},
             'SuitLoadoutCurrent': None,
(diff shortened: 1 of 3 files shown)


Athanasius:
Ah, no, we need to put the loaded data *into* the Journal-file sourced `BackPack` event.

monitor.py
```diff
@@ -866,17 +867,21 @@ def parse_entry(self, line: bytes) -> MutableMapping[str, Any]:  # noqa: C901, C
...
 
-            elif event_type == 'Backpack':
+            elif event_type == 'BackPack':
                 # TODO: v31 doc says this is`backpack.json` ... but Howard Chalkley
                 #       said it's `Backpack.json`
                 with open(join(self.currentdir, 'Backpack.json'), 'rb') as backpack:  # type: ignore
                     try:
-                        entry = json.load(backpack)
+                        # Preserve property order because why not?
+                        entry = json.load(backpack, object_pairs_hook=OrderedDict)
```
A-UNDERSCORE-D, line 876: This should be ordered anyway as normal python dicts are ordered now

Athanasius: I was cribbing from the equivalent in CargoJSON code.

A-UNDERSCORE-D: Ah okay

PR #1089, opened 2021-05-21
Made most event lookups case insensitive

### LINKED ISSUE #1087: Treat Journal event names case-insensitively
I know of no instance of two distinct Journal event names only differing in their case.

Things like `BackPack`, but `Backpack.json` are a pain to keep straight.

So we should refactor to specifically use a `lower()`'d copy of the event name in all matching code.

We should *not* change the actual event name, plugins get to decide how they treat it.

This entails making the `event_name = entry['name'].tolower()` and then using it instead in matches (maybe in logging too?).

This issue is the latest in the series entitled "Ath is getting sick of this shit" ;) .

A-UNDERSCORE-D:
Slightly more than `s/((el)?if event_type (in|==) \(?('(?:.*[A-Z].*).*')\)?:)$/\L$1`.

Closes #1087

### DIFF
--- a/monitor.py
+++ b/monitor.py
@@ -691,7 +691,7 @@ class EDLogs(FileSystemEventHandler):  # type: ignore # See below
                 #     • StationType
                 #     • MarketID
 
-                if entry.get('onstation', False):
+                if entry.get('OnStation', False):
                     self.station = entry.get('StationName', '')
 
                 else:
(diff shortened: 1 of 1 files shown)


Athanasius:
We won't rush this into 5.0.2 - also I only glanced, and want to be sure `entry_name` isn't used other than in the matching conditionals, i.e. side effects.

A-UNDERSCORE-D:
I did take a look for that. I didnt see anything but agreed it definitely needs to be double checked.

Athanasius:
Closed in favour of #1174

PR #1090, opened 2021-05-21
Don't spam a stacktrace on requests.ConnectionError

### LINKED ISSUE #1082: Catch name resolution errors to have slightly less log spam
CAPI DNS name is MIA currently, rather than have a huge multi-level exception for this, catch it and log something more succinct.

Review same for EDSM, EDDB etc too.

A-UNDERSCORE-D:
ConnectionErrors are expected and not something we can fix. The
exception handler that was catching these previously is a catchall
intended to stop EDMC as a whole from crashing from something
unexpected.

Closes #1082

### DIFF
--- a/EDMarketConnector.py
+++ b/EDMarketConnector.py
@@ -979,6 +979,11 @@ class AppWindow(object):
             companion.session.invalidate()
             self.login()
 
+        except companion.ServerConnectionError as e:
+            logger.debug(f'Exception while contacting server: {e}')
+            err = self.status['text'] = str(e)
+            play_bad = True
+
         except Exception as e:  # Including CredentialsError, ServerError
             logger.debug('"other" exception', exc_info=e)
             err = self.status['text'] = str(e)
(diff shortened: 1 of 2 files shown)


EDMarketConnector.py
```diff
@@ -967,6 +967,11 @@ def getandsend(self, event=None, retrying: bool = False):  # noqa: C901, CCR001
             companion.session.invalidate()
             self.login()
 
+        except companion.ServerConnectionError as e:
+            logger.debug(f'Exception while contacting server: {e}')
```
Athanasius, line 971: I think these should probably be logger.warning.

companion.py
```diff
@@ -539,6 +543,10 @@ def query(self, endpoint: str) -> CAPIData:
             logger.trace('Trying...')
             r = self.session.get(self.server + endpoint, timeout=timeout)  # type: ignore
 
+        except requests.ConnectionError as e:
+            logger.debug(f'Unable to resolve name for CAPI: {e} (for request: {endpoint})')
```
Athanasius, line 547: I think these should probably be logger.warning.

Athanasius (changes requested):
I should probably check if we need to define under what circumstances we use logger.warning and logger.error in Contributing.md ....

Athanasius (approved):

Issue #24, opened 2015-09-02
Translations welcome


Marginal:
If you'd like this app translated into your language please join the [POEditor project](https://poeditor.com/join/project/iX7FXzR12n) and add or update your language.

koreldan:
Italian done

Marginal:
> Italian done

Thanks. Added your changes in cafbde8.

Still welcome for more translations - particularly Spanish and Russian.

Uriei:
I could help on the Spanish one, but I don't play on Spanish client because I like to play in English :P

bubis7:
I can translate to Latvian, and get some of my friends to do it in Russian maybe. ;)

Marginal:
Please do join the [OneSky project](https://marginal.oneskyapp.com/collaboration/project?id=52710) if you would like to make a translation. There are currently 147 phrases, but about half of those are just the Federation/Empire/Combat/etc ranks.

bubis7:
Could you add the languages you want to be translated, or is it done by me? I dont see an a option to add Latvian or Russian.

Marginal:
I've added Latvian, Russian and Spanish.
I'll fill in the Russian ranks with [info](http://ru.elite-dangerous.wikia.com/wiki/%D0%A0%D0%B0%D0%BD%D0%B3%D0%B8) from the Russian Elite wiki.

Do you not see a "Manage Languages" option in the left-hand bar? (Sorry, this is the first time I'm using OneSky).

bubis7:
No im not seeing them, I assume that`s because i`m not a participant in that project, but then again, i dont see a option to participate in it.

Marginal:
I've just logged in under another account, and I see what you mean.
Anyway I've added Latvian and Russian so you should now be able to click on them and start editing.

nnemkin:
Hey, I've updated Russian translation, except a few OS X specific strings (no Macs around to check for "official" translations).

Marginal:
> I've updated Russian translation, except a few OS X specific strings (no Macs around to check for "official" translations).

Thanks. Given that on OSX hotkey/shortcuts are described as Сочетание клавиш, can you translate the [remaining two](https://marginal.oneskyapp.com/collaboration/translate/project/project/52710/language/11#/?keyword=shortcut)?

nnemkin:
I just did that, should be close enough.

Marginal:
Thanks. Added Russian in 82b1702.

[...]

PR #1108, opened 2021-05-25
Suits: Ensure a succinct and sane name for Suits in UI

Athanasius:
The aim here is to use the well-localised CAPI-sourced names whenever we can.

Further work in this branch/PR will look at constructing a lookup table/regex to map the symbolic names to **English** readable names.

### DIFF
--- a/edmc_data.py
+++ b/edmc_data.py
@@ -499,20 +499,28 @@ ship_name_map = {
 # Odyssey Suit Names
 edmc_suit_shortnames = {
     'Flight Suit':      'Flight',     # EN
-    'Flug-Anzug':       'Flug',       # DE
-    'Летный комбинезон': 'Летный',    # RU
-    'Traje de vuelo':   'de vuelo',   # ES
     'Artemis Suit':     'Artemis',    # EN
-    'Artemis-Anzug':    'Artemis',    # DE
-    'Traje Artemis':    'Artemis',    # ES
-    'Комбинезон Artemis': 'Artemis',  # RU
     'Dominator Suit':   'Dominator',  # EN
-    'Dominator-Anzug':  'Dominator',  # DE
-    'Traje Dominator':  'Dominator',  # ES
-    'Комбинезон Dominator': 'Dominator',  # RU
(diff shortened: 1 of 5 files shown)


Athanasius:
This *should* be ready now, although I'm seeing if anyone leaves feedback on the translations issue: https://github.com/EDCD/EDMarketConnector/issues/24#issuecomment-848835061

edmc_data.py
```diff
@@ -495,3 +495,77 @@
...
+
+    'Traje voador':           'voador',     # PT-BR
+    'Traje Artemis':          'Artemis',    # PT-BR
+    'Traje Dominator':        'Dominator',  # PT-BR
+    'Traje Maverick':         'Maverick',   # PT-BR
+
+    'Летный комбинезон':      'Летный',     # RU
+    'Комбинезон Artemis':     'Artemis',    # RU
+    'Комбинезон Dominator':   'Dominator',  # RU
+    'Комбинезон Maverick':    'Maverick',   # RU
```
A-UNDERSCORE-D, line 529: it LOOKS to me as if their localisation for this is just the the word suit in the language and the name of the suit otherwise. Which I guess makes sense? but that being the case we could probably drop this for everything but flight?

A-UNDERSCORE-D (approved):
Looks okay other than that comment

edmc_data.py
Athanasius: I feel that doing the two stages (symbol to actual game translation, then that to the short form) gives us the greater flexibility.  This way we, and plugins, can use the 'full form' if they want, and also have this single agreed upon short form as well.

I can see an argument for doing the mapping straight from symbol to short form, rather than via intermediary.  In part I'm being paranoid about Frontier changing a translation to a form where this shortening *doesn't* make sense.  We'd be left with the long form translation as inn-game then, rather than effectively forcing an English name.

A-UNDERSCORE-D (approved):

edmc_data.py
A-UNDERSCORE-D: Yeah in that case leaving it is the best idea.

PR #1117, opened 2021-05-27
Add configuration option for coriolis.io instance to use

A-UNDERSCORE-D:


A-UNDERSCORE-D:
Added a dropdown to select which one to go with, if not auto.

plugins/coriolis.py
```diff
@@ -13,13 +22,101 @@
 
 config.delete('shipyard', suppress=True)
 
+logger = get_main_logger()
+
+DEFAULT_NORMAL_URL = 'https://coriolis.io/import?data='
+DEFAULT_BETA_URL = 'https://beta.coriolis.io/import?data='
+DEFAULT_OVERRIDE_MODE = 'auto'
```
Athanasius, line 29: This starts out as 'auto' in the dropdown, but in the dropdown there's "normal", "beta" and **"default"**.  So one of 'auto' or 'default' needs to change.

plugins/coriolis.py
```diff
@@ -13,13 +22,101 @@
...
+    cur_row += 1
+
+    nb.Label(conf_frame, text=_('Normal URL')).grid(sticky=tk.W, row=cur_row, column=0, padx=PADX)
+    nb.Entry(conf_frame, textvariable=normal_textvar).grid(sticky=tk.EW, row=cur_row, column=1, padx=PADX)
+    cur_row += 1
+
+    nb.Label(conf_frame, text=_('Beta URL')).grid(sticky=tk.W, row=cur_row, column=0, padx=PADX)
+    nb.Entry(conf_frame, textvariable=beta_textvar).grid(sticky=tk.EW, row=cur_row, column=1, padx=PADX)
+    cur_row += 1
+
```
Athanasius, line 73: These each need a 'Default' button on the right hand side to make it *easy* to revert.

plugins/coriolis.py
A-UNDERSCORE-D: whoops, it should be auto

plugins/coriolis.py
A-UNDERSCORE-D: Added

plugins/coriolis.py
A-UNDERSCORE-D: Changed

Athanasius: s/auto/Auto/ - else first time you look it shows this "auto" as the current value, when it's "Auto" in the dropdown.

plugins/coriolis.py
```diff
@@ -13,13 +23,120 @@
...
+def _get_target_url(is_beta: bool) -> str:
+    global override_mode
+    if override_mode not in ('auto', 'normal', 'beta'):
+        show_error(_('Invalid Coriolis override mode!'))
+        logger.warning(f'Unexpected override mode {override_mode!r}! defaulting to auto!')
+        override_mode = 'auto'
+
+    if override_mode == 'beta':
+        return beta_url
+    elif override_mode == 'normal':
```
Athanasius, line 125: \*gasp\* - no blank line ?

L10n/en.template
```diff
@@ -601,3 +601,30 @@
...
+"Reset" = "Reset";
+
+/* Coriolis Normal URL label [plugins/coriolis.py] */
+"Normal URL" = "Normal URL";
+
+/* Coriolis Beta URL label [plugins/coriolis.py] */
+"Beta URL" = "Beta URL";
+
+/* Coriolis config title label [plugins/coriolis.py] */
+"Set the URL to use with coriolis.io ship loadouts. Note that this MUST end with 'import?data='" = "Set the URL to use with coriolis.io ship loadouts. Note that this MUST end with 'import?data='";
```
Athanasius, line 627: The code has `/import` here, with the `/`.

plugins/coriolis.py
```diff
@@ -13,13 +23,121 @@
...
+    conf_frame.columnconfigure(index=1, weight=1)
+    cur_row = 0
+    nb.Label(conf_frame, text=_(
+        "Set the URL to use with coriolis.io ship loadouts. Note that this MUST end with '/import?data='"
+    )).grid(sticky=tk.EW, row=cur_row, column=0, columnspan=3)
+    cur_row += 1
+
+    nb.Label(conf_frame, text=_('Normal URL')).grid(sticky=tk.W, row=cur_row, column=0, padx=PADX)
+    nb.Entry(conf_frame, textvariable=normal_textvar).grid(sticky=tk.EW, row=cur_row, column=1, padx=PADX)
+    nb.Button(conf_frame, text=_("Reset"), command=lambda: normal_textvar.set(value=DEFAULT_NORMAL_URL)).grid(
```
Athanasius, line 71: `Reset` needs adding to translation template.

A-UNDERSCORE-D: its in there

Issue #569, opened 2020-07-08
[Build] Move any "in case plugins" imports into setup.py


Athanasius:
e.g. companion.py has

`from http.cookiejar import LWPCookieJar  # No longer needed but retained in case plugins use it`

That should really just be listed in `setup.py` for inclusion at build time then.  Of course it's being imported as just `LWPCookieJar` not the full name, but specifically this import here will be *why* py2exe picks it up for packaging at all.

- [ ] Fix this instance up to work out the necessary changes.
- [ ] Look for other instances.

PR #1123, opened 2021-05-28
Cleanup imports where possible, remove incase imports

A-UNDERSCORE-D:
closes EDCD/EDMarketConnector#569

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -185,7 +185,6 @@ elif sys.platform == 'win32':
             'optimize': 2,
             'packages': [
                 'sqlite3',  # Included for plugins
-                'http',  # Included for plugins
             ],
             'includes': [
                 'dataclasses',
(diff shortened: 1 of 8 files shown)


A-UNDERSCORE-D:
Tested that migrating imports worked with the following test file:

```py
from http.cookiejar import LWPCookieJar
from threading import Thread

def plugin_start3(x):
    print("STARTED!!!!")
```

setup.py
```diff
@@ -190,6 +190,7 @@ def get_cfbundle_localizations() -> Set:
                 'dataclasses',
                 'shutil',  # Included for plugins
                 'timeout_session',
+                'threading',  # Included for plugins (though it should always be there anyway)
```
Athanasius, line 193: This too can be removed.  Core code definitely uses it and is highly unlikely to stop that.

PR #1130, opened 2021-05-31
Update INARA plugin to support new odyssey API endpoints

### LINKED ISSUE #1127: Implement new/changed Inara events 2021-05
From Discord:

> Artie (Inara):@Robby (EDDiscovery) @Athanasius (EDMC, ed.miggy.org) @T'kael (EDDI) and anybody who may be interested in it. :wink:

Inara API update for Odyssey:
- The header property isDeveloped was renamed to less confusing isBeingDeveloped. You don't need to change anything on your end though, the old property will still work.
- A new event addCommanderTravelLand for commanders landing on the celestial body was added (should be triggered at journal's Touchdown, DropShipDeploy and similar).
- The events addCommanderTravelCarrierJump, addCommanderTravelDock, addCommanderTravelFSDJump and setCommanderTravelLocation now contain optional properties for the star system coordinates, celestial body and celestial body coordinates. It's not required to set, it may however help to further precise commander's location when the star system is not in the database, etc.
- The events addCommanderTravelDock, addCommanderTravelFSDJump and addCommanderTravelLand are now having boolean properties isTaxiShuttle and isTaxiDropship for travelling with Apex or Frontline Solutions ships.
- New events setCommanderInventory, resetCommanderInventory, setCommanderInventoryItem, addCommanderInventoryItem and delCommanderInventoryItem added. Those are supposed for handling the new Odyssey events like ShipLockerMaterials and similar, you can however use it also for the existing engineering materials or ship cargo instead of the existing dedicated events (but you do not need to, they are not going anywhere).
- New events setCommanderSuitLoadout, updateCommanderSuitLoadout and delCommanderSuitLoadout for the suit loadouts added.

The documentation is as usually here: https://inara.cz/inara-api-docs/

A-UNDERSCORE-D:
To Do:
- [x] `addCommanderTravelLand` support
    - I believe everything we want is done here.
- [x] Update existing events with new taxi information
    - [x] `addCommanderTravelCarrierJump`
    - [x] `addCommanderTravelDock`
    - [x] `addCommanderTravelFSDJump`
    - [ ] `setCommanderTravelLocation`
        - I dont think there are any actual updates we can do here?
- [x] `setCommanderInventory` support
    - This SHOULD be done, it currently only sends on ShipLockerMaterials as thats the only time we can be sure data is correct.
- [x] Suit Loadouts
    - [x] `setCommanderSuitLoadout`
    - [x] `updateCommanderSuitLoadout`
    - [x] `delCommanderSuitLoadout`

Do **BEFORE** merging
- [x] Currently log is spammed to remind me to disable development mode in the headers we send
- [x] Disable development mode in the headers we send

Closes #1127

### DIFF
--- a/plugins/inara.py
+++ b/plugins/inara.py
@@ -20,8 +20,6 @@ import timeout_session
 from companion import CAPIData
 from config import applongname, appversion, config
 from EDMCLogging import get_main_logger
-# Yes I know. Im using it for `monitor.planet` Which probably should be rolled into state as a body and bodyType pair
-from monitor import monitor
 from ttkHyperlinkLabel import HyperlinkLabel
 
 logger = get_main_logger()
@@ -1046,7 +1044,7 @@ def journal_entry(  # noqa: C901, CCR001
             to_send_data: Optional[Dict[str, Any]] = {}  # This is a glorified sentinel until lower down.
             # On Horizons, neither of these exist on TouchDown
             star_system_name = entry.get('StarSystem', this.system)
-            body_name = entry.get('Body', monitor.planet)
+            body_name = entry.get('Body', state['Body'] if state['BodyType'] == 'Planet' else None)
 
             if star_system_name is None:
                 logger.warning('Refusing to update addCommanderTravelLand as we dont have a StarSystem!')
@@ -1150,7 +1148,8 @@ def journal_entry(  # noqa: C901, CCR001
                         'slotName': entry['SlotName'],
                         'itemName': entry['ModuleName'],
                         'itemGameID': entry['SuitModuleID'],
-                        # TODO: As of 4.0.0.200, this event does *NOT* include the class
+                        'itemClass': entry['Class'],
+                        'engineering': [],  # TODO: Check casing of names for this
                     }
                 ],
             }
--- a/plugins/inara.py
+++ b/plugins/inara.py
@@ -1432,7 +1432,6 @@ def new_worker():
                     'APIkey': creds.api_key,
                     'commanderName': creds.cmdr,
                     'commanderFrontierID': creds.fid,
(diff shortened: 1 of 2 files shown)


Athanasius:
NB: No implementations from `Disembark` events because they'll always be paired with either a `Docked` or `Touchdown` event.

plugins/inara.py
```diff
@@ -975,7 +1018,146 @@ def journal_entry(  # noqa: C901, CCR001
...
+                if state.get('ShipType') is not None:
+                    to_send_data['shipType'] = state['ShipType']
+
+                to_send_data['isTaxiShuttle'] = False
+                to_send_data['isTaxiDropShip'] = False
+
+                new_add_event('addCommanderTravelLand', entry['timestamp'], to_send_data)
+
+        elif event_name == 'ShipLockerMaterials':
+            odyssey_nonplural_microresource_types = ('Items', 'Components', 'Data', 'Consumables')
```
Athanasius, line 1090: These look distinctly plural to me ?

Athanasius:
NB: Only implementing the microresource tracking for `ShipLockerMaterials` as that's a definite total of current inventory, whereas the other events are known to be bugged in 4.0.0.200.

Athanasius:
> `setCommanderTravelLocation`
> 
>     * I dont think there are any actual updates we can do here?

We can set this from journal `Location` can't we?  I think that might indeed be the intent behind the API call.  For those times when someone might not have been running EDMC during the previous play, or Inara was down, or network issues, or ... such that we should 'catch' up now.

Athanasius:
Re: setCommanderTravelLocation - there might be a bug at the Inara end with not updating Cmdr profiles from this.  See https://discord.com/channels/102002327619121152/376773926220595201/850326733843857428 (Inara Discord)

Issue #1138, opened 2021-06-04
EDMC fails on an empty backpack.json file


lekeno:
**Please complete the following information:**
 - Version 5.0.4
 - Odyssey (latest patch, Update 2 + small patch)
 - OS: Windows 10
 - OS Locale: English
 - Log file: on a different machine, but the issue is that EDMC is trying to JSON parse a file, and that call breaks on empty files. As for why Elite produced an empty backpack.json in the first place:  ¯\_(ツ)_/¯

**Describe the bug**
EDMC crashed if backpack.json is empty. Possibly similar issues with other .json files.

A-UNDERSCORE-D:
Im inclined to leave this in explode mode, because we only try to read that file _after_ we're told it contains data

Athanasius:
> Im inclined to leave this in explode mode, because we only try to read that file _after_ we're told it contains data

On the other hand catching it and logging a single warning line is less to scroll through when checking for other stuff ....

PR #1139, opened 2021-06-04
Fixed erroring on empty backpack.json

A-UNDERSCORE-D:
closes EDCD/EDMarketConnector#1138

### DIFF
--- a/monitor.py
+++ b/monitor.py
@@ -883,21 +883,27 @@ class EDLogs(FileSystemEventHandler):  # type: ignore # See below
                 #       said it's `Backpack.json`
                 backpack_file = pathlib.Path(str(self.currentdir)) / 'Backpack.json'
                 backpack_data = None
-                if backpack_file.exists():
-                    backpack_data = backpack_file.read_bytes()
-                else:
+
+                if not backpack_file.exists():
                     logger.warning(f'Failed to find backpack.json file as it appears not to exist? {backpack_file=}')
+
+                else:
+                    backpack_data = backpack_file.read_bytes()
 
                 parsed = None
 
-                if backpack_data is not None and len(backpack_data) > 0:
-                    try:
-                        parsed = json.loads(backpack_data)
-                    except json.JSONDecodeError:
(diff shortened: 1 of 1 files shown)


monitor.py
```diff
@@ -880,43 +881,54 @@ def parse_entry(self, line: bytes) -> MutableMapping[str, Any]:  # noqa: C901, C
...
-                        clean_items = self.coalesce_cargo(entry['Items'])
-                        self.state['BackPack']['Item'].update(
-                            {self.canonicalise(x['Name']): x['Count'] for x in clean_items}
-                        )
-
-                        clean_data = self.coalesce_cargo(entry['Data'])
-                        self.state['BackPack']['Data'].update(
-                            {self.canonicalise(x['Name']): x['Count'] for x in clean_data}
-                        )
+                else:
```
Athanasius, line 899: I'd prefer splitting the conditional to separately log each case.

One is "Frontier said there was a file, there wasn't".
Another is "Frontier said there was a file, there was, but it was empty".
Another is "Frontier said there was a file, there was, it had content, but it didn't parse as JSON".

monitor.py
```diff
@@ -880,43 +881,54 @@ def parse_entry(self, line: bytes) -> MutableMapping[str, Any]:  # noqa: C901, C
             elif event_type in ('BackPack', 'Backpack'):  # WORKAROUND 4.0.0.200: BackPack becomes Backpack
                 # TODO: v31 doc says this is`backpack.json` ... but Howard Chalkley
                 #       said it's `Backpack.json`
-                with open(join(self.currentdir, 'Backpack.json'), 'rb') as backpack:  # type: ignore
-                    try:
-                        # Preserve property order because why not?
-                        entry = json.load(backpack, object_pairs_hook=OrderedDict)
+                backpack_file = pathlib.Path(str(self.currentdir)) / 'Backpack.json'
+                backpack_data = None
```
Athanasius, line 885: Something about the 'exception' being on the else irks me.  Invert conditionals in this to log in them, and ... well we're in this huge conditional tree so we can't return :/.

PR #1142, opened 2021-06-04
Made EDDN logs more verbose

### LINKED ISSUE #503: EDDN error reporting needs to be more verbose
Currently if the EDDN plugin can't connect/send to the EDDN Gateway for any reason all you'll see is "Error: Can't connect to EDDN" on the bottom line of the EDMC UI.  Unless you've set debug output active you won't even get extra output in the log file.

This needs to be changed so, for instance, the user can see if the EDDN Gateway is returning a specific error because their version of the client has been banned.

A-UNDERSCORE-D:
This doesnt try and guess if EDMC is banned, it just notes that it might be. 
Closes #503


Athanasius:
Resolve the en.template conflicts and this is good to merge.

PR #1161, opened 2021-06-10
Align ShipLocker and BackPack code with ED 4.0.0.400

Athanasius:
This should now allow for perfect tracking of both ShipLocker and BackPack.

My findings as [on Discord:EDCD:#odyssey-specific](https://discord.com/channels/164411426939600896/825041324121784401/852503268864688136):

4.0.0.400 (2021-06-10 update) findings about odyssey inventory events.

1. `ShipLocker` event written (and the file) on login in-ship in a station.

2. Disembark on body surface.
2a. *Empty* `ShipLocker` event triggers on *Disembark* on a body surface  (not docked).  File written.
2b. Full `Backpack`event written on *Disembark* on body surface.  File written.

3. Disembark at station - the same `ShipLocker` and `Backpack` as for body surface.

4. Transferred 2x Electrical Fuse *to* suit *from* ship using menu whilst on foot.
4a. *Empty* `ShipLocker` event, and file update.
4b. `BackpackChange`/Added or /Removed as appropriate.  *One event per specific inventory changed, whether in the same category or not*.
4c. No `Backpack` event for this, or file update.

5. Boarding ship from body-surface on-foot.
5a. First an empty `ShipLocker` *before* the `Embark` event. File updated?  Too quick to know for sure.
5b. But also then a full `ShipLocker` *after*: `Embark`, `Loadout`, `Music`.  File updated.

6. Throwing a grenade causes a `BackpackChange`/Removed event.  Backpack.json also written with updated full inventory.
6b. Using a medkit, or energy cell, causes both `UseConsumable` and `BackpackChange`/Removed.

7. Picking up an item out of a container.
7a. `CollectItems` event.
7b. `BackpackChange`/Added event as well. File updated (frag grenade from case).

8. Dropping an item.
8a. `DropItems` event.
8b. `BackpackChange`/Removed - file written.

9. Buying more consumables from Pioneer Supplies.
9a. `BuyMicroResource` - you can only buy one type before confirming, so only ever one event.
9b. *Empty* `ShipLocker` and file written.

10. Accepting a mission that provides a Power Regulator. *Three* **empty** `ShipLocker` events written - file written.

11. Trading with bar tender.
11a. *Empty* `ShipLocker` event - file written.
11b. `TradeMicroResources` event.

12. Selling to a bar tender.
12a. *Empty* `ShipLocker` event - file written.
12b. `SellMicroResources` event.

### DIFF
--- a/monitor.py
+++ b/monitor.py
@@ -683,10 +683,7 @@ class EDLogs(FileSystemEventHandler):  # type: ignore # See below
 
                 # We can't now have anything in the BackPack, it's all in the
                 # ShipLocker.
-                self.state['BackPack']['Component'] = defaultdict(int)
-                self.state['BackPack']['Consumable'] = defaultdict(int)
-                self.state['BackPack']['Item'] = defaultdict(int)
-                self.state['BackPack']['Data'] = defaultdict(int)
+                self.backpack_set_empty()
 
             elif event_type == 'Disembark':
                 # This event is logged when the player steps out of a ship or SRV
@@ -934,10 +931,7 @@ class EDLogs(FileSystemEventHandler):  # type: ignore # See below
                     self.state['BackpackJSON'] = entry
 
                     # Assume this reflects the current state when written
-                    self.state['BackPack']['Component'] = defaultdict(int)
-                    self.state['BackPack']['Consumable'] = defaultdict(int)
-                    self.state['BackPack']['Item'] = defaultdict(int)
-                    self.state['BackPack']['Data'] = defaultdict(int)
+                    self.backpack_set_empty()
 
                     clean_components = self.coalesce_cargo(entry['Components'])
                     self.state['BackPack']['Component'].update(
@@ -1587,6 +1581,13 @@ class EDLogs(FileSystemEventHandler):  # type: ignore # See below
         except Exception as ex:
             logger.debug(f'Invalid journal entry:\n{line!r}\n', exc_info=ex)
             return {'event': None}
+
+    def backpack_set_empty(self):
+        """Set the BackPack contents to be empty."""
+        self.state['BackPack']['Component'] = defaultdict(int)
+        self.state['BackPack']['Consumable'] = defaultdict(int)
+        self.state['BackPack']['Item'] = defaultdict(int)
+        self.state['BackPack']['Data'] = defaultdict(int)
 
     def suit_sane_name(self, name: str) -> str:
         """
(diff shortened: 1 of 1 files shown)


monitor.py
```diff
@@ -843,21 +847,30 @@ def parse_entry(self, line: bytes) -> MutableMapping[str, Any]:  # noqa: C901, C
...
                         self.state['Cargo'][name] -= c['Count']
 
-            elif event_type == 'ShipLockerMaterials':
+            elif event_type == 'ShipLocker':
+                # As of 4.0.0.400 (2021-06-10)
+                # "ShipLocker" will be a full list written to the journal at startup/boarding/disembarking, and also
+                # written to a separate shiplocker.json file - other updates will just update that file and mention it
+                # has changed with an empty shiplocker event in the main journal.
+
+                if not all([entry.get(t, False) for t in ('Components', 'Consumables', 'Data', 'Items')]):
```
A-UNDERSCORE-D, line 856: As ShipLocker always writes to the file _anyway_, possibly just always read it?

Additionally, all() can take a generator
```suggestion
                if not all(entry.get(t, False) for t in ('Components', 'Consumables', 'Data', 'Items')):
```

monitor.py
```diff
@@ -1002,87 +987,38 @@ def parse_entry(self, line: bytes) -> MutableMapping[str, Any]:  # noqa: C901, C
...
-
-                # This is handled by BackpackChange.
-                # for i in self.state['BackPack'][entry['Type']]:
-                #     if i == entry['Name']:
-                #         self.state['BackPack'][entry['Type']][i] -= entry['Count']
-                #         # Paranoia in case we lost track
-                #         if self.state['BackPack'][entry['Type']][i] < 0:
-                #             self.state['BackPack'][entry['Type']][i] = 0
+                # 4.0.0.400 (still) has a BackpackChange event as well, so
+                # ignore this for inventory purposes.
```
A-UNDERSCORE-D, line 1021: we can likely pack all of these into a tuple?

A-UNDERSCORE-D: ...thanks github. so very helpful. I meant the events `pass`-ed as we have replacements

monitor.py
Athanasius: Eh, if the data is in the event I'm happy just using that, save the file open/read.  I'll remove those `[]` then.

Athanasius: OK, I relented, we nose always load the file.

PR #1170, opened 2021-06-13
CAPI: Properly catch "CAPI says docked at EDO settlement, but we forgot where"

Athanasius:


companion.py
```diff
@@ -189,6 +190,22 @@ def __init__(self, *args) -> None:
             self.args = (_('Error: Frontier server is lagging'),)
 
 
+class NoMonitorStation(Exception):
```
A-UNDERSCORE-D, line 193: Does this maybe want to inherit from ServerLagging

Athanasius: That feels misleading.  In this case Frontier CAPI server isn't found to be lagging.  This is literally "we forgot the monitor.station, so can't even check".

A-UNDERSCORE-D: thats fair

A-UNDERSCORE-D (approved):

Issue #1185, opened 2021-06-30
Use status.json Balance field for assisting with tracking credit balance


A-UNDERSCORE-D:

A-UNDERSCORE-D:
From Discord

>Athanasius: If that doesn't already contain it ... I'd do something so that monitor.py can check the latest dashboard/status.json credits value and log whenever it doesn't match our total after the event processing is done.  Hmmm, this almost feels like wanting a separate log file to record every change to the balance from Status.json (not value on every read from it).

PR #1186, opened 2021-06-30
Enhancement/odyssey update 5

A-UNDERSCORE-D:
This adds support for the new fields and things added in patch 5. It does not include the status.json changes noted in #1185, and does not add special storage for `LandingPad` fields that exist on some events.

### DIFF
--- a/monitor.py
+++ b/monitor.py
@@ -20,6 +20,7 @@ if TYPE_CHECKING:
 
 import util_ships
 from config import config
+from dashboard import dashboard
 from edmc_data import edmc_suit_shortnames, edmc_suit_symbol_localised
 from EDMCLogging import get_main_logger
 
@@ -517,6 +518,13 @@ class EDLogs(FileSystemEventHandler):  # type: ignore # See below
                 self.live = True  # First event in 3.0
 
             elif event_type == 'loadgame':
+                # Odyssey Release Update 5
+                self.state['GameLanguage'] = entry.get('language', self.state['GameLanguage'])
+                self.state['GameVersion'] = entry.get('gameversion', self.state['GameVersion'])
+                self.state['GameBuild'] = entry.get('build', self.state['GameBuild'])
+                self.version = self.state['GameVersion']  # Update this just in case things above changed.
+                self.is_beta = any(v in str(self.version).lower() for v in ('alpha', 'beta'))
+
                 # alpha4
                 # Odyssey: bool
                 self.cmdr = entry['Commander']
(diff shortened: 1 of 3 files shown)


monitor.py
```diff
@@ -517,6 +517,13 @@ def parse_entry(self, line: bytes) -> MutableMapping[str, Any]:  # noqa: C901, C
                 self.live = True  # First event in 3.0
 
             elif event_type == 'loadgame':
+                # Odyssey Release Update 5
```
Athanasius, line 520: I don't see how we could possible have a `loadgame` without a `fileheader` - we're not pulling journals from the CAPI (which doesn't have `fileheader` and that is why the fields have been added to `loadgame` as well).

So this seems un-necessary.

monitor.py
```diff
@@ -1110,6 +1117,7 @@ def parse_entry(self, line: bytes) -> MutableMapping[str, Any]:  # noqa: C901, C
                     'edmcName':  self.suit_sane_name(loc_name),
                     'id':        None,  # Is this an FDev ID for suit type ?
                     'suitId':    entry['SuitID'],
+                    'mods':      entry['SuitMods'],  # Suits can (rarely) be bought with modules installed
```
Athanasius, line 1120: 'slots' *is* 'mods', isn't it ?

Because initially the data was only available in the CAPI `/profile` return I went with what it calls things.  See e.g. the code for `loadoutequipmodule`.

Athanasius (changes requested):
This definitely needs to review the slots vs mods thing.  I *could* be misremembering.

Athanasius:
Btw, I can be convinced that we should now switch over to 'mods' instead of 'slots' for the suit and weapon mods.  I'd wager no plugin developer has even made use of these bits yet so I'd feel fine breaking it without a major version bump.

monitor.py
A-UNDERSCORE-D: Put there as a "just in case", that said, it can be easily removed now that its been DRY-ed

monitor.py
Athanasius: So, we're moving towards using the Journal 'mods' for this.

But need to be sure the CAPI imported data isn't an issue.  When I checked this sort of 'slots' is definitely used on a weapon for things like Optics (the default one on TK Aphelion), but 'modifications' (with the naff 'NYI' content) seems to be what's used for installable mods.

So, we'll still have a slighty mapping issue.

A-UNDERSCORE-D: whoops didnt mean to resolve that. This is whats in the code now, no plugin devs have screamed as of yet about it.

Athanasius: Discord discussion has lead me to conclude that `'mods'` is the correct name for this.  The CAPI data is a mess, follow Journal.

PR #1191, opened 2021-07-07
Extend killswitch functionality to fields within events

A-UNDERSCORE-D:
This PR adds the ability for killswitches to simply modify data either before or after processing to prevent data leakage.

Fields can be deleted, redacted (set to the string "REDACTED") or directly set with values given. 
If a killswitch has any kind of action, it will be attempted, and if it succeeds, the event processing with continue with modified data. Thus a simple killswitch with no additional rules simply blocks processing entirely.

### DIFF
--- a/docs/Killswitches.md
+++ b/docs/Killswitches.md
@@ -14,10 +14,10 @@ Killswitches are stored in a JSON file that is queried by EDMC on startup. The f
 
 The `kill_switches` array contains kill switch objects. Each contains the following fields:
 
-|       Key |          Type          | Description                                                                  |
-| --------: | :--------------------: | :--------------------------------------------------------------------------- |
-| `version` |     `version spec`     | The version of EDMC these kill switches apply to (Must be valid semver spec) |
-|   `kills` | `Dict[str, Dict[...]]` | The various keys disabled -> definition of the killswitch behaviour          |
+|       Key |             Type              | Description                                                                  |
+| --------: | :---------------------------: | :--------------------------------------------------------------------------- |
+| `version` | `version spec (see Versions)` | The version of EDMC these kill switches apply to (Must be valid semver spec) |
+|   `kills` |    `Dict[str, Dict[...]]`     | The various keys disabled -> definition of the killswitch behaviour          |
 
 Each entry in `kills` must contain at least a `reason` field describing why the killswitch was added. EDMC will show
 this to the user (for internal killswitches, anyway).
@@ -66,7 +66,9 @@ An example follows:
 
 ### Versions
 
-Versions are checked using contains checks on `semantic_version.SimpleSpec` instances.
+Versions are checked using contains checks on `semantic_version.SimpleSpec` instances. SimpleSpec supports both specific
+versions (`1.2.3`), non-specific ranges (`1.0` will match `1.0.1` and `1.0.5` etc), wildcards (`1.2.*`),
+and ranges (`<1.0.0`, `>=2.0.0`)
 
 ## Plugin support
 
--- a/docs/Killswitches.md
+++ b/docs/Killswitches.md
@@ -31,8 +31,8 @@ killswitch was added. EDMC will show this to the user
 | -----------------: | :--------------: | :-------------------------------------------------------------------------------------------- |
 |          `reason`* |      `str`       | The reason that this killswitch was added                                                     |
 |       `set_fields` | `Dict[str, Any]` | A map of key -> contents to update (or overwrite) existing data with                          |
-|    `redact_fields` |   `List[str]`    | A list of fields to redact. This is equivalent to setting the fields to the string "REDACTED" |
-|    `delete_fields` |   `List[str]`    | A list of fields in the matching event to be removed, if they exist.                          |
+|    `redact_fields` |   `List[str]`    | A list of traversal paths to redact. This is the same as using set with a value of "REDACTED" |
+|    `delete_fields` |   `List[str]`    | A list of traversal paths in the matching event to be removed, if they exist.                 |
 
 The order listed above is the precedence for actions. i.e. All set fields are
 set, then all redact fields are redacted then all delete fields are deleted.
@@ -70,6 +70,52 @@ An example follows:
 - `plugins.some_plugin.some_thing` will never be allowed to continue
   (as all fields are blank)
 
+Indexing of lists (and any other `Sequence`) is done with numbers as normal.
+Negative numbers do work to reference the end of sequences.
+
+### Traversal paths
+
+`set_fields`, `delete_fields`, and `redact_fields` all accept a single traversal
+path to target particular fields.
+
+When following paths, there are some caveats one should know.
+
+- If a field (for example `a.b`) exists in the data dict
+  (ie, `{'a.b': True}`) it will be the thing accessed by a killswitch that
+  includes `a.b` at the current "level". This means that if both a key `a.b`
+  and a nested key exist with the same name (`{'a.b': 0, 'a': { 'b': 1 }}`),
+  the former is always referenced. No backtracking takes place to attempt to
+  resolve conflicts, even if the dotted field name is not at the end of a key.
+
+- Traversal can and will error, especially if you attempt to pass though keys
+  that do not exist (eg past the end of a `Sequence` or a key that does not
+  exist in a `Mapping`).
+
+- If any exception occurs during your traversal, **the entire killswitch is**
+  **assumed to be faulty, and will be treated as a permanent killswitch**
+
+### Individual action rules
+
+All actions (as noted above) can fail during traversal to their targets. If this
+happens, the entire killswitch is assumed to be a permanent one and execution
+stops.
+
+#### Deletes
+
+For both `MutableMapping` and `MutableSequence` targets, deletes never fail.
+If a key doesn't exist it simply is a no-op.
+
+#### Sets
+
+Sets always succeed for `MutableMapping` objects, either creating new keys
+or updating existing ones. For `MutableSequence`s, however, a set can fail if
+the index is not within `len(target)`. If the index is exactly `len(target)`,
+`append` is used.
+
+You can set values other than strings, but you are limited to what json itself
+supports, and the translation thereof to python. In general this means:
+`int`, `float`, `string`, `bool`, and `None` (json `null`) values may be set.
+
 ### Testing
 
 Killswitch files can be tested using the script in `scripts/killswitch_test.py`.
@@ -86,7 +132,7 @@ and ranges (`<1.0.0`, `>=2.0.0`)
 ## Plugin support
 
 Plugins may use the killswitch system simply by hosting their own version of the
-killswitch file, and fetching it using 
+killswitch file, and fetching it using
 `killswitch.get_kill_switches(target='https://example.com/myplugin_killswitches.json')`.
 The returned object can be used to query the kill switch set, see the docstrings
 for more information on specifying versions.
@@ -125,7 +171,7 @@ would not be in a form that could be easily understood (except to blank it)
 
 ## File location
 
-The main killswitch file is kept in the `releases` branch on the EDMC github
-repo. The file should NEVER be committed to any other repos. In the case that
-the killswitch file is found in other repos, the one in releases should always
-be taken as correct regardless of others.
+The main killswitch file (`killswitches_v2.json`) is kept in the `releases`
+branch on the EDMC github repo. The file should NEVER be committed to any other
+repos. In the case that the killswitch file is found in other repos, the one in
+releases should always be taken as correct regardless of others.
--- a/docs/Killswitches.md
+++ b/docs/Killswitches.md
@@ -94,12 +94,15 @@ The current recognised (to EDMC and its internal plugins) killswitch strings are
 | :------------------------------------------- | :---------------------: | :---------------------------------------------------------------------------------------- |
 | `plugins.eddn.send`                          |          eddn           | Disables all use of the send method on EDDN (effectively disables EDDN updates)           |
 | `plugins.<plugin>.journal`                   | eddn, inara, edsm, eddb | Disables all journal processing for the plugin                                            |
-| `plugins.<plugin>.worker`                    |       edsm, inara       | Disables the plugins worker thread (effectively disables updates) (does not close thread) |
+| `plugins.<plugin>.worker`                    |      edsm, *inara       | Disables the plugins worker thread (effectively disables updates) (does not close thread) |
 | `plugins.<plugin>.worker.<eventname>`        |       edsm, inara       | Disables the plugin worker for the given eventname                                        |
 | `plugins.<plugin>.journal.event.<eventname>` |    eddn, inara, edsm    | Specific events to disable processing for.                                                |
 
 Killswitches marked with `*` do **not** support modification of their values via set/redact/delete. And as such any match
 will simply stop processing.
+
+For `plugin.inara.worker`, events are checked individually later by the eventname version. Use that to modify individual
+inara events
 
 ## File location
 
--- a/docs/Killswitches.md
+++ b/docs/Killswitches.md
@@ -6,24 +6,27 @@ EDMarketConnector implements a Kill Switch system that allows us to disable feat
 
 Killswitches are stored in a JSON file that is queried by EDMC on startup. The format is as follows:
 
-|             Key |   Type   | Description                                                   |
-| --------------: | :------: | :------------------------------------------------------------ |
-|       `version` |  `int`   | the version of the Kill Switch JSON file, always 1            |
-|  `last_updated` | `string` | When last the kill switches were updated (for human use only) |
-| `kill_switches` | `array`  | The kill switches this file contains (expanded below)         |
+|             Key |   Type   | Description                                                                                  |
+| --------------: | :------: | :------------------------------------------------------------------------------------------- |
+|       `version` |  `int`   | the version of the Kill Switch JSON file, always 2, 1 exists and will be upgraded if needed. |
+|  `last_updated` | `string` | When last the kill switches were updated (for human use only)                                |
+| `kill_switches` | `array`  | The kill switches this file contains (expanded below)                                        |
 
 The `kill_switches` array contains kill switch objects. Each contains the following fields:
 
-|       Key |            Type             | Description                                                                  |
-| --------: | :-------------------------: | :--------------------------------------------------------------------------- |
-| `version` |       `version spec`        | The version of EDMC these kill switches apply to (Must be valid semver spec) |
-|   `kills` | `Dict[str, Dict[str, Any]]` | The different keys to disable, and the reason for the disable                |
+|       Key |          Type          | Description                                                                  |
+| --------: | :--------------------: | :--------------------------------------------------------------------------- |
+| `version` |     `version spec`     | The version of EDMC these kill switches apply to (Must be valid semver spec) |
+|   `kills` | `Dict[str, Dict[...]]` | The various keys disabled -> definition of the killswitch behaviour          |
 
-Each entry in `kills` contains a `reason` and `additional_data` field. `reason` is self explanatory, however
-additional_data can contain various things. They are outlaid below:
-|             Key |    Type     | Description                                                          |
-| --------------: | :---------: | :------------------------------------------------------------------- |
-| `delete_fields` | `List[str]` | A list of fields in the matching event to be removed, if they exist. |
+Each entry in `kills` must contain at least a `reason` field describing why the killswitch was added. EDMC will show
+this to the user (for internal killswitches, anyway).
+| Key (* = required) |       Type       | Description                                                                                   |
+| -----------------: | :--------------: | :-------------------------------------------------------------------------------------------- |
+|          `reason`* |      `str`       | The reason that this killswitch was added                                                     |
+|    `delete_fields` |   `List[str]`    | A list of fields in the matching event to be removed, if they exist.                          |
+|       `set_fields` | `Dict[str, Any]` | A map of key -> contents to update (or overwrite) existing data with                          |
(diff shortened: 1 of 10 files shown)


A-UNDERSCORE-D:
This is untested and needs unit tests, dont merge until then.

A-UNDERSCORE-D:
Tests added and passed -- good for review, then we need to add the new files before merging

Athanasius (changes requested):
I've only looked at the docs so far, but might as well get you moving on these changes whilst I look at the actual code.

docs/Killswitches.md
```diff
@@ -6,29 +6,47 @@ EDMarketConnector implements a Kill Switch system that allows us to disable feat
...
-    "version": 1,
-    "last_updated": "19 October 2020",
+    "version": 2,
+    "last_updated": "3 July 2021",
     "kill_switches": [
         {
             "version": "1.0.0",
             "kills": {
-                "plugins.eddn.send": "some reason"
+                "plugins.eddn.send": {
```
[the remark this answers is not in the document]
A-UNDERSCORE-D: include none of them, `{}`

docs/Killswitches.md
```diff
@@ -45,15 +63,19 @@ Plugins may use the killswitch system simply by hosting their own version of the
 using `killswitch.get_kill_switches(target='https://example.com/myplugin_killswitches.json')`. The returned object can
 be used to query the kill switch set, see the docstrings for more information on specifying versions.
 
+The version of the JSON file will be automatically upgraded if possible by the code KillSwitch code.
```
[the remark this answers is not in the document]
A-UNDERSCORE-D: yeah it just upgrades the format internally, it wont change behaviour

docs/Killswitches.md
```diff
@@ -45,15 +63,19 @@ Plugins may use the killswitch system simply by hosting their own version of the
...
+| :----------------------------------------------------- | :--------------------------------------------------------------------------------------------------------------------------------------------- |
+| `plugins.eddn.send`                                    | Disables all use of the send method on EDDN (effectively disables EDDN updates)                                                                |
+| `plugins.(eddn\|inara\|edsm\|eddb).journal`            | Disables all journal processing for EDDN/EDSM/INARA                                                                                            |
+| `plugins.(edsm\|inara).worker`                         | Disables the EDSM/INARA worker thread (effectively disables updates) (does not close thread)                                                   |
+| `plugins.(edsm\|inara).worker.$eventname`              | Disables the EDSM/INARA worker for the given eventname, OR if delete_fields exists, removes the fields from the event                          |
```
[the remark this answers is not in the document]
A-UNDERSCORE-D: > Also, you're using | just to show here that it could apply to any of the core plugins, rather than as a possible syntax in the actual killswitch right?
Yeah, I have a bad habit of sticking regex where it doesnt belong

>PyCharm's MK rendering barfs on this, still interpreting the | as a table column separator.
sounds like a bug report to me :D but yes Ill fix it

killswitch.py
```diff
@@ -1,54 +1,231 @@
...
     """Queryable set of kill switches."""
 
-    def __init__(self, kill_switches: List[KillSwitch]) -> None:
+    def __init__(self, kill_switches: List[KillSwitches]) -> None:
         self.kill_switches = kill_switches
 
-    def get_disabled(self, id: str, *, version=_current_version) -> DisabledResult:
+    def get_disabled(self, id: str, *, version: Union[Version, str] = _current_version) -> DisabledResult:
         """
         Return whether or not the given feature ID is disabled by a killswitch for the given version.
```
Athanasius, line 213: flake8: no blank line after function docstring (yes, I hate the inconsistency with *requiring* such after class docstrings).

killswitch.py
```diff
@@ -1,54 +1,231 @@
...
+        else:
+            target[key] = to_set
+
+    elif isinstance(target, MutableSequence):
+        idx = _get_int(key)
+        if idx is None:
+            raise ValueError(f'Cannot use string {key!r} as int for index into Sequence')
+
+        if delete:
+            if len(target) > idx:
```
Athanasius, line 85: This doesn't catch where a *negative* index is too (negatively) large.

killswitch.py
```diff
@@ -1,54 +1,231 @@
...
+                target.pop(idx)
+
+        elif len(target) == idx:
+            target.append(to_set)
+
+        else:
+            target[idx] = to_set  # this can raise, that's fine
```
Athanasius, line 92: Negative index also an issue here.

We could decide to just not support them, but I could see the usefulness given it's valid python slicing syntax.

killswitch.py
```diff
@@ -1,54 +1,231 @@
...
+
+    else:
+        raise ValueError(f'Dont know how to apply data to {type(target)} {target!r}')
+
+
+def _deep_apply(target: UPDATABLE_DATA, path: str, to_set=None, delete=False):  # noqa: CCR001 # Recursive silliness.
+    """
+    Set the given path to the given value, if it exists.
+
+    if the path has dots (ascii period -- '.'), it will be successively split if possible for deeper indices into
```
Athanasius, line 102: docstrings & comments should stay to 80 columns at most

killswitch.py
```diff
@@ -1,54 +1,231 @@
...
         :param id: The feature ID to check
         :param version: The version to check killswitches for, defaults to the current EDMC version
         :return: a namedtuple indicating status and reason, if any
         """
+
+        if isinstance(version, str):
+            version = semantic_version.Version.coerce(version)
+
         for ks in self.kill_switches:
             if version not in ks.version:
```
Athanasius, line 223: I had to look up the semantic_version docs *and* run a test to be sure, so add a comment here that, yes, this 'in' syntax does support things like '<1.2.3' for range comparisons.

Athanasius: NB: This could be in Killswitches.md#versions rather than in the code.  All it does is say it's a SimpleSpec, and examples are simple specific versions, not comparison versions.  We need to document that range comparison versions are supported.

killswitch.py
```diff
@@ -76,17 +286,28 @@ def __repr__(self) -> str:
...
+
 
+class KillSwitchJSONFile(TypedDict):  # noqa: D101
+    version: int
+    last_updated: str
+    kill_switches: List[KillSwitchSetJSON]
 
-def fetch_kill_switches(target=DEFAULT_KILLSWITCH_URL) -> Optional[KILL_SWITCH_JSON_DICT]:
+
+def fetch_kill_switches(target=DEFAULT_KILLSWITCH_URL) -> Optional[KillSwitchJSONFile]:
```
Athanasius, line 310: I'm antsy about this remote-data request not being in a thread, but on reflection:

1. If we have core-code killswitches then we **really** *do* want to load them before running any code they might affect.
2. For now I doubt any plugin is using them, and at very worst we'll take a 10 second hit for core code plus each plugin attempting to load killswitches.

Resolving this to ensure core code always runs with it kill switches is mostly moot because what are we going to do if they're not loaded yet?  Hang around for them to be loaded!

For plugins we could document how plugins can load them in a thread (provide a helper?) and also how to ensure they don't process data until the load/check is performed.

Athanasius:
We could use a script to test a killswitch set for what it will actually do.  At the very least ensure it parses OK without running the full program (so can be tested without GUI-ness available).

Athanasius:
Killswitch.py:380:
```python
    logger.info(f'Kill switches last updated {last_updated}')
```
is fine if it was only core killswitches, but this will be ambiguous for extra/plugin killswitches having been loaded.

Athanasius:
Killswitch.py:425
```
def setup_main_list():
    """
    Set up the global set of kill switches for querying.

    Plugins should NOT call this EVER.
```
I don't suppose we want to play silly buggers with checking the call stack to see if a plugin has done that ....  Probably not worth it.

Athanasius (changes requested):
OK, first full review done.  As you can tell from the out-of-review comments this might be the first time I took a serious look at the code at all :P

Athanasius:
> We could use a script to test a killswitch set for what it will actually do. At the very least ensure it parses OK without running the full program (so can be tested without GUI-ness available).

I guess maybe tests/killswitch.py/test_apply could be altered to allow for an input file ?

Athanasius:
> > We could use a script to test a killswitch set for what it will actually do. At the very least ensure it parses OK without running the full program (so can be tested without GUI-ness available).
> 
> I guess maybe tests/killswitch.py/test_apply could be altered to allow for an input file ?

Failing that *some* script, because then we could potentially do something with a push action that only works on the `releases` branch to check the current killswitches file is actually parsable.  That and I could probably lever it into the pre-commit config as well.

A-UNDERSCORE-D:
> Killswitch.py:425
> 
> ```
> def setup_main_list():
>     """
>     Set up the global set of kill switches for querying.
> 
>     Plugins should NOT call this EVER.
> ```
> 
> I don't suppose we want to play silly buggers with checking the call stack to see if a plugin has done that .... Probably not worth it.

Not at all.

>We could use a script to test a killswitch set for what it will actually do. At the very least ensure it parses OK without running the full program (so can be tested without GUI-ness available).

What exactly would you like, just verifying that everything parses correctly?

Athanasius:
> > We could use a script to test a killswitch set for what it will actually do. At the very least ensure it parses OK without running the full program (so can be tested without GUI-ness available).
> 
> What exactly would you like, just verifying that everything parses correctly?

1. The bare minimum is at least that, for the purposes of verifying the file before making it live (rather than a type/thinko meaning it's invalid and we in effect lose all the prior killswitches as a result).
2. Beyond that something to ensure that, e.g. an id is at least citing a plugin that *exists* to catch typos.
3. Beyond that something that literally tells you "disables plugins.edsm.worker.scan", "forces plugins.edsm.journal.event.horizons to True" etc.

A-UNDERSCORE-D:
While working on the requested script I realised that two of our four plugins havent been updated to support the new stuff, thats additionally blocking this

killswitch.py
```diff
@@ -1,54 +1,231 @@
...
+        Apply the rules this SingleKill instance has to make some data okay to send.
+
+        Note that this MODIFIES DATA IN PLACE.
+
+        :param target: data to apply a rule to
+        """
+        # fields that contain . might be going deeper into the dict. check here FIRST to see if they exist at the top
+        # level, if not, then work our way down
+
+        for key in (self.redact_fields if self.redact_fields is not None else []):
```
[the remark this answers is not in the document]
A-UNDERSCORE-D: Order is now explicitly defined, and tested

killswitch.py
A-UNDERSCORE-D: Clarified in docs

killswitch.py
A-UNDERSCORE-D: Added a oneliner helper and doccumented it. As for core code, yeah, I want us to sit and wait, plugins can just defer anything they send until its there (or do it on-thread and block)

A-UNDERSCORE-D:
Killswitch test has been added, plugins updated.

docs/Releasing.md
```diff
@@ -240,7 +240,7 @@ the .exe and .msi because ChangeLog.md is bundled with the install.
 If anything in this new release addresses a bug that causes, e.g. bad data
 to be sent over EDDN, then you should add an appropriate entry to the
 killswitches.json file *in the `releases` branch*.  That file **must only ever
-be commited to the `releases` branch!!!**  See [docs/Killswitches.md](Killswitches.md).
+be committed to the `releases` branch!!!**  See [docs/Killswitches.md](Killswitches.md).
```
Athanasius, line 244: Mention the testing script here, although we probably want it adding into the pre-commit config as well.

killswitch.py
```diff
@@ -67,6 +246,58 @@ def kills_for_version(self, version: semantic_version.Version = _current_version
...
 
+    def check_killswitch(
+        self, name: str, data: T, log=logger, version=_current_version
+    ) -> Tuple[bool, T]:
+        """
+        Check whether or not a killswitch is enabled. If it is, apply rules if any.
+
+        :param name: The killswitch to check
+        :param data: The data to modify if needed
+        :return: A bool indicating if the caller should return, and either the
```
Athanasius, line 257: I would word this as "A two value tuple consisting of ...".

plugins/edsm.py
```diff
@@ -642,35 +647,57 @@ def worker() -> None:  # noqa: CCR001 C901 # Cant be broken up currently
...
-                    if 'edsm-locations' in trace_on and \
-                            any(p for p in pending if p['event'] in ('CarrierJump', 'FSDJump', 'Location', 'Docked')):
-                        logger.trace("pending has at least one of "
-                                     "('CarrierJump', 'FSDJump', 'Location', 'Docked')"
-                                     " and it passed should_send()")
+                    if (
+                        'edsm-locations' in trace_on
+                        and any(p for p in pending if p['event'] in ('CarrierJump', 'FSDJump', 'Location', 'Docked'))
+                    ):
+                        logger.trace(
```
Athanasius, line 688: This can use trace_if(...) now.

plugins/edsm.py
```diff
@@ -642,35 +647,57 @@ def worker() -> None:  # noqa: CCR001 C901 # Cant be broken up currently
...
+                            "pending has at least one of ('CarrierJump', 'FSDJump', 'Location', 'Docked')"
+                            " and it passed should_send()"
+                        )
+
                         for p in pending:
                             if p['event'] in ('Location'):
-                                logger.trace('"Location" event in pending passed should_send(), '
-                                             f'timestamp: {p["timestamp"]}')
+                                logger.trace(
```
Athanasius, line 695: trace_if

docs/Killswitches.md
```diff
@@ -76,8 +76,15 @@ Plugins may use the killswitch system simply by hosting their own version of the
...
-The version of the JSON file will be automatically upgraded if possible by the code KillSwitch code. No behaviour changes will occur--Any killswitches defined in older
- versions will simply become unconditional kills in the new version.
+A helper method `killswitch.get_kill_switch_thread` is provided to allow for simple nonblocking requests for
+KillSwitches. It starts a new thread, performs the HTTP request, and sends the results to the given callback.
+
+**Note that your callback is invoked off-thread. Take precaution for locking if required, and do _NOT_ use tkinter**
+**methods**
+
+The version of the JSON file will be automatically upgraded if possible by the code KillSwitch code.
+No behaviour changes will occur--Any killswitches defined in older versions will simply become unconditional kills in
```
Athanasius, line 86: This `occur--Any` just looks weird.  I'd use:

> No behaviour changes will occur, any killswitches...

If not:

> No behaviour changes will occur.  Any killswitches...

or were you going for:

> No behaviour changes will occur -- any killswitches...

?

killswitch.py
```diff
@@ -277,6 +277,25 @@ def check_killswitch(
         log.info('Rules applied successfully, allowing execution to continue')
         return False, new_data
 
+    def check_multiple_killswitches(self, data: UPDATABLE_DATA, *names: str, log=logger, version=_current_version):
```
Athanasius: declare return type ?

docs/Killswitches.md
```diff
@@ -94,13 +94,16 @@ The current recognised (to EDMC and its internal plugins) killswitch strings are
...
 | `plugins.<plugin>.journal`                   | eddn, inara, edsm, eddb | Disables all journal processing for the plugin                                            |
-| `plugins.<plugin>.worker`                    |       edsm, inara       | Disables the plugins worker thread (effectively disables updates) (does not close thread) |
+| `plugins.<plugin>.worker`                    |      edsm, *inara       | Disables the plugins worker thread (effectively disables updates) (does not close thread) |
 | `plugins.<plugin>.worker.<eventname>`        |       edsm, inara       | Disables the plugin worker for the given eventname                                        |
 | `plugins.<plugin>.journal.event.<eventname>` |    eddn, inara, edsm    | Specific events to disable processing for.                                                |
 
 Killswitches marked with `*` do **not** support modification of their values via set/redact/delete. And as such any match
 will simply stop processing.
 
+For `plugin.inara.worker`, events are checked individually later by the eventname version. Use that to modify individual
```
Athanasius, line 104: Line over 80 chars long.  Actually now I check in VIM to be sure, the whole file could do with a pass to bring all non-table/link stuff within that limit.

Athanasius (changes requested):
Just a few niggles in docs and formatting therein.

docs/Killswitches.md
A-UNDERSCORE-D: Sorry bad habit, I use -- as an em-dash which shouldnt have spaces, but a period works fine there as well

killswitch.py
A-UNDERSCORE-D: Done where possible

plugins/edsm.py
A-UNDERSCORE-D: Resolved with rebase

plugins/edsm.py
A-UNDERSCORE-D: Resolved with rebase

killswitch.py
A-UNDERSCORE-D: I *think* I hadnt pushed changes or something for this. They should be there now.

docs/Killswitches.md
A-UNDERSCORE-D: Resolved in all the files I touched docs wise

killswitch.py
Athanasius: 1. Either this needs to allow for setting a non-existent index that isn't an append.

2. OR Killswitches.md needs to document that an index on a set must either be an existing one, or only a simple append.

3. OR this function needs to be marked as explicitly possibly raising `IndexError`.

2 or 3 needs a test put in to ensure that happens as documented.

I added:

```python
        (['test append'], '2', '', 'yay', ['test append', 'yay']),
```
to the 'test_apply()' parameters in `test_apply.py` and it blew up with an IndexError.

tests/killswitch.py/test_apply.py
```diff
@@ -0,0 +1,92 @@
...
+        ({'now': 'with', 'a': 'dict'}, 'now', '', None, {'now': None, 'a': 'dict'}),
+        ({'depth': {'is': 'important'}}, 'depth.is', '', 'nonexistent', {'depth': {'is': 'nonexistent'}}),
+        ([{'test': ['stuff']}], '0.test.0', '', 'things', [{'test': ['things']}]),
+        (({'test': {'with': ['a', 'tuple']}},), '0.test.with.0', 'delete', '', ({'test': {'with': ['tuple']}},)),
+        ({'test': ['with a', {'set', 'of', 'stuff'}]}, 'test.1', 'delete', '', {'test': ['with a']}),
+        ({'keys.can.have.': 'dots!'}, 'keys.can.have.', '', '.s!', {'keys.can.have.': '.s!'}),
+        ({'multilevel.keys': {'with.dots': False}}, 'multilevel.keys.with.dots',
+         '', True, {'multilevel.keys': {'with.dots': True}}),
+        ({'dotted.key.one.level': False}, 'dotted.key.one.level', '', True, {'dotted.key.one.level': True}),
+        ({'dotted.key.one.level': {"oops.i.lied": False}}, 'dotted.key.one.level.oops.i.lied',
```
Athanasius, line 84: This is essentially the same test as `multilevel.keys` one two lines above.

A-UNDERSCORE-D: I was apparently half asleep while writing those. I dont think theres any harm in leaving it?

docs/Killswitches.md
```diff
@@ -1,62 +1,177 @@
...
 }
 ```
 
+- `plugins.edsm.send` will have fields deleted, set, and redacted, and then
+  will *not* be halted, the send will continue with the modified data.
+- `plugins.some_plugin.some_thing` will never be allowed to continue
+  (as all fields are blank)
+
+Indexing of lists (and any other `Sequence`) is done with numbers as normal.
+Negative numbers do work to reference the end of sequences.
```
Athanasius, line 74: As I went searching yesterday, add a note that:

```
To append you will need to use `len(array/set/whatever)`, there's no way to do it just with indexing.
```

I wish `-0` was special in that way, but obviously all of python's parsing would need the concept of 'negative zero' for that to work.

Athanasius: Ah never mind, you covered this below in the Sets section.

docs/Killswitches.md
```diff
@@ -1,62 +1,177 @@
...
+#### Sets
+
+Sets always succeed for `MutableMapping` objects, either creating new keys
+or updating existing ones. For `MutableSequence`s, however, a set can fail if
+the index is not within `len(target)`. If the index is exactly `len(target)`,
+`append` is used.
+
+You can set values other than strings, but you are limited to what json itself
+supports, and the translation thereof to python. In general this means:
+`int`, `float`, `string`, `bool`, and `None` (json `null`) values may be set.
```
Athanasius, line 117: list/array and dict ?

killswitch.py
```diff
@@ -1,54 +1,232 @@
...
+
+def _apply(target: UPDATABLE_DATA, key: str, to_set: Any = None, delete: bool = False):
+    """
+    Set or delete the given target key on the given target.
+
+    :param target: The thing to set data on
+    :param key: the key or index to set the data to
+    :param to_set: the data to set, if any, defaults to None
+    :param delete: whether or not to delete the key or index, defaults to False
+    :raises ValueError: when an unexpected target type is passed
```
Athanasius, line 71:
```
NB: Other exceptions may be raised, i.e. IndexError if supplying a bad index.
```

My point here is to not have someone who's not read all the code/docs think that this can *only* raise ValueError.

tests/killswitch.py/test_apply.py
Athanasius: No harm, other than someone checking this wondering *why* there's essentially the same test twice, and going down a rabbithole trying to figure out how it's different.  So remove this second one please.

PR #1206, opened 2021-08-01
Added monitor state TypedDict definitions

A-UNDERSCORE-D:
These don't actually change types or make anything more safe. What they
WILL do is make mypy and co complain loudly if we make mistakes.

Im looking for comments on this, as there are a few issues currently

For one, as it stands we None a lot of things that "will" exist always at runtime, which while _we_ know are safe, the type checker cannot, thus a large number of errors appear.

There are two ways to fix this:
-  do None checks everywhere (which is probably what we SHOULD do but given the shape of monitors interactions with this, will be messy and create a lot more confusing code)
- lie to the type checkers -- By telling them the real type of $thing, setting None anyway and telling them to ignore that we set None

Im not sure which to go with, but either way this will prevent typo issues in future and do some type checking.


Athanasius:
I definitely like the general idea of this.  I think I'll need to delve into the code for some of those Optional[None] to get a feel for how much work it would be to do the extra None checking.

monitor_state_dict.py
```diff
@@ -0,0 +1,110 @@
...
+from __future__ import annotations
+
+from typing import MutableMapping, TYPE_CHECKING, Any, DefaultDict, Dict, List, Literal, Optional, Set, Tuple, TypedDict
+
+
+class MonitorStateDict(TypedDict):
+    """Top level state dictionary for monitor.py."""
+
+    # Game related
+    GameLanguage:           Optional[str]               # From `Fileheader`
```
Athanasius, line 15: For these `str` we could just change the code to use `''` instead of `None` ?  It's as falsey.

monitor_state_dict.py
```diff
@@ -0,0 +1,110 @@
...
+    GameVersion:            Optional[str]               # From `Fileheader`
+    GameBuild:              Optional[str]               # From `Fileheader`
+    Horizons:               Optional[bool]                        # Does the player have Horizons?
```
Athanasius, line 18: I'm not seeing how defaulting this to `False`, so never `None` would cause an issue.  Currently it's literally only ever set from `LoadGame`, which we need in order to be doing anything with it anyway.

monitor_state_dict.py
```diff
@@ -0,0 +1,110 @@
...
+    Odyssey:                bool                        # Have we detected Odyssey?
+
+    # Multi-crew
+
+    Captain:                Optional[str]               # If on a crew, the captian's name
```
Athanasius, line 23: This should also be OK using `''` as the `None` value.

monitor_state_dict.py
```diff
@@ -0,0 +1,110 @@
...
+    Role:                   Optional[Literal['Idle', 'FireCon', 'FighterCon']]  # Role in crew
```
Athanasius, line 24: This should also be ok with `''` for `None`. EDMarketConnector.py:crewroletext() will simply need an extra entry in the dict it uses.

monitor_state_dict.py
```diff
@@ -0,0 +1,110 @@
...
+
+    # Cmdr state
+    FID:                    Optional[str]               # Frontier CMDR ID
```
Athanasius, line 27: We don't even check if this *is* set before use, so again it could use `''`.

monitor_state_dict.py
```diff
@@ -0,0 +1,110 @@
...
+    Friends:                Set[str]                    # Online Friends
+    Credits:                int
+    Loan:                   Optional[int]
```
Athanasius, line 30: This literally defaults to `0` in Journal `LoadGame`, so there's no reason for us to not use that as the default value.

monitor_state_dict.py
```diff
@@ -0,0 +1,110 @@
...
+    Reputation:             Dict[Any, Any]              # TODO
+    Statistics:             Dict[Any, Any]              # This is very freeform.
+
+    # Engineering Materials
+    Raw:                    DefaultDict[str, int]
+    Encoded:                DefaultDict[str, int]
+    Manufactured:           DefaultDict[str, int]
+
+    # Ship
+    ShipID:                 Optional[str]
```
Athanasius, line 42: This gets compared to the CAPI data ship->id so we'd need to be sure the CAPI never brainfarts and returns that.

Might be a wrinkle with EDO starting on foot in a new account ?

monitor_state_dict.py
```diff
@@ -0,0 +1,110 @@
...
+    ShipIdent:              Optional[str]
```
Athanasius, line 43: Should be OK as `''`.

monitor_state_dict.py
```diff
@@ -0,0 +1,110 @@
...
+    ShipName:               Optional[str]
```
Athanasius, line 44: ShipName could be problematic.  Check the multi-line comment at monitor.py:596 (`event_type == 'loadout'` code).

Although given `''` is falsey that code doesn't actually need to check for that after  the `if entry['ShipName']`, so perhaps it will be OK.

monitor_state_dict.py
```diff
@@ -0,0 +1,110 @@
...
+    ShipType:               Optional[str]
```
Athanasius, line 45: Comparison to CAPI `ship->name` so the same caution as for ShipID.

monitor_state_dict.py
```diff
@@ -0,0 +1,110 @@
...
+
+    HullValue:              Optional[int]
```
Athanasius, line 47: For all three of `HullValue`, `ModulesValue` and `Rebuy` we should be OK switching to a default of `0`.  It's falsey, so `if <var>` checks won't change.  We *will* need to specify a non-default default on `entry.get(...)` calls to use `0` instead of `None`.

Athanasius:
The one thing about all my comments re: changing from `None` to `''`, `0`, or other type-appropriate falsey value is that whilst we can be sure core code will work with it, we don't know what plugins are doing with the state values.

All it takes is one plugin dev doing `if state['ShipName'] is not None` or similar and we're obligated to bump our major version number if we make those changes.

I'm inclined to just do that.

monitor_state_dict.py
```diff
@@ -0,0 +1,110 @@
...
+    ModulesValue:           Optional[int]
+    Rebuy:                  Optional[int]
+    Modules:                Optional[Dict[Any, Any]]    # TODO
```
Athanasius, line 50: Use `{}` as default ?

monitor_state_dict.py
```diff
@@ -0,0 +1,110 @@
...
+
+    # Cargo (yes technically its on the cmdr not the ship but this makes more sense.)
+    CargoJSON:              Optional[MutableMapping[str, Any]]    # Raw data from the last cargo.json read
+    Cargo:                  DefaultDict[str, int]
+
+    # Navigation
+    Route:                  Optional[NavRoute]         # Last route plotted
```
Athanasius, line 57: `{}` shoudl work again.

But also, I think we have a bug.  Current `develop` has `'Route'` indeed, but also then has `self.state['NavRoute'] = entry`.   Our docs say `NavRoute`, so this needs changing to `NavRoute` anyway.

monitor_state_dict.py
```diff
@@ -0,0 +1,110 @@
...
+    Body:                   Optional[str]
```
Athanasius, line 58: `Body` and `BodyType` should be OK with `''` default, just requires some `entry.get(...)` adjustments and a conditional or two adjusting.

monitor_state_dict.py
```diff
@@ -0,0 +1,110 @@
...
+    BodyType:               Optional[str]
```
Athanasius, line 59: `None` is equivalent to `False` for `Taxi` and `Dropship`.

monitor_state_dict.py
```diff
@@ -0,0 +1,110 @@
...
+    Dropship:               Optional[bool]
+
+    # Odyssey
+    OnFoot:                 bool
+    Component:              DefaultDict[str, int]
+    Item:                   DefaultDict[str, int]
+    Consumable:             DefaultDict[str, int]
+    Data:                   DefaultDict[str, int]
+    BackPack:               OdysseyBackpack
+    BackpackJSON:           Optional[MutableMapping[str, Any]]              # Direct from Game
```
Athanasius, line 70: No reason to not use `{}` as default for `BackpackJSON` and `ShipLockerJSON`.

monitor_state_dict.py
A-UNDERSCORE-D: I dont quite want to use False there, None has specific meaning for "dont know" whereas True is yes and False is no

monitor_state_dict.py
```diff
@@ -0,0 +1,110 @@
...
+    ShipLockerJSON:         Optional[MutableMapping[str, Any]]              # Direct from Game
+
+    SuitCurrent:            Optional[int]               # TODO: int?
```
Athanasius, line 73: Not an int, it's a dict, set by indexing the Suits by int id.  And thus probably OK to default to `{}` instead.

monitor_state_dict.py
```diff
@@ -0,0 +1,110 @@
...
+    Suits:                  Dict[Any, Any]              # TODO: With additional class
+    SuitLoadoutCurrent:     Optional[int]               # TODO: int?
```
Athanasius, line 75: Similar to SuitCurrent, this isn't an int, it's a dict.

monitor_state_dict.py
A-UNDERSCORE-D: Direct checks for `ShipName` appear to always do a falsey check so it should be fine -- obviously this entire PR will need testing

A-UNDERSCORE-D:
Something that I want to add, a heavy mode that checks types on monitor.state after each journal line, to ensure that things are okay

monitor_state_dict.py
A-UNDERSCORE-D: ShipType looks to always be assumed to be a string as well

companion.py
```diff
@@ -713,28 +714,37 @@ def suit_update(self, data: CAPIData) -> None:
...
+            current_suit['edmcName'] = monitor.suit_sane_name(loc_name)
+
         for s in monitor.state['Suits']:
             loc_name = monitor.state['Suits'][s].get('locName', monitor.state['Suits'][s]['name'])
             monitor.state['Suits'][s]['edmcName'] = monitor.suit_sane_name(loc_name)
 
+        suit_loadouts: Optional[Union[List[SuitLoadoutDict], Dict[int, SuitLoadoutDict]]]
         if (suit_loadouts := data.get('loadouts')) is None:
             logger.warning('CAPI data had "suit" but no (suit) "loadouts"')
+            return
```
Athanasius, line 747: I assume this is the change of behaviour you mentioned on Discord.

Before this if CAPI says no suit loadouts we'll None our copy.  With this, it won't.

So now the question is do we really distrust CAPI that much ?  I'm inclined to allow this change as the Journal is in much better shape and we *should* always be picking things up properly from it now.

monitor_state_dict.py
```diff
@@ -0,0 +1,190 @@
...
+    OnFoot:                 bool
+    Component:              DefaultDict[str, int]
+    Item:                   DefaultDict[str, int]
+    Consumable:             DefaultDict[str, int]
+    Data:                   DefaultDict[str, int]
+    BackPack:               OdysseyBackpack
+    BackpackJSON:           MutableMapping[str, Any]        # Direct from Game
+    ShipLockerJSON:         MutableMapping[str, Any]        # Direct from Game
+
+    SuitCurrent:            Optional[SuitDict]
```
Athanasius, line 77: Without checking the code that uses this, I am wondering if we can't default this to `{}` and lose the `Optional`.  Same for SuitLoadoutCurrent below.

monitor.py
```diff
@@ -383,7 +391,8 @@ def worker(self) -> None:  # noqa: C901, CCR001
 
         # Watchdog thread -- there is a way to get this by using self.observer.emitters and checking for an attribute:
         # watch, but that may have unforseen differences in behaviour.
-        emitter = self.observed and self.observer._emitter_for_watch[self.observed]  # Note: Uses undocumented attribute
+        # Note: Uses undocumented attribute
```
Athanasius, line 394: Should probably be a HACK comment ?

monitor.py
```diff
@@ -539,23 +548,24 @@ def parse_entry(self, line: bytes) -> MutableMapping[str, Any]:  # noqa: C901, C
...
-                    'Statistics': {},
-                    'Role':       None,
-                    'Taxi':       None,
-                    'Dropship':   None,
-                    'Body':       None,
-                    'BodyType':   None,
-                })
+
+                # Cant use update() without the entire thing, do stuff manually here
+                self.state['Captain'] = None
```
Athanasius, line 553: Should be `''` now surely.

monitor.py
```diff
@@ -539,23 +548,24 @@ def parse_entry(self, line: bytes) -> MutableMapping[str, Any]:  # noqa: C901, C
...
+                self.state['Credits'] = entry['Credits']
+                self.state['FID'] = entry.get('FID', '')   # From 3.3
+                self.state['Horizons'] = entry['Horizons']  # From 3.0
+                self.state['Odyssey'] = entry.get('Odyssey', False)  # From 4.0 Odyssey
+                self.state['Loan'] = entry['Loan']
+                self.state['Engineers'] = {}
+                self.state['Rank'] = {}
+                self.state['Reputation'] = {}
+                self.state['Statistics'] = {}
+                self.state['Role'] = None
```
Athanasius, line 563: Should be `''`

monitor.py
```diff
@@ -1337,24 +1356,26 @@ def parse_entry(self, line: bytes) -> MutableMapping[str, Any]:  # noqa: C901, C
 
             elif event_type == 'materials':
                 for category in ('Raw', 'Manufactured', 'Encoded'):
+                    category = cast(Literal['Raw', 'Manufactured', 'Encoded'], category)
```
Athanasius, line 1359: Any reason this one isn't TYPE_CHECKING gated, but one above is ?

monitor.py
```diff
@@ -1673,8 +1700,12 @@ def suitloadout_store_from_event(self, entry) -> Tuple[int, int]:
             suitname = entry.get('SuitName_Localised', entry['SuitName'])
             edmc_suitname = self.suit_sane_name(suitname)
             suit = {
-                'edmcName': edmc_suitname,
-                'locName':  suitname,
+                'edmcName':     edmc_suitname,
+                'locName':      suitname,
+                'suitId':       0-1,
```
Athanasius, line 1705: # WORKAROUND

monitor_state_dict.py
A-UNDERSCORE-D: specifically because the "default" cannot be `{}`, it must be a dict that matches `SuitDict`, so the minimum default is:
```python
{
'name': '',
'locName': '',
'weaponrackId': 0,
'id': None,
'suitId': 0,
'mods': [],
}
```
which is not going to be falsy

companion.py
A-UNDERSCORE-D: yeah Im undecided here. The other option is to blank our loadouts and return?

monitor.py
A-UNDERSCORE-D: Whoops, spotted a few of those. Fixed now

monitor.py
A-UNDERSCORE-D: Because who needs consistency! fixed by removing the guard on the other one (as there were more without than with)

A-UNDERSCORE-D:
> Something that I want to add, a heavy mode that checks types on monitor.state after each journal line, to ensure that things are okay

This isnt going to happen, the introspection required isnt there, half of the types will be strings

A-UNDERSCORE-D:
This probably is going to wholly need a rewrite at this point.

Rixxan:
@A-UNDERSCORE-D you mentioned above this will need a rewrite, should I close this PR at this point then?

A-UNDERSCORE-D:
Likely yeah. It was intended to bring safety to the main loop, but we were changing a lot at the time so we left it. It may be salvageable in some places

Athanasius:
It was kept around as easy reference for maybe completing the work at some point.

The issue it was designed to address is that the monitor state dictionary is just that, a dictionary, and as such it's all too easy to mis-type/capitalise a member of it and thus introduce a bug.  We can't just wholesale change it to an object without huge impact on plugin developers, so this was an attempt at a way to move towards that whilst still supporting dictionary key access.

Issue #1198, opened 2021-07-14
CAPI Lagging at Florianz Deposits in HIP 41526


outsidefactor:
- Version: [5.1.1]
 - Game Version: ['Odyssey']
 - OS: [Linux Manjaro]
 - OS Locale: [English]
 - If applicable: Browser [chrome]
 - Please attach **BOTH** log fil
[EDMarketConnector-debug.log](https://github.com/EDCD/EDMarketConnector/files/6815603/EDMarketConnector-debug.log)
es, by dragging and dropping them into this input:
     1. `%TEMP%\EDMarketConnector.log` from *immediately* after the bug occurs (re-running the application overwrites this file).
     1. `%TEMP%\EDMarketConnector\EDMarketConnector-debug.log`.  See [Debug Log File](https://github.com/EDCD/EDMarketConnector/wiki/Troubleshooting#debug-log-files).  NB: If you don't have this log file then you're not running the latest version of the application and should update first to see if we already fixed the bug you're reporting.

**Describe the bug**
When docking at a particular station EDMC always reports "Error: Frontier server is lagging". Restarting the client and EDMC does not resolve. Happens predictably.

**To Reproduce**
1. Travel to HIP 41526
2. Dock at Florianz Deposits
3. See error

**Expected behavior**
Not to see an error

**Screenshots**

**Additional context**
[Journal.210714172115.01.log](https://github.com/EDCD/EDMarketConnector/files/6815627/Journal.210714172115.01.log)
[HIP 41526.Florianz Deposits ++.2021-07-14T21.10.zip](https://github.com/EDCD/EDMarketConnector/files/6815650/HIP.41526.Florianz.Deposits.%2B%2B.2021-07-14T21.10.zip)
[HIP 41526.Florianz Deposits ++.2021-07-14T21.13.zip](https://github.com/EDCD/EDMarketConnector/files/6815652/HIP.41526.Florianz.Deposits.%2B%2B.2021-07-14T21.13.zip)

Athanasius:
This is because the CAPI data has the "settlement size/security" `+`s on the name:

```
    "name": "Florianz Deposits ++",
```

Whereas, of course, the Journal correctly has only the name:

```
"StationName":"Florianz Deposits",
```
(Docked event)

I'll have a think about how much I want to implement a workaround for this.

[...]

PR #1220, opened 2021-08-06
re #1198 work around +size suffixes in CAPI responses

inorton:
Strip the extra + chars in surface port names that are wrongly returned by the CAPI.

Also rely on station ID checks rather than name comparison before we assume the CAPI is lagging

### DIFF
--- a/companion.py
+++ b/companion.py
@@ -670,6 +670,9 @@ class Session(object):
             logger.warning("No lastStarport name!")
             return data
 
+        # WORKAROUND: n/a | 06-08-2021: Issue 1198 and https://issues.frontierstore.net/issue-detail/40706 -- strip "+" chars off star port names returned by the CAPI
+        last_starport_name = last_starport["name"] = last_starport_name.rstrip(" +")
+
         services = last_starport.get('services', {})
         if not isinstance(services, dict):
             # Odyssey Alpha Phase 3 4.0.0.20 has been observed having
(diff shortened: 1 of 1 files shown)


companion.py
```diff
@@ -670,6 +670,11 @@ def station(self) -> CAPIData:  # noqa: CCR001
             logger.warning("No lastStarport name!")
             return data
 
+        # work around issue 1198 -
+        # https://issues.frontierstore.net/issue-detail/40706
+        # strip "+" chars off star port names returned by the CAPI
```
A-UNDERSCORE-D, line 675:
```suggestion
        # WORKAROUND: n/a | 06-08-2021: Issue 1198 and https://issues.frontierstore.net/issue-detail/40706 -- strip "+" chars off star port names returned by the CAPI
```

companion.py
```diff
@@ -687,22 +692,22 @@ def station(self) -> CAPIData:  # noqa: CCR001
...
-            if last_starport_name != shipdata['name'] or last_starport_id != int(shipdata['id']):
-                logger.warning(f"{last_starport_name!r} != {shipdata['name']!r} or "
-                               f"{last_starport_id!r} != {int(shipdata['id'])!r}")
+            if last_starport_id != int(shipdata['id']):
+                logger.warning(f"{last_starport_id!r} != {int(shipdata['id'])!r}")
                 raise ServerLagging()
 
             else:
+                shipdata['name'] = last_starport_name
                 data['lastStarport'].update(shipdata)
```
A-UNDERSCORE-D, line 712:
```suggestion
# WORKAROUND END

```

A-UNDERSCORE-D (changes requested):

A-UNDERSCORE-D:
Please switch this to point at `develop`, other than the requested comment changes to fit with [the new standard](https://github.com/EDCD/EDMarketConnector/blob/main/Contributing.md#comments) this LGTM

A-UNDERSCORE-D (approved):

A-UNDERSCORE-D:
Looks good, should work fine but Ill let @Athanasius give it a look over in case I missed any gotchas

Athanasius:
> Looks good, should work fine but Ill let @Athanasius give it a look over in case I missed any gotchas

That all looks good, I'll give it a test this afternoon and then merge if I don't run into any issues.

PR #1229, opened 2021-08-09
Include a copy of any relevant licenses

### LINKED ISSUE #1057: Ensure we are reproducing any necessary licenses
When I was looking into `infi.systray` licensing with a view to either straight up including it in our source, possibly in an edited form, or picking out the necessary win32 API calls to implement it ourselves ... I found it is BSD 3-clause.  That means we'd need (even if only packaging with py2exe) to include a copy of its license.

We should review *all* the third-party software that we package for Windows builds and ensure we are carrying a copy of their license and/or a reference to it upstream.

I propose `docs/licenses/<package>/<file>` as the standard for this.

Athanasius:
This includes licenses for:

1. python.org distributions.
2. All the tools we use to build for Windows: py2exe, WiX, Windows SDK Kit.
3. WinSparkle (Windows update checking).
4. Any python modules we can tell end up included in either the executables or the library.zip that py2exe produces.

Closes #1057


Athanasius:
Notably we're making very minimal effort to have users *see* the licenses.  I've added a note at the top of ChangeLog.md for a minimal effort version of this.

Anything else would be dependent on https://github.com/EDCD/EDMarketConnector/projects/10 (redo installer).

A-UNDERSCORE-D (approved):

Issue #1226, opened 2021-08-08
Define & Document a function name prefix for core us in plugins


Athanasius:
i.e. we have `journal_entry()` now and a proposed `journal_entry_cqc()` (see #1030 ).  But what if a plugin developer already has that function name in their code, with matching number of arguments?

We need to define a prefix for function names that plugin developers know not to use, and then always use it ourselves for any new API function names.

inorton:
For a long time now I've thought that the plugin API entrypoint/discovery should be different. I'd like to see it become class based where plugins inherit from a base class we provide and override methods.

Athanasius:
> For a long time now I've thought that the plugin API entrypoint/discovery should be different. I'd like to see it become class based where plugins inherit from a base class we provide and override methods.

Yes, see @A-UNDERSCORE-D 's work on a wholly new plugin API/paradigm (but with backwards compatibility).

*This* issue is more about avoiding an issue in the time before we get that into a mergable state, and would still apply during the backwards compatibility stage.

PR #1232, opened 2021-08-09
Feature/1030/journal entry cqc

### LINKED ISSUE #1030: Review CQC detection.
Being able to queue for CQC whilst playing the actual game, i.e. not from Main Menu beforehand, is causing some issues on EDSM.

CQC re-uses 'real' star systems. Ref: https://forums.frontier.co.uk/threads/cqc-systems.234394/

`Theta Ursae Majoris`
`Boepp SU-E d12-818`
`Eta Cephei`
`Bleae Aewsy GA-Y d1-14`

Check the EDMC code for CQC detection.  We might want to start checking for these specific system names in, I think, `Location` events and then not send to EDSM (or Inara? it might have good enough code to cope with this already).

norohind:
Closes #1030 
It uses `LoadGame` events to detect enter to CQC: it is definetly LoadGame to CQC if in `LoadGame` event missing fields `Ship` and `GameMode`. I've tested the correctness of this method in some different cases:
1. Horizons
    1. Log in to cqc from main menu as first action after game loading
    2. Log in to cqc from main menu after playing in main game
    3. Log in to cqc from ingame cqc queuing
    4. Training missions

2. Odyssey
    1. Log in to cqc from main menu as first action after game loading
    2. Log in to cqc from main menu after playing in main game
    3. Log in to cqc from ingame cqc queuing
    4. Log in to main game when on foot
    
3. Pre horizons:
    1. Log in to cqc from main menu as first action after game loading
    2. Log in to cqc from main menu after playing in main game
    3. Log in to cqc from ingame cqc queuing

4. Arena standalone
    1. Log in to cqc from main menu as first action after game loading
    2. Log in to cqc from main menu after playing in cqc

Usually, when you are in CQC, EDMC doesn't pass any events to plugins `journal_entry` function but now it passes these events to new defined `journal_entry_cqc` function, args for `journal_entry_cqc` slighty different from args for typical `journal_entry`: `cmdr, is_beta, entry, state`. I've thrown out `system` and `station` since these are meaningless in CQC. It will allow creating plugins that rely on cqc events, for example it can be helpful for [EDMC-Discord-Presence](https://github.com/Elite-Kode/EDMC-Discord-Presence/) plugin.

One thing confuses me - monitor.mode, in original code it set to None on EDMC startup and changing on every `LoadGame` according to `GameMode` field. On the one side, it will cause passing any journal events to `journal_entry_cqc` even if player already left CQC and idling in main menu. On the other side, it already causes passing any journal events to `journal_entry` when player has playing main game and logged out to main menu and nobody complains to it. Give me a pointer about it please.

### DIFF
--- a/EDMarketConnector.py
+++ b/EDMarketConnector.py
@@ -872,7 +872,8 @@ class AppWindow(object):
         play_bad = False
         err: Optional[str] = None
 
-        if not monitor.cmdr or not monitor.mode or monitor.state['Captain'] or not monitor.system:
+        if not monitor.cmdr or not monitor.mode or monitor.state['Captain'] or not monitor.system or \
+                monitor.mode == 'CQC':
             return  # In CQC or on crew - do nothing
 
         if companion.session.state == companion.Session.STATE_AUTH:
@@ -1171,6 +1172,15 @@ class AppWindow(object):
                     config.set('cmdrs', config.get_list('cmdrs', default=[]) + [monitor.cmdr])
                 self.login()
 
+            if monitor.mode == 'CQC':
+                err = plug.notify_journal_entry_cqc(monitor.cmdr, monitor.is_beta, entry, monitor.state)
+                if err:
+                    self.status['text'] = err
+                    if not config.get_int('hotkey_mute'):
+                        hotkeymgr.play_bad()
+
+                return  # in CQC
+
             if not entry['event'] or not monitor.mode:
                 # logger.trace('Startup or in CQC, returning')
                 return  # Startup or in CQC
@@ -1335,6 +1345,7 @@ class AppWindow(object):
             self.button['text'] = self.theme_button['text'] = _('Update')  # LANG: Update button in main window
             self.button['state'] = self.theme_button['state'] = (monitor.cmdr and
                                                                  monitor.mode and
+                                                                 # and monitor.mode == 'CQC' and
                                                                  not monitor.state['Captain'] and
                                                                  monitor.system and
                                                                  tk.NORMAL or tk.DISABLED)
--- a/EDMarketConnector.py
+++ b/EDMarketConnector.py
@@ -1345,7 +1345,7 @@ class AppWindow(object):
             self.button['text'] = self.theme_button['text'] = _('Update')  # LANG: Update button in main window
             self.button['state'] = self.theme_button['state'] = (monitor.cmdr and
                                                                  monitor.mode and
-                                                                 # and monitor.mode == 'CQC' and
+                                                                 monitor.mode == 'CQC' and
                                                                  not monitor.state['Captain'] and
                                                                  monitor.system and
                                                                  tk.NORMAL or tk.DISABLED)
--- a/EDMarketConnector.py
+++ b/EDMarketConnector.py
@@ -872,8 +872,10 @@ class AppWindow(object):
         play_bad = False
         err: Optional[str] = None
 
-        if not monitor.cmdr or not monitor.mode or monitor.state['Captain'] or not monitor.system or \
-                monitor.mode == 'CQC':
+        if (
+                not monitor.cmdr or not monitor.mode or monitor.state['Captain']
+                or not monitor.system or monitor.mode == 'CQC'
+        ):
             return  # In CQC or on crew - do nothing
 
         if companion.session.state == companion.Session.STATE_AUTH:
--- a/EDMarketConnector.py
+++ b/EDMarketConnector.py
@@ -1172,7 +1172,7 @@ class AppWindow(object):
                     config.set('cmdrs', config.get_list('cmdrs', default=[]) + [monitor.cmdr])
                 self.login()
 
-            if monitor.mode == 'CQC':
+            if monitor.mode == 'CQC' and entry['event']:
                 err = plug.notify_journal_entry_cqc(monitor.cmdr, monitor.is_beta, entry, monitor.state)
                 if err:
                     self.status['text'] = err
--- a/EDMarketConnector.py
+++ b/EDMarketConnector.py
@@ -1184,8 +1184,8 @@ class AppWindow(object):
                 return  # in CQC
 
             if not entry['event'] or not monitor.mode:
-                # logger.trace('Startup or in CQC, returning')
-                return  # Startup or in CQC
+                # logger.trace('Startup, returning')
+                return  # Startup
 
             if entry['event'] in ['StartUp', 'LoadGame'] and monitor.started:
                 logger.info('Startup or LoadGame event')
--- a/monitor.py
+++ b/monitor.py
@@ -531,7 +531,7 @@ class EDLogs(FileSystemEventHandler):  # type: ignore # See below
                 self.cmdr = entry['Commander']
                 # 'Open', 'Solo', 'Group', or None for CQC (and Training - but no LoadGame event)
                 if not entry.get('Ship') and not entry.get('GameMode') or entry.get('GameMode').lower() == 'CQC':
-                    logger.debug(f'loadgame to cqc: {entry}')
+                    logger.trace(f'loadgame to cqc: {entry}')
                     self.mode = 'CQC'
                 else:
                     self.mode = entry.get('GameMode')
--- a/monitor.py
+++ b/monitor.py
@@ -530,7 +530,9 @@ class EDLogs(FileSystemEventHandler):  # type: ignore # See below
                 self.cmdr = entry['Commander']
                 # 'Open', 'Solo', 'Group', or None for CQC (and Training - but no LoadGame event)
                 if not entry.get('Ship') and not entry.get('GameMode') or entry.get('GameMode').lower() == 'CQC':
-                    logger.trace(f'loadgame to cqc: {entry}')
+                    if 'cqc-loadgame-events' in trace_on:
+                        logger.trace(f'loadgame to cqc: {entry}')
+
                     self.mode = 'CQC'
 
                 else:
--- a/monitor.py
+++ b/monitor.py
@@ -529,7 +529,7 @@ class EDLogs(FileSystemEventHandler):  # type: ignore # See below
                 # Odyssey: bool
                 self.cmdr = entry['Commander']
                 # 'Open', 'Solo', 'Group', or None for CQC (and Training - but no LoadGame event)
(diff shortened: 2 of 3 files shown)


monitor.py
```diff
@@ -529,7 +530,11 @@ def parse_entry(self, line: bytes) -> MutableMapping[str, Any]:  # noqa: C901, C
                 # Odyssey: bool
                 self.cmdr = entry['Commander']
                 # 'Open', 'Solo', 'Group', or None for CQC (and Training - but no LoadGame event)
-                self.mode = entry.get('GameMode')
+                if not entry.get('Ship') and not entry.get('GameMode') or entry.get('GameMode').lower() == 'CQC':
+                    logger.debug(f'loadgame to cqc: {entry}')
```
A-UNDERSCORE-D, line 534: use trace here

plug.py
```diff
@@ -312,6 +312,29 @@ def notify_newdata(data, is_beta):
...
+    :param cmdr: The Cmdr name, or None if not yet known
+    :param entry: The journal entry as a dictionary
+    :param state: A dictionary containing info about the Cmdr, current ship and cargo
+    :param is_beta: whether the player is in a Beta universe.
+    :returns: Error message from the first plugin that returns one (if any)
+    """
+
+    error = None
+    for plugin in PLUGINS:
+        journal_entry = plugin._get_func('journal_entry_cqc')
```
A-UNDERSCORE-D, line 327: probably rename this cqc_callback? or something a bit less opaque than `journal_entry`

plug.py
```diff
@@ -312,6 +312,29 @@ def notify_newdata(data, is_beta):
...
+        if journal_entry:
```
A-UNDERSCORE-D, line 328:
```suggestion
        if journal_entry is not None and callable(journal_entry):
```
I think be paranoid here about what this is, just in case someone has used this name

monitor.py
```diff
@@ -529,7 +530,11 @@ def parse_entry(self, line: bytes) -> MutableMapping[str, Any]:  # noqa: C901, C
...
+                    self.mode = 'CQC'
+                else:
```
A-UNDERSCORE-D, line 536:
```suggestion

                else:
```

plug.py
```diff
@@ -312,6 +312,29 @@ def notify_newdata(data, is_beta):
...
+            try:
+                # Pass a copy of the journal entry in case the callee modifies it
+                newerror = journal_entry(cmdr, is_beta, dict(entry), dict(state))
+                error = error or newerror
+            except Exception as e:
```
A-UNDERSCORE-D, line 333:
```suggestion

            except Exception as e:
```

plug.py
```diff
@@ -312,6 +312,29 @@ def notify_newdata(data, is_beta):
...
+                logger.exception(f'Plugin "{plugin.name}" failed')
+    return error
```
A-UNDERSCORE-D, line 335:
```suggestion

    return error
```

EDMarketConnector.py
```diff
@@ -872,7 +872,8 @@ def getandsend(self, event=None, retrying: bool = False):  # noqa: C901, CCR001
         play_bad = False
         err: Optional[str] = None
 
-        if not monitor.cmdr or not monitor.mode or monitor.state['Captain'] or not monitor.system:
+        if not monitor.cmdr or not monitor.mode or monitor.state['Captain'] or not monitor.system or \
+                monitor.mode == 'CQC':
```
A-UNDERSCORE-D, line 876:
```suggestion
        if (
                not monitor.cmdr or not monitor.mode or monitor.state['Captain'] 
                or not monitor.system or  monitor.mode == 'CQC'
        ):
```

plug.py
norohind: should I also remove ` as e` since it not used and make linter angry despite the fact that in the old code it exists for something?

plug.py
norohind: Probably yes but may be time has come to define custom prefix for such kind of functions (https://github.com/EDCD/EDMarketConnector/issues/1226) ?

monitor.py
```diff
@@ -529,7 +529,13 @@ def parse_entry(self, line: bytes) -> MutableMapping[str, Any]:  # noqa: C901, C
                 # Odyssey: bool
                 self.cmdr = entry['Commander']
                 # 'Open', 'Solo', 'Group', or None for CQC (and Training - but no LoadGame event)
-                self.mode = entry.get('GameMode')
+                if not entry.get('Ship') and not entry.get('GameMode') or entry.get('GameMode').lower() == 'CQC':
+                    logger.trace(f'loadgame to cqc: {entry}')
```
Athanasius, line 533: All new logging.trace should probably make use of a --trace-on switch, so it can be toggled on/off as needed.

See the last part of https://github.com/EDCD/EDMarketConnector/blob/main/Contributing.md#use-logging-not-print-and-definitely-not-sysstdoutwrite

plug.py
A-UNDERSCORE-D: I meant the variable, not the plugin side function name

plug.py
A-UNDERSCORE-D: Yes

monitor.py
norohind: I'll add it, no problems, but for some unknown reason, `trace_on` in this place always is an empty list so it doesn't gonna working.

plug.py
```diff
@@ -312,6 +312,31 @@ def notify_newdata(data, is_beta):
...
+    :returns: Error message from the first plugin that returns one (if any)
+    """
+
+    error = None
+    for plugin in PLUGINS:
+        cqc_callback = plugin._get_func('journal_entry_cqc')
+        if cqc_callback is not None and callable(cqc_callback):
+            try:
+                # Pass a copy of the journal entry in case the callee modifies it
+                newerror = cqc_callback(cmdr, is_beta, dict(entry), dict(state))
```
A-UNDERSCORE-D, line 331: dont use `dict()`, that creates a shallow copy, use `copy.deepcopy` if you want to be sure

plug.py
```diff
@@ -312,6 +312,31 @@ def notify_newdata(data, is_beta):
...
+                error = error or newerror
+
+            except Exception:
+                logger.exception(f'Plugin "{plugin.name}" failed')
```
A-UNDERSCORE-D, line 335: maybe a bit more info here on where it failed? maybe just `failed while handling CQC mode journal entry`?

norohind: I don't think so. `qualname` will show exact place so you can detect if exception raised exactly in `plug.notify_journal_entry_cqc`. But I can add this if you insist.

Athanasius: Whilst we added the detailed "here" logging stuff it's still no substitute for a good log message.  With such you don't *have* to go look up the exact line of code cited.

norohind: Indeed, you're right.

A-UNDERSCORE-D:
>One thing confuses me - monitor.mode, in original code it set to None on EDMC startup and changing on every LoadGame according to GameMode field. On the one side, it will cause passing any journal events to journal_entry_cqc even if player already left CQC and idling in main menu. On the other side, it already causes passing any journal events to journal_entry when player has playing main game and logged out to main menu and nobody complains to it. Give me a pointer about it please.

I missed this in my initial readthrough. I that leaving it as it stands should be fine, the only events that the games gonna give out will be chat (maybe) and music. I dont see any issue in those going into _cqc

norohind:
Thank you for opinion, then I'll left it that.

A-UNDERSCORE-D (approved):
I think Im happy with this if you are @Athanasius

EDMarketConnector.py
```diff
@@ -1171,6 +1174,15 @@ def crewroletext(role: str) -> str:
...
+                err = plug.notify_journal_entry_cqc(monitor.cmdr, monitor.is_beta, entry, monitor.state)
+                if err:
+                    self.status['text'] = err
+                    if not config.get_int('hotkey_mute'):
+                        hotkeymgr.play_bad()
+
+                return  # in CQC
+
             if not entry['event'] or not monitor.mode:
                 # logger.trace('Startup or in CQC, returning')
```
Athanasius, line 1187: If we think we've handled CQC above then this log message, and the return comment, need changing to reflect that this conditional is no longer about that.

monitor.py
```diff
@@ -529,7 +529,15 @@ def parse_entry(self, line: bytes) -> MutableMapping[str, Any]:  # noqa: C901, C
                 # Odyssey: bool
                 self.cmdr = entry['Commander']
                 # 'Open', 'Solo', 'Group', or None for CQC (and Training - but no LoadGame event)
-                self.mode = entry.get('GameMode')
+                if not entry.get('Ship') and not entry.get('GameMode') or entry.get('GameMode').lower() == 'CQC':
```
Athanasius, line 532: `.lower()` here means the other side should be `== 'cqc'`

Athanasius (changes requested):
Just a few minor changes.

norohind:
One more thing that worries me is trace_on in monitor.py, I've tried to trace `cqc-loadgame-events` by including `cqc-loadgame-events` in --trace-on, with --trace, of course, but didn't get it working, trace_on always was an empty list there.

monitor.py
```diff
@@ -529,7 +529,15 @@ def parse_entry(self, line: bytes) -> MutableMapping[str, Any]:  # noqa: C901, C
                 # Odyssey: bool
                 self.cmdr = entry['Commander']
                 # 'Open', 'Solo', 'Group', or None for CQC (and Training - but no LoadGame event)
-                self.mode = entry.get('GameMode')
+                if not entry.get('Ship') and not entry.get('GameMode') or entry.get('GameMode').lower() == 'cqc':
```
norohind, line 532: Just noticed this, if in `entry` key `GameMode` doesn't exists then `entry.get('GameMode').lower()` will raise `AttributeError: 'NoneType' object has no attribute 'lower'`, going to fix it by adding default value `''`.

Athanasius:
> One more thing that worries me is trace_on in monitor.py, I've tried to trace `cqc-loadgame-events` by including `cqc-loadgame-events` in --trace-on, with --trace, of course, but didn't get it working, trace_on always was an empty list there.

We think we've identified the issue with this, but it might get a little tricky to fix properly, so I'll open an issue and address it in another PR.

Issue #1234, opened 2021-08-11
monitor.py: Needs import of config.trace_on adjusting


Athanasius:
It being `from config import ..., trace_on` means it's not seeing the same object as set by:

EDMarketConnector.py
```python
import config as conf_module

conf_module.trace_on = ...
```

Proposed fix, change monitor.py to also `import config as conf_module`

PR #1235, opened 2021-08-11
monitor.py: Access trace_on via whole-config import

Athanasius:
This is necessary so as to see the same instance of config.trace_on as is set up by EDMarketConnector.py.

I've not actually tested this in-game, but debugger showed monitor.py as seeing the same conf_module.trace_on as confirmed parsed in EDMarketConnector.py.

### DIFF
--- a/monitor.py
+++ b/monitor.py
@@ -18,8 +18,9 @@ from typing import Tuple
 if TYPE_CHECKING:
     import tkinter
 
+import config as conf_module  # Necessary to see the same config.trace_on as elsewhere
 import util_ships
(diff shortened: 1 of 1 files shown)


A-UNDERSCORE-D (approved):

Athanasius:
@norohind if you do any more testing could you 100% confirm the logging now works.

norohind:
Well, you already have tested it by yourself but I can still confirm, for me now It also works.

Athanasius:
Closes #1234

PR #1238, opened 2021-08-11
PLUGINS.md: update for `journal_entry_cqc`

norohind:
Maybe I should throw out at all from `state` table keys, that I know for sure will never appear with any meaningful values in CQC.
Also will be needed to correct first version EDMC where `journal_entry_cqc` is present now I stated it as `5.1.3+`.

### DIFF
--- a/PLUGINS.md
+++ b/PLUGINS.md
@@ -748,7 +748,7 @@ New in version 5.1.3+
 def journal_entry_cqc(cmdr: str, is_beta: bool, entry: Dict[str, Any], state: Dict[str, Any]) -> None:
     if entry['event'] == 'Location':
         # We loaded to CQC match, lets detect map!
-        cqc_maps = {  # dict to map systems names to CQC maps
+        cqc_maps = {  # dict to map systems names to CQC maps, ref: https://forums.frontier.co.uk/threads/cqc-systems.234394/
         'Bleae Aewsy GA-Y d1-14': 'Asteria Point',
         'Eta Cephei':             'Cluster Compound',
         'Theta Ursae Majoris':    'Elevate',
--- a/PLUGINS.md
+++ b/PLUGINS.md
@@ -743,7 +743,7 @@ Examples of this are:
 ---
 
 ### Journal entry CQC
-New in version 5.1.3+
+New in version 5.2.0
 ```python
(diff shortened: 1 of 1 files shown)


PLUGINS.md
```diff
@@ -737,6 +740,91 @@ Examples of this are:
    
 ---
 
+---
+
+### Journal entry CQC
+New in version 5.1.3+
```
Athanasius, line 746: I'm targeting 5.2.0 for this, along with CAPI changes.

PLUGINS.md
```diff
@@ -737,6 +740,91 @@ Examples of this are:
...
+```python
+def journal_entry_cqc(cmdr: str, is_beta: bool, entry: Dict[str, Any], state: Dict[str, Any]) -> None:
+    if entry['event'] == 'Location':
+        # We loaded to CQC match, lets detect map!
+        cqc_maps = {  # dict to map systems names to CQC maps
+        'Bleae Aewsy GA-Y d1-14': 'Asteria Point',
```
Athanasius, line 752: The forum reference I know of for these is <https://forums.frontier.co.uk/threads/cqc-systems.234394/>, please add that to the comments here.

Athanasius (changes requested):

Athanasius:
> Maybe I should throw out at all from `state` table keys, that I know for sure will never appear with any meaningful values in CQC.

If you're 100% certain about them, then maybe.  On the other hand this is the same `monitor.state` dict being passed in as for `journal_entry()` and I'm more inclined to remove the table for the _cqc version in the docs and just point out that some of it will not be populated when in CQC.

Has someone checked that entering CQC, particularly from in-game, definitely leads to us resetting `monitor.state` to default/initial values ?

PLUGINS.md
norohind: But this thread doesn't contain such mapping, good news is I am posted it there.

Athanasius:
> Has someone checked that entering CQC, particularly from in-game, definitely leads to us resetting `monitor.state` to default/initial values ?

Answer:  It's literal done in the same `LoadGame` handling that is detecting entering CQC.  So, yes, it'll be the correct state.

norohind:
` I'm more inclined to remove the table for the _cqc version in the docs and just point out that some of it will not be populated when in CQC` that's what I meant.

Athanasius (approved):

PR #1240, opened 2021-08-12
Add helper methods around trace-on and override options

### LINKED ISSUE #1237: Ensure we can quickly find all the current in-code trace_on values
1. Update documentations to **mandate** a fixed format for the tests, including always `import config as conf_module` (or whatever we decide it *must* be).
2. Script to quickly find all of those and compare to a proposed new string value to ensure no clashes.
3. ???
4. Profit!

A-UNDERSCORE-D:
This adds a `trace_if` method to loggers that does the same as checking for trace-on contents, however, it _also_ logs all messages that otherwise wouldnt be logged to a new `trace-all` log level, below trace, which can be activated using --trace-all

closes #1237

### DIFF
--- a/EDMCLogging.py
+++ b/EDMCLogging.py
@@ -48,6 +48,7 @@ from threading import get_native_id as thread_native_id
 from traceback import print_exc
 from typing import TYPE_CHECKING, Tuple, cast
 
+import config as config_mod
 from config import appcmdname, appname, config
 
 # TODO: Tests:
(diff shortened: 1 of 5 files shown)


monitor.py
```diff
@@ -18,13 +18,13 @@
...
     import tkinter
 
-import config as conf_module  # Necessary to see the same config.trace_on as elsewhere
 import util_ships
 from config import config
 from edmc_data import edmc_suit_shortnames, edmc_suit_symbol_localised
 from EDMCLogging import get_main_logger
 
 logger = get_main_logger()
+STARTUP = 'startup'
```
Athanasius, line 27: Now I'm having thoughts about all such constants being defined in `EDMCLogging`.

And if they're part of a dict, set, list, whatever we *might* even be able to list them all on `--help`.

They should certainly have a mandated prefix on the name if not part of some data structure.

Athanasius (changes requested):

EDMCLogging.py
```diff
@@ -77,15 +78,33 @@
...
+        self._log(logging.TRACE_ALL, message, args, **kwargs)  # type: ignore # we added it
+
+    else:
+        # its in there, it gets to end up in regular TRACE
+        self._log(logging.TRACE, message, args, **kwargs)  # type: ignore # we added it
+
+
+logging.Logger.trace_if = _trace_if  # type: ignore
+
+# we cant hide this from `from xxx` imports and I'd really rather no-one other than `logging` had access to it
```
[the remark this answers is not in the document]
A-UNDERSCORE-D: not a hack, just a design choice -- Im just removing the name from the local namespace (and thus what can be imported from the module)

monitor.py
A-UNDERSCORE-D: do we want to couple stuff that much? I didnt really want to make it super constrained, they were just there as a "I dont want to type 'startup' 30 times"

EDMarketConnector.py
```diff
@@ -76,6 +76,12 @@
         action='store_true',
     )
 
+    parser.add_argument(
+        "--trace-all",
+        help="Disable trace-on functionality",
```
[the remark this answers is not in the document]
A-UNDERSCORE-D: It now triggers on both

EDMarketConnector.py
```diff
@@ -107,7 +107,7 @@
 
     parser.add_argument(
         '--trace-on',
-        help='Mark the selected trace logging as active.',
+        help='Mark the selected trace logging as active. * or all will ensure that every possible trace log appears (in the same way as --trace-all)',
```
Athanasius, line 110: More succinctly:

`help='Mark the selected trace logging as active. "*" or "all" is the same as using --trace-all'`

and might make flake8 happy about the line length as a bonus ;)

monitor.py
Athanasius: Saving a whole two characters every time 🤣 OK.

Issue #1239, opened 2021-08-11
EDMC logging on trace level without --trace option


norohind:
- Version: 5.1.3+1a0b4ae
 - Game Version: Live
 - OS: Windows 10
 - OS Locale: Russian

[5.1.3_00.log](https://github.com/EDCD/EDMarketConnector/files/6971782/5.1.3_00.log)
[EDMarketConnector-debug-5.1.3_00.log](https://github.com/EDCD/EDMarketConnector/files/6971783/EDMarketConnector-debug-5.1.3_00.log)

**Describe the bug**
EDMC putting log entries on TRACE level but isn't ran with `--trace` option

**To Reproduce**
1. Run EDMC, I use 5.1.3 release from msi installer
2. Check out debug logs
3. See TRACE level log entries

**Expected behavior**
No TRACE level log entries without --trace CL option

**Screenshots**
None

**Additional context**
I cutted inara events from debug log since it contains my inara api key

A-UNDERSCORE-D:
Intentional. see:

 [# Yes, we always want these rotated files to be at TRACE level](https://github.com/EDCD/EDMarketConnector/blob/e7f8fea250b95e35d274c1a6f36b9dee4bab721d/EDMCLogging.py#L160-L160)

The loglevel is for the console log, rather than for the output. Users on windows running from our binaries wont see the console, but if they could it wouldnt be flooded with trace logs. This is why we have two (sometimes three!) files

norohind:
Then any chance some big trace messages will be under `trace_on`? Inara plugin - one thing worries me.

PR #1241, opened 2021-08-12
inara.py: put events trace logging under `inara-events` trace_on option

norohind:
Idea from #1239 
Since inara plugin produce very huge trace entries and now debug EDMC log is writings at trace level by default, it will be helpful to hide inara plugin's trace messages.

### DIFF
--- a/plugins/inara.py
+++ b/plugins/inara.py
@@ -13,6 +13,7 @@ from typing import Sequence, Union, cast
 
 import requests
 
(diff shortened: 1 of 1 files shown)


norohind:
Then I will replace it by `logger.trace_if` when/if #1240 be merged.

Athanasius:
> Then I will replace it by `logger.trace_if` when/if #1240 be merged.

All merged, so have at it :) .

PR #1246, opened 2021-08-13
Add CL arg `--capi-pretend-down` in order to pretend that CAPI is down

### LINKED ISSUE #1096: CL arg to make all CAPI queries pretend that CAPI is down
**Describe the enhancement**

Title.

**Expected behavior**

Raise a ServerError or similar on any CAPI query

norohind:
Closes #1096 
This PR adds CL arg for EDMarketConnector.py `--capi-pretend-down`, using this arg will make any request to CAPI raise `companion.ServerError` as if CAPI indeed down.
Also it will require add new phrase to translations but now I just want to get feedback about implemnation if all is ok then I'll add translation.

### DIFF
--- a/EDMarketConnector.py
+++ b/EDMarketConnector.py
@@ -130,7 +130,8 @@ if __name__ == '__main__':  # noqa: C901
     parser.add_argument(
         '--capi-pretend-down',
         help='Force to raise ServerError on any CAPI query',
-        action='store_true')
+        action='store_true'
+    )
 
     args = parser.parse_args()
 
--- a/EDMarketConnector.py
+++ b/EDMarketConnector.py
@@ -127,7 +127,17 @@ if __name__ == '__main__':  # noqa: C901
                         nargs='*'
                         )
 
+    parser.add_argument(
+        '--capi-pretend-down',
+        help='Force to raise ServerError on any CAPI query',
+        action='store_true')
+
     args = parser.parse_args()
+
+    if args.capi_pretend_down:
+        import config as conf_module
+        logger.info('Pretending CAPI is down')
+        conf_module.capi_pretend_down = True
 
     level_to_set: Optional[int] = None
     if args.trace or args.trace_on:
--- a/companion.py
+++ b/companion.py
@@ -25,6 +25,7 @@ from typing import TYPE_CHECKING, Any, Dict, List, Mapping, Optional, OrderedDic
 import requests
 
 from config import appname, appversion, config
+import config as conf_module
 from edmc_data import companion_category_map as category_map
 from EDMCLogging import get_main_logger
 from monitor import monitor
@@ -276,6 +277,9 @@ class Auth(object):
 
             logger.debug('Attempting refresh with Frontier...')
             try:
+                if conf_module.capi_pretend_down:
+                    raise ServerError(_('Pretending CAPI is down'))
+
                 r = self.session.post(SERVER_AUTH + URL_TOKEN, data=data, timeout=auth_timeout)
                 if r.status_code == requests.codes.ok:
                     data = r.json()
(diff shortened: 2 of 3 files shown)


Athanasius:
Closes #1096

companion.py
```diff
@@ -276,6 +277,9 @@ def refresh(self) -> Optional[str]:
 
             logger.debug('Attempting refresh with Frontier...')
             try:
+                if conf_module.capi_pretend_down:
```
Athanasius, line 280: This isn't CAPI, this is Frontier Auth, they're not the same.  Remove this.

companion.py
```diff
@@ -360,6 +364,9 @@ def authorize(self, payload: str) -> str:  # noqa: CCR001
             # requests_log.setLevel(logging.DEBUG)
             # requests_log.propagate = True
 
+            if conf_module.capi_pretend_down:
```
Athanasius, line 367: This is also Frontier Auth, not CAPI.  Remove the change.

companion.py
```diff
@@ -565,6 +572,9 @@ def query(self, endpoint: str) -> CAPIData:  # noqa: CCR001
 
         try:
             logger.trace_if('capi.query', 'Trying...')
+            if conf_module.capi_pretend_down:
```
Athanasius, line 575: This is actually CAPI.  However, please add `{endpoint}` to the exception text.  It's that or you'd instead need to do it at the station etc call sites *into* `query()`.

Athanasius: It doesn't matter if that results in a very wide GUI window due to it being in the status line, this is only for developer use anyway.

Athanasius (changes requested):
This should only affect actual CAPI queries, not the Frontier Auth flow.

companion.py
norohind: In GUI it will show `Frontier CAPI query failure: {endpoint}` anyway, but I've added endpoint to `ServerError` exception's message, no problems.

EDMarketConnector.py
```diff
@@ -127,8 +127,18 @@
                         nargs='*'
                         )
 
+    parser.add_argument(
+        '--capi-pretend-down',
+        help='Force to raise ServerError on any CAPI query',
+        action='store_true')
```
A-UNDERSCORE-D, line 133:
```suggestion
        action='store_true'
    )
```
Please use hanging braces if you want to multiline like this

A-UNDERSCORE-D (changes requested):

companion.py
Athanasius: > In GUI it will show `Frontier CAPI query failure: {endpoint}` anyway, but I've added endpoint to `ServerError` exception's message, no problems.

Except this means I now realise the code needs further adjustments.  I do want it to be explicit when it's the "pretend" case, stay tuned for further review comments.

companion.py
```diff
@@ -565,6 +566,9 @@ def query(self, endpoint: str) -> CAPIData:  # noqa: CCR001
 
         try:
             logger.trace_if('capi.query', 'Trying...')
```
Athanasius, line 568: Move this logging and the pretend check just above the `try:`.

companion.py
```diff
@@ -565,6 +566,9 @@ def query(self, endpoint: str) -> CAPIData:  # noqa: CCR001
...
+            if conf_module.capi_pretend_down:
+                raise ServerError(f'Pretending CAPI is down for {endpoint} endpoint')
```
Athanasius, line 570: Change this to

`raise ServerConnectionError(f'Pretending CAPI down: {endpoint}')`

Athanasius (approved):

A-UNDERSCORE-D (approved):

PR #1261, opened 2021-08-24
Change all CAPI queries over to a queue/thread model

Athanasius:
Basic tests of both the GUI app and EDMC.py show this as working.

### DIFF
--- a/EDMarketConnector.py
+++ b/EDMarketConnector.py
@@ -1611,9 +1611,11 @@ class AppWindow(object):
             initialdir=config.get_str('outdir'),
             initialfile=f'{monitor.system}{monitor.station}.{timestamp}'
         )
-        if f:
-            with open(f, 'wb') as h:
-                h.write(str(companion.session.capi_raw_data).encode(encoding='utf-8'))
+        if not f:
+            return
+
+        with open(f, 'wb') as h:
+            h.write(str(companion.session.capi_raw_data).encode(encoding='utf-8'))
 
     # def exit_tray(self, systray: 'SysTrayIcon') -> None:
     #     """Tray icon is shutting down."""
--- a/EDMarketConnector.py
+++ b/EDMarketConnector.py
@@ -1113,10 +1113,9 @@ class AppWindow(object):
 
         except companion.CredentialsError:
             companion.session.retrying = False
-            # TODO: Might need to .invalidate() here to avoid "already auth'd"
             companion.session.invalidate()
             companion.session.login()
-            return
+            return  # We need to give Auth time to complete, so can't set a timed retry
 
         # Companion API problem
         except companion.ServerLagging as e:
--- a/EDMarketConnector.py
+++ b/EDMarketConnector.py
@@ -1113,6 +1113,7 @@ class AppWindow(object):
 
         except companion.CredentialsError:
             companion.session.retrying = False
+            # TODO: Might need to .invalidate() here to avoid "already auth'd"
             companion.session.login()
 
         # Companion API problem
--- a/EDMarketConnector.py
+++ b/EDMarketConnector.py
@@ -1106,6 +1106,8 @@ class AppWindow(object):
                                 self.suit['text'] = f'{suitname} ({loadout_name})'
 
                 self.suit_show_if_set()
+                # Update Odyssey Suit data
+                companion.session.suit_update(capi_response.capi_data)
 
                 if capi_response.capi_data['commander'].get('credits') is not None:
                     monitor.state['Credits'] = capi_response.capi_data['commander']['credits']
@@ -1173,9 +1175,6 @@ class AppWindow(object):
 
         if capi_response.play_sound and play_bad:
             hotkeymgr.play_bad()
-
-        # Update Odyssey Suit data
-        companion.session.suit_update(capi_response.capi_data)
 
         logger.trace_if('capi.worker', 'Updating suit and cooldown...')
         self.update_suit_text()
--- a/EDMarketConnector.py
+++ b/EDMarketConnector.py
@@ -900,6 +900,7 @@ class AppWindow(object):
 
         :param event: Tk generated event details.
         """
+        logger.trace_if('capi.worker', 'Begin')
         auto_update = not event
         play_sound = (auto_update or int(event.type) == self.EVENT_VIRTUAL) and not config.get_int('hotkey_mute')
 
@@ -937,6 +938,7 @@ class AppWindow(object):
         query_time = int(time())
         logger.trace_if('capi.worker', 'Requesting full station data')
         config.set('querytime', query_time)
+        logger.trace_if('capi.worker', 'Calling companion.session.station')
         companion.session.station(
             query_time=query_time, tk_response_event=self._CAPI_RESPONSE_TK_EVENT_NAME,
             play_sound=play_sound
@@ -944,6 +946,7 @@ class AppWindow(object):
 
     def capi_handle_response(self, event=None):  # noqa: C901, CCR001
         """Handle the resulting data from a CAPI query."""
+        logger.trace_if('capi.worker', 'Handling response')
         play_bad: bool = False
         err: Optional[str] = None
 
@@ -1151,9 +1154,11 @@ class AppWindow(object):
         # Update Odyssey Suit data
         companion.session.suit_update(capi_response.capi_data)
 
+        logger.trace_if('capi.worker', 'Updating suit and cooldown...')
         self.update_suit_text()
         self.suit_show_if_set()
         self.cooldown()
+        logger.trace_if('capi.worker', '...done')
 
     def journal_event(self, event):  # noqa: C901, CCR001 # Currently not easily broken up.
         """
--- a/EDMarketConnector.py
+++ b/EDMarketConnector.py
@@ -935,77 +935,86 @@ class AppWindow(object):
 
         querytime = int(time())
         logger.trace_if('capi.worker', 'Requesting full station data')
-        companion.session.station(querytime=querytime, play_sound=play_sound)
+        companion.session.station(
+            querytime=querytime, retrying=retrying, play_sound=play_sound
+        )
         config.set('querytime', querytime)
 
-
-    def capi_handle_response(self, event=None):
+    def capi_handle_response(self, event=None):  # noqa: C901, CCR001
         """Handle the resulting data from a CAPI query."""
-        play_bad = False
+        play_bad: bool = False
         err: Optional[str] = None
 
-        data: Union[companion.CAPIData, companion.CAPIFailedRequest]
-        querytime: int
-        play_sound: bool
-        auto_update: bool
+        capi_response: Union[companion.CAPIFailedRequest, companion.EDMCCAPIResponse]
         try:
             logger.trace_if('capi.worker', 'Pulling answer off queue')
-            data, querytime, play_sound, auto_update = self.capi_response_queue.get(block=False)
-            if isinstance(data, companion.CAPIFailedRequest):
-                logger.trace_if('capi.worker', f'Failed Request: {data.message}')
-                if data.exception:
-                    raise data.exception
+            capi_response = self.capi_response_queue.get(block=False)
+            if isinstance(capi_response, companion.CAPIFailedRequest):
+                logger.trace_if('capi.worker', f'Failed Request: {capi_response.message}')
+                if capi_response.exception:
+                    raise capi_response.exception
 
                 else:
-                    raise ValueError(data.message)
+                    raise ValueError(capi_response.message)
 
             logger.trace_if('capi.worker', 'Answer is not a Failure')
+            if not isinstance(capi_response, companion.EDMCCAPIResponse):
+                msg = f"Response was neither CAPIFailedRequest nor EDMCAPIResponse: {type(capi_response)}"
+                logger.error(msg)
+                raise ValueError(msg)
+
             # Validation
-            if 'commander' not in data:
+            if 'commander' not in capi_response.capi_data:
                 # This can happen with EGS Auth if no commander created yet
                 # LANG: No data was returned for the commander from the Frontier CAPI
                 err = self.status['text'] = _('CAPI: No commander data returned')
 
-            elif not data.get('commander', {}).get('name'):
+            elif not capi_response.capi_data.get('commander', {}).get('name'):
                 # LANG: We didn't have the commander name when we should have
                 err = self.status['text'] = _("Who are you?!")  # Shouldn't happen
 
-            elif (not data.get('lastSystem', {}).get('name')
-                  or (data['commander'].get('docked')
-                      and not data.get('lastStarport', {}).get('name'))):
+            elif (not capi_response.capi_data.get('lastSystem', {}).get('name')
+                  or (capi_response.capi_data['commander'].get('docked')
+                      and not capi_response.capi_data.get('lastStarport', {}).get('name'))):
                 # LANG: We don't know where the commander is, when we should
                 err = self.status['text'] = _("Where are you?!")  # Shouldn't happen
 
-            elif not data.get('ship', {}).get('name') or not data.get('ship', {}).get('modules'):
+            elif (
+                    not capi_response.capi_data.get('ship', {}).get('name')
+                    or not capi_response.capi_data.get('ship', {}).get('modules')
+            ):
                 # LANG: We don't know what ship the commander is in, when we should
                 err = self.status['text'] = _("What are you flying?!")  # Shouldn't happen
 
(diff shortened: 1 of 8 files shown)


Athanasius:
Still to do:

- [ ] File > Status (stats.py) needs changing to use the cached data.
- [ ] Instead of doing *any* (re)auth in the capi request flow it needs to return a response in the queue to get the main thread to do that.

companion.py
```diff
@@ -463,27 +520,119 @@ def base64_url_encode(self, text: bytes) -> str:
...
+            target=self.capi_query_worker,
+            daemon=True,
+            name='CAPI worker'
+        )
+        self.capi_query_thread.start()
+        logger.debug('Done')
+
+    def set_capi_response_queue(self, capi_response_queue: Queue) -> None:
+        """Set a reference to the CAPI response data queue."""
+        self.capi_response_queue = capi_response_queue
```
A-UNDERSCORE-D, line 611: does this need to be a setter?
if so, would a property not work better?

companion.py
```diff
@@ -463,27 +520,119 @@ def base64_url_encode(self, text: bytes) -> str:
...
         self.credentials: Optional[Dict[str, Any]] = None
         self.session: Optional[requests.Session] = None
         self.auth: Optional[Auth] = None
         self.retrying = False  # Avoid infinite loop when successful auth / unsuccessful query
+        self.tk_master: Optional[tk.Tk] = None
+
+        self.capi_raw_data = CAPIDataRaw()
+        logger.debug('Starting CAPI queries thread...')
+        self.capi_response_queue: Queue
+        self.capi_query_queue: Queue = Queue()
```
A-UNDERSCORE-D, line 600: Please annotate the content of these queues

companion.py
```diff
@@ -545,176 +694,356 @@ def auth_callback(self) -> None:
...
+                        (
+                            CAPIFailedRequest(
+                                message=e.args,
+                                exception=e,
+                                query_time=query.query_time,
+                                play_sound=query.play_sound,
+                                auto_update=query.auto_update
+                            ),
+                        )
+                    )
```
A-UNDERSCORE-D, line 904: This is a tuple, but the below isnt. Is that intentional?

companion.py
```diff
@@ -545,176 +694,356 @@ def auth_callback(self) -> None:
...
+            # If the query came from EDMC.(py|exe) there's no tk to send an
+            # event too, so assume it will be polling there response queue.
+            if query.tk_response_event is not None:
+                self.tk_master.event_generate('<<CAPIResponse>>')
+
+        logger.info('CAPI worker thread DONE')
+
+    def capi_query_close_worker(self) -> None:
+        """Ask the CAPI query thread to finish."""
+        self.capi_query_queue.put(None)
```
A-UNDERSCORE-D, line 952: do we maybe want a better sentinel here?

companion.py
```diff
@@ -545,176 +694,356 @@ def auth_callback(self) -> None:
...
+                retrying=retrying,
+                query_time=query_time,
+                play_sound=play_sound,
+                auto_update=auto_update
+            )
+        )
+
+    def profile(
+            self,
+            query_time: int = int(time.time()),
```
A-UNDERSCORE-D, line 999: This does not work as you expect. Its evaluated once at class instantiation time.

```
In [1]: from time import time

In [2]: class Test:
   ...:     def print_time(self, t: int = int(time())):
   ...:         print(t)
   ...:

In [3]: x = Test()

In [4]: x.print_time()
1629847121

In [5]: x.print_time()
1629847121
```

you will want to use an `Optional[int]` and set it within the method

A-UNDERSCORE-D (changes requested):
First pass, this will want multiple passes and some playtesting

EDMarketConnector.py
```diff
@@ -4,19 +4,20 @@
...
 import pathlib
+import queue
 import re
 import sys
 # import threading
 import webbrowser
 from builtins import object, str
 from os import chdir, environ
 from os.path import dirname, join
+from queue import Queue
```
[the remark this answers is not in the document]
Athanasius: Converted the bare 'Queue' to 'queue.Queue', as it's only in the declaration and initialisation of the queue.

EDMarketConnector.py
```diff
@@ -925,56 +935,89 @@ def getandsend(self, event=None, retrying: bool = False):  # noqa: C901, CCR001
             self.button['state'] = self.theme_button['state'] = tk.DISABLED
             self.w.update_idletasks()
 
+        query_time = int(time())
+        logger.trace_if('capi.worker', 'Requesting full station data')
+        companion.session.station(
+            query_time=query_time, tk_response_event=self._CAPI_RESPONSE_TK_EVENT_NAME,
+            retrying=retrying, play_sound=play_sound
+        )
+        config.set('querytime', query_time)
```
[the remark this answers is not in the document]
Athanasius: Yes, moved it up for now.  Ideally this would get set exactly where we (first) make an actual CAPI query, as the idea is to be on cooldown versus that.  But that will require more-intrusive changes.

companion.py
Athanasius: Actually I'll rework this to create the queue in Session now.  This is mostly an artifact of the order I started changing the code.  The callers will have `companion.session` to reference it.

companion.py
Athanasius: Well, we could define a special '__EDMC_WORKER_SHUTDOWN' 'endpoint' and pass that in, and that would at least remove a need for 'Optional' on the queue member types declaration.

companion.py
Athanasius: Changed ... although we might end up removing the whole function.  It's only still there 'for' stats.py, and I intend to change that to use cached data anyway.

companion.py
A-UNDERSCORE-D: That works for me

companion.py
```diff
@@ -128,6 +125,49 @@ def check_modules_ships(self) -> None:
...
+
+class CAPIDataRaw:
+    """The last obtained raw CAPI response for each endpoint."""
+
+    raw_data: Dict[str, CAPIDataRawEndpoint] = {}
+
+    def record_endpoint(
+            self, endpoint: str,
+            raw_data: str,
+            query_time: datetime.datetime = datetime.datetime.utcnow()
```
A-UNDERSCORE-D, line 153: Only evaluated at class parse time.

companion.py
```diff
@@ -463,27 +520,121 @@ def base64_url_encode(self, text: bytes) -> str:
...
+    ######################################################################
+    def start_frontier_auth(self, access_token: str) -> None:
+        """Start an oAuth2 session."""
+        logger.debug('Starting session')
+        self.session = requests.Session()
+        self.session.headers['Authorization'] = f'Bearer {access_token}'
+        self.session.headers['User-Agent'] = USER_AGENT
+        self.state = Session.STATE_OK
 
     def login(self, cmdr: str = None, is_beta: Union[None, bool] = None) -> bool:
```
A-UNDERSCORE-D, line 632: I know you didnt touch this, but, `Optional[bool]`

companion.py
```diff
@@ -545,176 +696,357 @@ def auth_callback(self) -> None:
...
+                return station_data
+
+            # Sanity checks in case data isn't as we expect, and maybe 'docked' flag
+            # is also lagging.
+            if (last_starport := station_data.get('lastStarport')) is None:
+                logger.error("No lastStarport in data!")
+                return station_data
+
+            if ((last_starport_name := last_starport.get('name')) is None
+                    or last_starport_name == ''):
```
A-UNDERSCORE-D, line 831:
```suggestion
            if (
                (last_starport_name := last_starport.get('name')) is None
                or last_starport_name == ''
            ):
```

companion.py
```diff
@@ -545,176 +696,357 @@ def auth_callback(self) -> None:
...
+                shipyard_data = capi_single_query(self.FRONTIER_CAPI_PATH_SHIPYARD, timeout=timeout)
+                if last_starport_id != int(shipyard_data['id']):
+                    logger.warning(f"{last_starport_id!r} != {int(shipyard_data['id'])!r}")
+                    raise ServerLagging()
 
-        return data
+                else:
+                    shipyard_data['name'] = last_starport_name
+                    station_data['lastStarport'].update(shipyard_data)
+            # WORKAROUND END
```
A-UNDERSCORE-D, line 876: Not entirely sure this should be here?

A-UNDERSCORE-D (changes requested):

companion.py
```diff
@@ -545,176 +698,361 @@ def auth_callback(self) -> None:
...
+                    self.capi_response_queue.put(
+                        (
+                            EDMCCAPIFailedRequest(
+                                message=e.args,
+                                exception=e,
+                                query_time=query.query_time,
+                                play_sound=query.play_sound,
+                                auto_update=query.auto_update
+                            ),
+                        )
```
A-UNDERSCORE-D, line 931: tuple

companion.py
Athanasius: Changed to mandatory, and set at the call site.

companion.py
Athanasius: It's the same coverage of the WORKAROUND as in `develop`.  Ask past A_D why they had the end of it be that far down ;)

A-UNDERSCORE-D: Oh. well that guy obviously made a mistake. I thiiink? Not sure. probably leave it in that case there was probably a reason

EDMarketConnector.py
```diff
@@ -925,56 +934,89 @@ def getandsend(self, event=None, retrying: bool = False):  # noqa: C901, CCR001
...
+                logger.trace_if('capi.worker', f'Failed Request: {capi_response.message}')
+                if capi_response.exception:
+                    raise capi_response.exception
+
+                else:
+                    raise ValueError(capi_response.message)
+
+            logger.trace_if('capi.worker', 'Answer is not a Failure')
+            if not isinstance(capi_response, companion.EDMCCAPIResponse):
+                msg = f"Response was neither CAPIFailedRequest nor EDMCAPIResponse: {type(capi_response)}"
```
A-UNDERSCORE-D, line 964: Quote style

```suggestion
                msg = f'Response was neither CAPIFailedRequest nor EDMCAPIResponse: {type(capi_response)}'
```

companion.py
```diff
@@ -463,30 +528,127 @@ def base64_url_encode(self, text: bytes) -> str:
         return base64.urlsafe_b64encode(text).decode().replace('=', '')
 
 
+class EDMCCAPIReturn:
```
A-UNDERSCORE-D, line 531: This could be a dataclass, or namedtuple

A-UNDERSCORE-D (changes requested):
Not a complete review, just looked over while waiting for something

companion.py
```diff
@@ -167,6 +167,14 @@ def __str__(self):
 
         return capi_data_str
 
+    def __iter__(self):
+        """Make this iterable on its raw_data dict."""
+        yield from self.raw_data
+
+    def __getitem__(self, item):
+        """Make the raw_data dict's items get'able."""
+        return self.raw_data.get(item)
```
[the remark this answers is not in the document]
Athanasius: OK, as we only *need* `__getitem__` I've changed it to using that.

There are currently two uses of this:

1. stats.py explicitly checks the key exists before trying to use it.
2. EDMarketConnector.py AppWindow.save_raw_data() just iterates over all the data.

EDMarketConnector.py
Athanasius: Probably a consequence of original wording have an apostrophe in it.  Changed.

companion.py
Athanasius: I don't think it could be a namedtuple, as it's the base class for the Request, Response and Failed classes.

I'll check if they can all be dataclasses though.  Looks like I can lose the `super().__init__()` calls if so.

Athanasius: Nope, due to the mix of defaulted and non-defaulted variables that's proving problematic with inheritance:

```python
@dataclass
class Base:

  var1: int
  var2: bool = False


class Inherits(Base):

  var3: str
```
Leads to it complaining that `Inherits.var3` "follows default argument", because it's effectively after `var2`.

Athanasius:
Later today I'll review the EDMCCAPI* classes to see if it does make sense to make them dataclasses, by reducing the number of arguments with defaults.  However, my thinking is that the callers should not have to care about supplying the values unless they know they need them to be non-default.  EDMC.py in particular is good to call companion.Session.station() with only query_time specified.

Once that is resolved I think this is good enough to merge into `develop` and start testing in actual use as we work on other CAPI related issues ready for 5.2.0.

EDMarketConnector.py
```diff
@@ -886,32 +891,37 @@ def export_market_data(self, data: CAPIData) -> bool:  # noqa: CCR001
...
-        play_bad = False
-        err: Optional[str] = None
 
         if (
                 not monitor.cmdr or not monitor.mode or monitor.state['Captain']
                 or not monitor.system or monitor.mode == 'CQC'
         ):
+            logger.trace_if('capi.worker', 'CQC detected, aborting query')
+            # LANG: CAPI queries aborted because player is in CQC (Arena)
+            self.status['text'] = _('CQC detected, aborting CAPI query')
```
norohind, line 913: But it also could be a mutlicrew, should it be clarified?

Athanasius: Good point. I'll go fix that up.

Athanasius: Ugh, I kind of think we want to test for things individually here, so as to have a specific message in each case.  More translation strings incoming!

companion.py
```diff
@@ -545,176 +707,344 @@ def auth_callback(self) -> None:
...
+
+                else:
+                    self.capi_response_queue.put(
+                        EDMCCAPIResponse(
+                            capi_data=capi_data,
+                            query_time=query.query_time,
+                            play_sound=query.play_sound,
+                            auto_update=query.auto_update
+                        )
+                    )
```
A-UNDERSCORE-D, line 937: Im sure we can maybe mess this such that theres a single try/except/else ? possibly with a wrapper function?

companion.py
```diff
@@ -128,6 +125,57 @@ def check_modules_ships(self) -> None:
...
+        """Return a more readable string form of the data."""
+        capi_data_str = '{'
+        for e in self.raw_data.keys():
+            capi_data_str += f'"{e}":\n{{\n\t"query_time": "{self.raw_data[e].query_time}",\n\t' \
+                             f'"raw_data": {self.raw_data[e].raw_data}\n}},\n\n'
+
+        capi_data_str = capi_data_str.removesuffix(',\n\n')
+        capi_data_str += '\n\n}'
+
+        return capi_data_str
```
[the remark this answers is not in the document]
Athanasius: This isn't working.

1. The {} weren't encompassing the correct thing, but after fixing that...
2. the v.raw_data is a *string*, and that means when json.dumps()'d it all gets escaped, rather than being the raw data.  I do **not** want to json.loads() that only to json.dumps() it again, as it's meant to be the *raw* string (other than decode/encode passes).

I'll have a bit more of a play to see if I can at least use a generator.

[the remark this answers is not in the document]
Athanasius: Nope, wanting the **raw** data is fundamentally incompatible with doing this.  I have changed my loop to `for k, v in self.raw_data.items()`, and as a consequence the references in the loop, though.

companion.py
```diff
@@ -128,6 +125,57 @@ def check_modules_ships(self) -> None:
...
+
+    def __iter__(self):
+        """Make this iterable on its raw_data dict."""
+        yield from self.raw_data
```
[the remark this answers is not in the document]
Athanasius:
```python
    def __init__(self):
        self.__iter__ = self.raw_data.__iter__
        self.__getitem__ = self.raw_data.__iter__

  File "C:\Users\Athan\Documents\Devel\EDMarketConnector\stats.py", line 272, in __init__
    self.showstats()
  File "C:\Users\Athan\Documents\Devel\EDMarketConnector\stats.py", line 280, in showstats
    if companion.session.FRONTIER_CAPI_PATH_PROFILE not in companion.session.capi_raw_data:
TypeError: argument of type 'CAPIDataRaw' is not iterable```

Same with just `__iter__ = ...`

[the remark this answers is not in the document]
Athanasius: And it doesn't help if I define dummy 'pass' version of the functions either.  Weirdly PyCharm debugging of the `__init__()` doesn't even show them as members at all.

companion.py
Athanasius: Possibly, and I am now wondering why I did `str(e.args)` in one place but not the other.

I think the real fix is to move the try/except clauses in an indent.  I'll play with that.

A-UNDERSCORE-D (approved):

PR #1283, opened 2021-09-24
Remove protocol.protocolhandler singleton auto-init

A-UNDERSCORE-D:
In general, doing things like this on import is bad, but this was
changed specifically to remove a bug that causes
--force-localserver-auth to do nothing.

That is caused because while we were careful not to import protocol
until after we were done doing arg things, we did not check to make sure
something else didn't. Companion imports protocol, which thus always
instantiates WindowsProtocolHandler before we can modify some config
state to indicate that we want LinuxProtocolHandler.

### DIFF
--- a/EDMarketConnector.py
+++ b/EDMarketConnector.py
@@ -377,7 +377,7 @@ from edmc_data import ship_name_map
 from hotkey import hotkeymgr
 from l10n import Translations
 from monitor import monitor
-from protocol import protocolhandler
+import protocol
 from theme import theme
 from ttkHyperlinkLabel import HyperlinkLabel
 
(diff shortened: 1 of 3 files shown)


Athanasius:
Could we not just move the 'import companion' late enough to not cause this issue ?   I've experimentally commented out the top-level import and put one into the `__main__` code just before the "UI Scaling" section.  It does appear to work (including forgetting auth then re-running and confirming auth works, when running from source).

A-UNDERSCORE-D:
We *can* do that, but that just shuffles the problem down to six  more months down the road when some clever person helpfully reorders stuff)

PR #1343, opened 2021-11-12
Fully implement systray, and minimising to it.

### LINKED ISSUE #126: Minimize to system tray option (Windows)

Athanasius:
I finally got around to uncommenting all the infi.systray code.  It works with the py2exe we're now using.  I also made it actually minimise **to** the  system tray, if that option is active.  Before the tweak the taskbar icon would still show.

Closes #126

### DIFF
--- a/prefs.py
+++ b/prefs.py
@@ -60,17 +60,17 @@ class PrefsVersion:
     def __init__(self):
         return
 
-    def stringToSerial(self, version_str: str) -> int:  # noqa: N802 # used in plugins
+    def stringToSerial(self, versionStr: str) -> int:  # noqa: N802, N803 # used in plugins
         """
         Convert a version string into a preferences version serial number.
 
         If the version string isn't known returns the 'current' (latest) serial number.
 
-        :param version_str:
+        :param versionStr:
         :return int:
         """
-        if version_str in self.versions:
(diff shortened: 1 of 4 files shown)


prefs.py
```diff
@@ -60,17 +60,17 @@ class PrefsVersion:
     def __init__(self):
         return
 
-    def stringToSerial(self, versionStr: str) -> int:  # noqa: N802 # used in plugins
+    def stringToSerial(self, version_str: str) -> int:  # noqa: N802 # used in plugins
         """
```
A-UNDERSCORE-D, line 64: Cant rename the param; will break keyword args, unless you capture everything in `**kwargs` and check

A-UNDERSCORE-D (approved):

prefs.py
Athanasius: Despite the comment, this isn't documented in PLUGINS.md, so isn't actually a published interface.  I probably had a reason for the 'used in plugins' comment though.

PR #1368, opened 2021-12-07
Automate always including **all** files that py2exe places in `dist.win32`

Athanasius:
**NB: This is currently based off the Python 3.10 branch, I'll rebase once that is merged.**

Until now we've been winging it, hoping there's an obvious error if there's a file that py2exe generates inside `dist.win32`, but that we don't then package into the Windows installer.

1. Use WiX heat.exe to generate the list of files inside `dist.win32`.
2. Load that, find the appropriate tree, change the attributes as necessary.
3. Load our template file, find the ProgramFilesFolder element and insert all of the file list into it.
4. Find the Feature element in the template and insert both the RegistryEntries and all file Components as ComponentRefs with the matching Id attributes.
5. Write this out as a new EDMarketConnector.wxs.
6. Continue with the installer build as before.

In testing, as 5.3.0-beta2, this resulted in all files matching between `dist.win32` and the eventual installation folder ~~the only file in `dist.win32` that didn't end up in the installation being `EDMarketConnector.wixobj`.  That's because it's generated *later*, and we should perhaps move its location to either the root or the `wix` directory.~~


Athanasius:
NB: I neglected to ensure the `L10n/zh-Hans.strings` -> `L10n/zh-CN.strings`  rename in the template-using code.  So that needs addressing before this is merged.

Athanasius:
> NB: I neglected to ensure the `L10n/zh-Hans.strings` -> `L10n/zh-CN.strings` rename in the template-using code. So that needs addressing before this is merged.

Given this would also affect users running from source I'm more inclined to fix this by making a script to do the "unzip the onesky ZIP file, copy files over the in-tree ones" and include the rename in that.

Athanasius:
Actually, now that I check, it Just Works from source.

1. We iterate the `L10n` directory for all `*.strings` files to build the list of available languages.
2. From that we use the prefix (before `.strings`) of the filename as the key.
3. We then use that key throughout the code, so it doesn't even matter what it *is* other than giving someone looking in the directory an idea of what language each file is for.

So I'm inclined to not bother preserving this rename.

I also note that L10n.py:_Translations.available() gets called a lot, scanning the directory each time.  Upside is you could change a language file and swap to another language and back in order to reload it, even if it's a *new* translation you then switch to.  So, despite this feeling like a waste of resources I'm inclined to leave it.

Athanasius:
Thining about this more, that zh-Hans -> zh-CN mapping is likely to do with the 'Default' language.

This is selected using `l10n.py:_Locale:preferred_languages()`.  We probably need to tweak this to perform such mappings on the found arrays.

l10n.py
```diff
@@ -320,34 +320,45 @@ def preferred_languages(self) -> Iterable[str]:
...
+                buf = ctypes.create_unicode_buffer(size.value)
+
+                if GetUserPreferredUILanguages(
+                        MUI_LANGUAGE_NAME, ctypes.byref(num), ctypes.byref(buf), ctypes.byref(size)
+                ):
+                    languages = wszarray_to_list(buf)
+
+        # OneSky calls "Chinese Simplified" "zh-Hans" in the name of the file,
+        # but that will be zh-CN in terms of locale.  So map zh-CN -> zh-Hans
+        languages = ['zh-Hans' if lang == 'zh-CN' else lang for lang in languages]
```
A-UNDERSCORE-D, line 359: Probably a HACK or other note on this?

Issue #1155, opened 2021-06-08
Include geometry (window position) in --reset-ui


fasteddy516:
**Describe the enhancement**

It would be helpful to delete/reset the `geometry` (main window start position) setting as part of the `--reset-ui` process.

**Expected behavior**

After running EDMC with the `--reset-ui` option, the main window opens in its default position on the primary display.

**Considered workarounds**

When needed, I currently manually delete the `geometry` registry key (on a Windows machine) to restore the window to its default position on startup.

**Additional context**

On a Windows machine, the EDMC main window doesn't seem to be moveable unless you can see it and grab it with the mouse. 
 This makes it possible to "lose" the window in complicated/distributed multi-monitor setups, or on systems that have "phantom" displays that are always present for remote access apps, portable USB monitors and such.  When other apps open up off-screen, I can use Windows+Cursor keys or shift-click on the icon in the taskbar and select `Move` to reposition them to a visible screen without actually being able to see them.  These options don't work with EDMC's main window.  (Tkinter-related?  Maybe because the window size is locked?)  Deleting the `geometry` key seems to be the simplest way to get the window back to a known position/display.

Athanasius:
And also reset the `ui_scale` as well.

A-UNDERSCORE-D:
Closed by above PR

PR #1404, opened 2022-01-20
Reset geometry and UI scale with --reset-ui

A-UNDERSCORE-D:
Just a simple addon of a few config resets.

This likely needs a test as I have not yet done so.

closes EDCD/EDMarketConnector#1155

### DIFF
--- a/EDMarketConnector.py
+++ b/EDMarketConnector.py
@@ -74,7 +74,7 @@ if __name__ == '__main__':  # noqa: C901
     ###########################################################################
     parser.add_argument(
         '--reset-ui',
-        help='reset UI theme and transparency to defaults',
+        help='Reset UI theme, transparency, font, font size, ui scale, and ui geometry to default',
         action='store_true'
     )
     ###########################################################################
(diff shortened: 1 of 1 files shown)


EDMarketConnector.py
```diff
@@ -1867,7 +1867,11 @@ def show_killswitch_poppup(root=None):
         config.set('ui_transparency', 100)  # 100 is completely opaque
         config.delete('font')
         config.delete('font_size')
-        logger.info('reset theme, font, font size, and transparency to default.')
+
+        config.set('ui_scale', 100)  # 100% is the default here
+        config.delete('geometry')    # unset is recreated by other code
+
+        logger.info('reset theme, transparency, font, font size, ui scale, and ui geometry to default.')
```
Athanasius, line 1874: Needs the `help` on the parser.add_argument() also listing all of this.

EDMarketConnector.py
```diff
@@ -1867,7 +1867,11 @@ def show_killswitch_poppup(root=None):
         config.set('ui_transparency', 100)  # 100 is completely opaque
         config.delete('font')
```
Athanasius, line 1868: Looks like either config.py needs to handle "that key doesn't exist", or this needs a try/except:

```python
  File "C:\Users\Athan\Documents\Devel\EDMarketConnector\EDMarketConnector.py", line 1868, in <module>
    config.delete('font')
  File "C:\Users\Athan\Documents\Devel\EDMarketConnector\config.py", line 666, in delete
    winreg.DeleteValue(self.__reg_handle, key)
FileNotFoundError: [WinError 2] The system cannot find the file specified
```

Athanasius (approved):

PR #1407, opened 2022-01-20
Switch logging times to be in UTC

A-UNDERSCORE-D:
Draft because there are some questions of format involved

### DIFF
--- a/EDMCLogging.py
+++ b/EDMCLogging.py
@@ -92,9 +92,10 @@ logging.Logger.trace = lambda self, message, *args, **kwargs: self._log(  # type
     **kwargs
 )
 
-# make logging use UTC for times, to help make our logs congruent with journals etc.
-# Note that as this is the local system vs the remote system (ED servers, for journals), times may not be perfectly
-# in sync. (something something NTP)
+# MAGIC n/a:  2020-01-20: make logging use UTC for times, to help make our logs congruent with journals etc.
+# MAGIC-CONT: Note that as this is the local system vs the remote system (ED servers, for journals),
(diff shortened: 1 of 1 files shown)


EDMCLogging.py
```diff
@@ -91,6 +92,9 @@
     **kwargs
 )
 
+# make logging use UTC for times
```
Athanasius, line 95: "... because we want to be able to more easily match our log files with e.g. game journals, which are always in ed-server's idea of UTC"

EDMCLogging.py
```diff
@@ -162,7 +166,10 @@ def __init__(self, logger_name: str, loglevel: int = _default_loglevel):
 
         self.logger_formatter = logging.Formatter('%(asctime)s - %(levelname)s - %(process)d:%(thread)d:%(osthreadid)d %(module)s.%(qualname)s:%(lineno)d: %(message)s')  # noqa: E501
         self.logger_formatter.default_time_format = '%Y-%m-%d %H:%M:%S'
-        self.logger_formatter.default_msec_format = '%s.%03d'
+        # WORKAROUND n/a | 2022-01-20: This is concatted to the above time format, so including %Z there is broken
```
Athanasius, line 169:
```
# As of Python 3.10.2 you can*not* use either `%s.%03.d` in default_time_format *or* use `%Z` in default_time_msec
```

without such a comment one of us/someone else *will* waste time trying it.

Athanasius: I missed a bit:
```
Both cause the logging/strftime code to throw at least one stacktrace per invocation, it's far from silent.
```

Issue #1390, opened 2022-01-07
Ensure we handle `413` from EDDN properly


Athanasius:
It recently came to light that [EDDN has a maximum accepted message size](https://github.com/EDCD/EDDN/issues/158) .  If any message is larger than this then the sender will receive an HTTP `413` status code.

We need to ensure that we handle this sanely.  In particular ensure that such a message (probably a large `navroute`) doesn't cause any subsequent EDDN messages to be blocked.

Athanasius:
Side note:  We don't currently gzip our sent messages, we could start doing that so as to mostly avoid the issue.

Athanasius:
I think we tested the merged branch to my satisfaction, so closing this to clear down 5.3.0 milestone.

PR #1409, opened 2022-01-20
Compress outgoing EDDN data if its large

A-UNDERSCORE-D:
This is a sidestep solution to #1390. It doesn't attempt to directly resend data, only compressing with gzip over a given size. If that STILL returns a 413, its dropped, as without introspection of the message we cannot make it any smaller


util/http.py
```diff
@@ -0,0 +1,24 @@
+"""Utilities for dealing with HTTP."""
+from __future__ import annotations
+
+from gzip import compress
+
+__all__ = ['gzip']
+
+
+def gzip(data: str | bytes, max_size: int = 512, encoding='utf-8') -> tuple[bytes, bool]:
```
Athanasius, line 9: 512 does seem reasonable for this, but I think looking at the typical sizes of EDDN messages might be useful before setting this in stone.

Every message is always going to have:

1. The overall `{...}` enclosure.
2. The `header`, with something like a 20 character uploaderID.
3. The `$schemaRef`
4. A bare minimum of `message:{...}` with at least `timestamp` key:value and at least one other key:value.

So, let's show our working in comments *somewhere*.  512 is probably OK for this util function, but show some working for the EDDN use-case at the call site.

plugins/eddn.py
```diff
@@ -226,15 +237,31 @@ def send(self, cmdr: str, msg: Mapping[str, Any]) -> None:
                 logger.trace_if('plugin.eddn', "EDDN is still objecting to empty commodities data")
                 return  # We want to silence warnings otherwise
 
+            from base64 import b64encode  # we dont need this to be around until this point, which may never hit
+            if r.status_code == 413:
+                logger.debug(dedent(
+                    f"""\
```
Athanasius, line 252: `"""` or `'''` (line 257 below), pick one.  Our standard is currently `'''` by default IIRC.  Although I switch to `"` if it's more convenient due to `'` in the strings.  But at least be consistent within this PR :)

util/http.py
```diff
@@ -0,0 +1,27 @@
...
+    """
```
Athanasius, line 10: Technically this is code that can be used for more than "we're about to sling out a http body".

`util/text.py` ?

plugins/eddn.py
```diff
@@ -213,7 +219,21 @@ def send(self, cmdr: str, msg: Mapping[str, Any]) -> None:
             ('message', msg['message']),
         ])
 
-        r = self.session.post(self.eddn_url, data=json.dumps(to_send), timeout=self.TIMEOUT)
+        # About the smallest request is going to be (newlines added for brevity):
+        # {"$schemaRef":"https://eddn.edcd.io/schemas/shipyard/2","header":{"softwareName":"E:D Market
+        # Connector Windows","softwareVersion":"5.3.0-beta4extra","uploaderID":"abcdefghijklm"},"messa
+        # ge":{"systemName":"delphi","stationName":"The Oracle","marketID":128782803,"timestamp":"xxxx
+        # -xx-xxTxx:xx:xxZ","ships":[]}}
```
Athanasius, line 226: Change this to the actually-validates-currently:

```json
{"$schemaRef":"https://eddn.edcd.io/schemas/commodity/3","header":{"softwareName":"E:D Market Connector Windows","softwareVersion":"5.3.0-beta4extra","uploaderID":"abcdefghijklm"},"message":{"systemName":"delphi","stationName":"The Oracle","marketId":128782803,"timestamp":"2022-01-26T12:00:00Z","commodities":[]}}
```

which is 315 bytes (includes the terminating `\n`) and in a test compressed to 244 bytes, so still a saving, if not the 90% saving seen on large messages.

Please do cite the above paragraph too.

debug_webserver.py
```diff
@@ -30,7 +31,15 @@ def log_message(self, format: str, *args: Any) -> None:
     def do_POST(self) -> None:  # noqa: N802 # I cant change it
         """Handle POST."""
         logger.info(f"Received a POST for {self.path!r}!")
-        data = self.rfile.read(int(self.headers['Content-Length'])).decode('utf-8', errors='replace')
+        data_raw: bytes = self.rfile.read(int(self.headers['Content-Length']))
+        data: str | bytes
+
+        if self.headers.get('Content-Encoding') == 'gzip':
```
Athanasius, line 37: Although we know how the current code is sending, is there any reason to *not* also have `'deflate'` checked here as well?  EDDN Gateway code has:

```python
if content_encoding in ['gzip', 'deflate']:
```

plugins/eddn.py
```diff
@@ -168,6 +170,10 @@ def load_journal_replay(self) -> bool:
 
     def flush(self):
         """Flush the replay file, clearing any data currently there that is not in the replaylog list."""
+        if self.replayfile is None:
```
Athanasius, line 173: This shouldn't be possible, as the only call sites are in EDDN.sendreplay(), which itself has a check for `self.replayfile` (not) being 'truthy'.

But, what I'd like isn't removing this check, but instead logging it at ERROR level, as it triggering would mean we've changed code elsewhere without thinking it through fully.

A-UNDERSCORE-D: Changed. I added this mostly to silence an annoying mypy complaint :D

plugins/eddn.py
A-UNDERSCORE-D: for `'''` `'` wont matter  -- sorry habit from switching between some codebases. Either way, fixed

util/http.py
A-UNDERSCORE-D: moved

util/text.py
```diff
@@ -0,0 +1,27 @@
+"""Utilities for dealing with HTTP."""
```
Athanasius, line 1: Not HTTP specific any more.

A-UNDERSCORE-D: Whoops!

debug_webserver.py
```diff
@@ -1,8 +1,10 @@
 """Simple HTTP listener to be used with debugging various EDMC sends."""
+import gzip
```
Athanasius, line 2: flake8 is catching that `gzip` is now an unused import

plugins/eddn.py
```diff
@@ -226,18 +245,40 @@ def send(self, cmdr: str, msg: Mapping[str, Any]) -> None:
                 logger.trace_if('plugin.eddn', "EDDN is still objecting to empty commodities data")
                 return  # We want to silence warnings otherwise
 
+            if r.status_code == 413:
+                self._log_response(r, header_msg='Got a 413 while POSTing data', sent_data_len=str(len(encoded)))
```
Athanasius, line 249: Can we log at least the `$schemaRef` for the message, and if it's `/journal/` the `event` value ?

debug_webserver.py
```diff
@@ -30,7 +32,20 @@ def log_message(self, format: str, *args: Any) -> None:
     def do_POST(self) -> None:  # noqa: N802 # I cant change it
```
Athanasius, line 32:
```
debug_webserver.py:32:5: CCR001 Cognitive complexity is too high (16 > 15)
1     CCR001 Cognitive complexity is too high (16 > 15)
```

PR #1410, opened 2022-01-22
Retry parsing navroute after failure

A-UNDERSCORE-D:
Navroute parsing can fail due to racy writes if we're too quick. 

This PR solves this by trying on the 10 subsequent events after a NavRoute.

### DIFF
--- a/monitor.py
+++ b/monitor.py
@@ -115,6 +115,7 @@ class EDLogs(FileSystemEventHandler):  # type: ignore # See below
         self.started: Optional[int] = None  # Timestamp of the LoadGame event
 
         self._navroute_retries = 0
+        self._navroute_orig_time: Optional[float] = None
 
         self.__init_state()
 
@@ -505,9 +506,13 @@ class EDLogs(FileSystemEventHandler):  # type: ignore # See below
                 logger.debug(f'Navroute read retry [{self._navroute_retries}]')
                 self._navroute_retries -= 1
                 nv_res = self._parse_navroute_file()
+                if self._navroute_orig_time is None:
+                    logger.critical('Asked to retry for navroute but also no set time to compare? This is a bug.')
+
                 if (
                     nv_res is None
-                    or time() - mktime(strptime(nv_res['timestamp'], '%Y-%m-%dT%H:%M:%SZ')) > MAX_NAVROUTE_DISCREPANCY
+                    or self._navroute_orig_time is None
+                    or self._navroute_orig_time - self._parse_time(nv_res['timestamp']) > MAX_NAVROUTE_DISCREPANCY
                 ):
                     logger.debug(
                         'Failed to parse navroute. ' + 'trying again...' if self._navroute_retries > 0 else 'Giving up'
@@ -520,6 +525,7 @@ class EDLogs(FileSystemEventHandler):  # type: ignore # See below
                     self.state['NavRoute'] = nv_res
                     logger.debug('successfully read navroute file')
                     self._navroute_retries = 0
+                    self._navroute_orig_time = None
 
             event_type = entry['event'].lower()
             if event_type == 'fileheader':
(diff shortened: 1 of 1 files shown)


monitor.py
```diff
@@ -2157,6 +2194,23 @@ def suit_loadout_slots_array_to_dict(self, loadout: dict) -> dict:
...
+    def _parse_navroute_file(self) -> dict[str, Any] | None:
+        """Read and parse NavRoute.json."""
+        if self.currentdir is None:
+            raise ValueError('currentdir unset')
+
+        with open(join(self.currentdir, 'NavRoute.json'), 'r') as f:
+            raw = f.read()
+
+        try:
+            data = json.loads(raw)
```
Athanasius, line 2206: What does this do if you pass it an empty string ?

monitor.py
```diff
@@ -2157,6 +2194,23 @@ def suit_loadout_slots_array_to_dict(self, loadout: dict) -> dict:
 
         return slots
 
+    def _parse_navroute_file(self) -> dict[str, Any] | None:
+        """Read and parse NavRoute.json."""
+        if self.currentdir is None:
+            raise ValueError('currentdir unset')
+
+        with open(join(self.currentdir, 'NavRoute.json'), 'r') as f:
+            raw = f.read()
```
Athanasius, line 2203: If we're being paranoid, then we ought to catch things like FileNotFound in this open().

The function signature very much implies we'll always get `None` back for any error, not an exception thrown (which as far as I can see won't get caught in code anywhere near to the call point).

Athanasius: Sure, there's a top-level try/except in EDLogs.parse_entry(), but that's going to imply the journal entry was bad, not that the additional file caused an issue.

monitor.py
```diff
@@ -1306,16 +1333,26 @@ def parse_entry(self, line: bytes) -> MutableMapping[str, Any]:  # noqa: C901, C
 
             elif event_type == 'navroute':
                 # Added in ED 3.7 - multi-hop route details in NavRoute.json
-                with open(join(self.currentdir, 'NavRoute.json'), 'rb') as rf:  # type: ignore
-                    try:
-                        entry = json.load(rf)
+                if (nv_json := self._parse_navroute_file()) is not None:
+                    entry_time = mktime(strptime(nv_json['timestamp'], '%Y-%m-%dT%H:%M:%SZ'))
```
Athanasius, line 1337: This also needs some paranoia.  Don't just assume `timestamp` is present in the file.  I don't trust Frontier *at all* at this point.  They might give us a file with literally just `{}` in it.

monitor.py
```diff
@@ -1306,16 +1333,26 @@ def parse_entry(self, line: bytes) -> MutableMapping[str, Any]:  # noqa: C901, C
...
+                    c_time = time()
```
Athanasius, line 1338: This check should be versus the `timestamp` in the Journal event that signalled an updated file.

Local time *can* be substantially off from the server time that's used in the Journal files (and I assume also in the additional files).  So this could easily have a false positive when trying to detect "that's an old file".

monitor.py
A-UNDERSCORE-D: I've added an exists check but this can still technically race. Would you rather I wrap it anyway?

Athanasius: I'd wrap it, as that means no race *in that code section*, in terms of "is the file there and we can open it?".  Whether data is yet available to be read is a whole other racey matter.

monitor.py
```diff
@@ -1305,16 +1338,30 @@ def parse_entry(self, line: bytes) -> MutableMapping[str, Any]:  # noqa: C901, C
...
                 # Added in ED 3.7 - multi-hop route details in NavRoute.json
-                with open(join(self.currentdir, 'NavRoute.json'), 'rb') as rf:  # type: ignore
-                    try:
-                        entry = json.load(rf)
+                if (nv_json := self._parse_navroute_file()) is not None and 'timestamp' in nv_json:
+                    file_time = mktime(strptime(nv_json['timestamp'], '%Y-%m-%dT%H:%M:%SZ'))
 
-                    except json.JSONDecodeError:
-                        logger.exception('Failed decoding NavRoute.json', exc_info=True)
+                    if self._navroute_orig_time - file_time > MAX_NAVROUTE_DISCREPANCY:
```
Athanasius, line 1348: For clarity/ease of reading, this could use a comment something like:

```python
# Check if NavRoute.json probably hasn't been updated since any *previous* Journal NavRoute event.
```
It took me a while to think through the possible game order of writing things to conclude what this check is actually trying to achieve.

monitor.py
```diff
@@ -494,6 +502,31 @@ def parse_entry(self, line: bytes) -> MutableMapping[str, Any]:  # noqa: C901, C
             entry: MutableMapping[str, Any] = json.loads(line, object_pairs_hook=OrderedDict)
             entry['timestamp']  # we expect this to exist # TODO: replace with assert? or an if key in check
 
+            if self._navroute_retries > 0:
```
Athanasius, line 505: It feels like this would be cleaner if brought out into its own function.

monitor.py
```diff
@@ -1305,16 +1338,30 @@ def parse_entry(self, line: bytes) -> MutableMapping[str, Any]:  # noqa: C901, C
...
+                        # TODO: rather update entry here, overwriting existing fields but leaving others?
+
+                        # everything is happy, dont retry, dont set the original time, proceed as expected.
+                        entry = nv_json
                         self.state['NavRoute'] = entry
+                        self._navroute_retries = 0
+                        self._navroute_orig_time = None
+
+                else:
+                    logger.info('Failed to decode NavRoute.json. Trying again')
```
Athanasius, line 1364: s/Trying again/Will try again on next Journal event of any type/

monitor.py
```diff
@@ -2157,6 +2165,75 @@ def suit_loadout_slots_array_to_dict(self, loadout: dict) -> dict:
 
         return slots
 
+    def _parse_navroute_file(self) -> Optional[dict[str, Any]]:
+        """Read and parse NavRoute.json."""
+        if self.currentdir is None:
+            raise ValueError('currentdir unset')
+
+        if not (pathlib.Path(self.currentdir) / 'NavRoute.json').exists():
```
Athanasius, line 2173: This case should be caught by the try/except in the open.  So this can be removed, right ?

monitor.py
```diff
@@ -23,8 +23,11 @@
 from edmc_data import edmc_suit_shortnames, edmc_suit_symbol_localised
 from EDMCLogging import get_main_logger
 
+# spell-checker: words navroute
+
 logger = get_main_logger()
 STARTUP = 'journal.startup'
+MAX_NAVROUTE_DISCREPANCY = 5
```
Athanasius, line 30: `# Timestamp difference in seconds`

monitor.py
```diff
@@ -111,6 +114,9 @@ def __init__(self) -> None:
         self.systempopulation: Optional[int] = None
         self.started: Optional[int] = None  # Timestamp of the LoadGame event
 
+        self._navroute_retries = 0
```
Athanasius, line 117: `# Retries currently remaining`

This calls out that's not the 'opposite' of "retries already attempted".

monitor.py
```diff
@@ -111,6 +114,9 @@ def __init__(self) -> None:
...
+        self._navroute_orig_time: Optional[float] = None
```
Athanasius, line 118: I was going to make another "add this comment" comment, but instead I think the variable name should be changed to `self._navroute_journal_timestamp`.

monitor.py
```diff
@@ -2157,6 +2165,75 @@ def suit_loadout_slots_array_to_dict(self, loadout: dict) -> dict:
...
+            logger.exception('Failed to decode NavRoute.json', exc_info=True)
+            return None
+
+        if 'timestamp' not in data:  # quick sanity check
+            return None
+
+        return data
+
+    @staticmethod
+    def _parse_time(source: str) -> float:
```
Athanasius, line 2198: Call this `_parse_journal_timestamp`.

Athanasius (changes requested):
I was going to just do things like add comments myself after merge, but then hit that "do we really need the .exists() now?" so a full review ensued.

Athanasius (approved):

PR #1411, opened 2022-01-24
Ensure that platform guards are recognised by mypy

A-UNDERSCORE-D:
Mypy can and will understand platform guards if they are in the format `sys.platform $comparison 'name'`.
This unfortunately does *NOT* work if platform is assigned to a var (such as when `from sys import platform` is used)

This PR switches all of our `from sys import platform` checks to the format mypy recognizes. It also moves config to its own module and splits the various implementations.


.mypy.ini
```diff
@@ -2,3 +2,4 @@
 follow_imports = skip
 ignore_missing_imports = True
 scripts_are_modules = True
+; platform = darwin
```
Athanasius, line 5: Leftover from testing ?

pyproject.toml
```diff
@@ -3,10 +3,13 @@ max_line_length = 120
...
 
 [tool.pytest.ini_options]
 testpaths = ["tests"] # Search for tests in tests/
 
 [tool.coverage.run]
-omit = ["venv/*"]  # when running pytest --cov, dont report coverage in venv directories
+omit = ["venv/*"] # when running pytest --cov, dont report coverage in venv directories
+
+[tool.pyright]
+# pythonPlatform = 'Darwin'
```
Athanasius, line 15: Also leftover from testing?  There's some value to keeping it as reference of course.

Athanasius:
There *has* to be some documentation that needs updating to make this clear policy.  Probably `Contributing.md` ?

.mypy.ini
A-UNDERSCORE-D: left there (this and the one below) are commented out; but left there intentionally to allow for testing

Issue #1378, opened 2021-12-18
EDMC is not syncing Inara ranks and rank progress when (re)logging while on foot.


mIRCon:
**Please complete the following information:**
 - Version: 5.3.0-beta2
 - Game Version: Odyssey 
 - OS: Windows 10
 - OS Locale: English
 - If applicable: Browser N/A

**Describe the bug**
EDMC is not syncing Inara ranks and rank progress when (re)logging while on foot.
In order to have ranks updated on Inara you have to enter the ship and then relog the game.
It doesn't matter how many times you relog on foot consecutively, ranks and rank progress won't be synced until you enter the ship and relog the game.

**To Reproduce**
1. Enter the game and do some bounties on foot or exobiology or any other activity which affects ranks and rank progress %.
2. After handing in those activities while still on foot just relog the game.
3. If you use --trace-on plugin.inara.events you will see that no ranks and rank progression is being sent to Inara.
4. Check Inara CMDR page for rank and rank progress change just to see that none has been updated.
5. Enter the ship and relog the game.
6. If you use --trace-on plugin.inara.events you will see that ranks and rank progression are sent to Inara.
7. Check Inara CMDR page for rank and rank progress change just to see that it's now updated properly.

**Expected behavior**
Since journal contains ranks and rank progress values whenever you (re)log the game no matter whether you're on foot or in the ship EDMC should use those values when syncing with Inara.

**Screenshots**
N/A

**Additional context**
I have included EDMarketConnector-debug.log since last night during testing with --trace-on plugin.inara.events and cleared fdevid, apikey and other "personal" data (other than cmdr name that is) along with two journal events when relogging the game while on foot and while in the ship so you can compare those events and see why EDMC is triggering Inara rank and rank progression sync while in the ship and not while on foot relog.
Maybe it would be good to check whether other things are not being synced as well when relogging on foot but are provided with journal entries too. (just as precaution while you're investigating this issue anyway)
[EDMarketConnector-debug.log](https://github.com/EDCD/EDMarketConnector/files/7739374/EDMarketConnector-debug.log)
[Journal on foot relog events.txt](https://github.com/EDCD/EDMarketConnector/files/7739375/Journal.on.foot.relog.events.txt)
[Journal in ship relog events.txt](https://github.com/EDCD/EDMarketConnector/files/7739376/Journal.in.ship.relog.events.txt)

Athanasius:
[...]

PR #1412, opened 2022-01-25
Inara: send setCommanderRankPilot message when on-foot

Athanasius:
Due to the existing code waiting for a `Cargo` event, in order to ensure batching of startup/login events, and there being no such event when logging in on foot, this message was never being sent when (re)logging in on foot.  You *had* to relog whilst in your ship in order to trigger it.

As we now have a nice timed queue for sending events, just queue this up when we see a `Rank` Journal event.  Docs state this is only at startup, so we don't need to worry about it occurring mid-session due to a rank-up (that's a `Progress` event instead) and spamming Inara.

Closing #1378

### DIFF
--- a/plugins/inara.py
+++ b/plugins/inara.py
@@ -76,7 +76,7 @@ class This:
         self.suppress_docked = False  # Skip initial Docked event if started docked
         self.cargo: Optional[List[OrderedDictT[str, Any]]] = None
         self.materials: Optional[List[OrderedDictT[str, Any]]] = None
-        self.lastcredits: int = 0  # Send credit update soon after Startup / new game
+        self.last_credits: int = 0  # Send credit update soon after Startup / new game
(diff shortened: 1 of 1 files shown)


A-UNDERSCORE-D (approved):
A test is likely a good idea, but the code looks fine

Athanasius:
> A test is likely a good idea, but the code looks fine

If by that you mean an actual unit test, I have no idea how we'd add such with the code in its current state.

If you mean trying it out and checking logs, and breakpoint in debugger, I did.  For both logging into ship and on foot.  Both caused an appropriate batch of messages to get sent shortly after login.

A-UNDERSCORE-D:
> If you mean trying it out and checking logs, and breakpoint in debugger, I did. For both logging into ship and on foot. Both caused an appropriate batch of messages to get sent shortly after login.

This was what I meant

PR #1417, opened 2022-01-27
Add .editorconfig and reference in Contributing.md

Athanasius:


.editorconfig
```diff
@@ -0,0 +1,21 @@
...
+# 4-space indents, no TABs
+indent_style = space
+tab_width = 4
+indent_size = tab
+
+# Hard-wrap at 120 columns
+max_line_length = 119
+
+# Unix EOL
+end_of_line = crlf
```
A-UNDERSCORE-D, line 15: Unix EOL should be lf, so either the comment needs changing or the config (preferably for me, the config), or just remove it entirely and let git handle it?

Athanasius: Changed the comment, and I even resisted saying 'hysterical' rather than 'historical'.

A-UNDERSCORE-D (approved):

Issue #1255, opened 2021-08-20
Inara: Review the "At least 5% more" check on credits balance.


Athanasius:
https://forums.frontier.co.uk/threads/elite-dangerous-market-connector-edmc.548869/post-9472569

5% might have sounded like a good idea before so many users had billions of credits.

Start by reviewing if this check is even needed now.  After all, we have a good queue system to not bust through the Inara rate limit.

If we still think it's necessary then make it "At least 5% more *or* at least X credits more".

Athanasius:
1. Noobs will want small deltas to be registered on Inara.
2. High Rollers will want much larger, but still small %age of their current wallet balance amounts registered.
3. We shouldn't register truly puny amounts like "I just sold 10 limpets".

So:

```python
if math.abs(previous_balance - new_balance) > math.floor(previous_balance * THRESHOLD_FRACTION, TRIGGER_ABS_DELTA):
   # Send
```

```
           Prev               New  Fraction    Absolute  Send
         10,000            15,000     5%     10,000,000  Yes (> 5%, although < 10mil)
      1,000,000         1,010,000     5%     10,000,000  No (< 5%, and also < 10mil)
      1,000,000         1,100,000     5%     10,000,000  Yes (> 5%, although < 10mil)
  6,000,000,000     6,009,999,999     5%     10,000,000  No (< 5%, and < 10mil)
  6,000,000,000     6,010,000,000     5%     10,000,000  Yes (< 5%, but >= 10mil)
  6,000,000,000     6,100,000,000     5%     10,000,000  Yes (> 5%, and > 10mil)
100,000,000,000   100,100,000,000     5%     10,000,000  Yes (<5%, but > 10mil)
```

Basically, if the delta is at least 10 million, always send.  Else if the delta is at least 5%, send.  If neither true, don't send.  There are no absolutely perfect, everyone will be happy, values to set, but these seem reasonable.

Athanasius:
NB: `math.floor()` above is a brainfart, we need just `min()`.

Athanasius:
I've asked the forum-OP to test with 5.3.0-beta7

Athanasius:
Aaaaaaand no reply from the forum-OP, so in light of my own testing, closing this.

Athanasius:
And now because #1456 has removed the CAPI-data as a source for this we don't even *have* the check any more.

PR #1419, opened 2022-01-31
Inara: Change the way we determine if a credits delta is interesting

Athanasius:
Now based on the *minimum* of a fractional or absolute change.

Closing #1255

### DIFF
--- a/plugins/inara.py
+++ b/plugins/inara.py
@@ -33,7 +33,10 @@ if TYPE_CHECKING:
 
 _TIMEOUT = 20
 FAKE = ('CQC', 'Training', 'Destination')  # Fake systems that shouldn't be sent to Inara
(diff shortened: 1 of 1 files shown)


Athanasius:
Testing
===

Starting with wallet value of 499_991_195.  `this.last_credits == 0`, so balance sent to Inara as expected.

Transfer 144_000_000 to my FC's wallet, trigger 'Update'.  The delta is >5% (and > 10mil), so balance sent to Inara as expected.

Transfer only 9_990_000 back to my wallet.  Delta is <5%, and < 10mil, so *not* sent to Inara, as expected.

Transfer an additional 20_000 back to my wallet.  Delta from `this.last_credits` is now > 10mil, so balance sent to Inara as expected.

Issue #1393, opened 2022-01-07
EDDN: Some `codexentry` messages are invalid


Athanasius:
The new EDDN logging has brought to light things like:

```
2022-01-07 14:45:20,076 - ERROR - Gateway:233: Failed Validation "[<ValidationError: "'' is too short">]" (729, "E:D Market Connector [Windows]", "5.2.4", "https://eddn.edcd.io/schemas/codexentry/1", "-")
```

Presumably we're allowing through some key where the value is an empty string, whereas the schema thinks there should always be a value of at least one character.  This *might* mean relaxing the EDDN schema, or it might mean we need to do some checks and pre-processing.

Athanasius:
Checking the schema could be any of:

1. System (name) - unlikely.
2. Name - also unlikely I'd hope.
3. Region - probably not?
4. Category - please tell me Frontier set this on them all.
5. SubCategory - perhaps not always applicable?
6. Traits - but I'd hope it would just not populate the array if so.

Looking at my own examples of `codexentry` events all of them have everything but `Traits`, of which I have zero examples.  For all I know it somtimes looks like:

```json
Traits: [
 ''
]
```

or

```json
Traits: [
  'foo',
  ''
]
```

Athanasius:
It *looks* like this might be due to a game bug, sometimes writing `''` as the `System` (name) value.  CometBorne supplied a file of *all* their CodexEntry events, and in it I found over 8000 such instances, the latest being:

```json
{
  "Category": "$Codex_Category_StellarBodies;",
  "Category_Localised": "Astronomical Bodies",
  "EntryID": 1200502,
  "IsNewEntry": true,
  "Name": "$Codex_Ent_Green_Sudarsky_Class_I_Name;",
  "Name_Localised": "Green gas giant",
  "Region": "$Codex_RegionName_34;",
  "Region_Localised": "Sanguineous Rim",
  "SubCategory": "$Codex_SubCategory_Gas_Giants;",
  "SubCategory_Localised": "Gas giant planets",
  "System": "",
  "SystemAddress": 388652648075,
  "event": "CodexEntry",
  "timestamp": "2022-01-03T22:27:36Z"
}
```
That system is [Pheia Aewsy LV-Y d11 (EDSM)](https://www.edsm.net/en/system/id/48121471/name/Pheia+Aewsy+LV-Y+d11).

We probably want to start doing some validation of `CodexEntry` data in the EDDN plugin.

[...]

PR #1426, opened 2022-02-01
eddn: Address EDDN server complains about some codexentry messages

Athanasius:
I still can't explicitly explain what causes these errors. All my example events have all the correct data.  So this turned into tracing the code and adding some more paranoid checks and logging.

Closing #1393

### DIFF
--- a/plugins/eddn.py
+++ b/plugins/eddn.py
@@ -951,6 +951,10 @@ class EDDN:
                 entry['BodyID'] = this.body_id
         #######################################################################
 
+        for k, v in entry.items():
+            if v is None or isinstance(v, str) and v == '':
+                logger.warning(f'post-processing entry contains entry["{k}"] = {v}]')
+
         msg = {
             '$schemaRef': f'https://eddn.edcd.io/schemas/codexentry/1{"/test" if is_beta else ""}',
             'message': entry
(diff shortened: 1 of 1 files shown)


plugins/eddn.py
```diff
@@ -1413,14 +1431,14 @@ def journal_entry(  # noqa: C901, CCR001
         # add mandatory StarSystem, StarPos and SystemAddress properties to Scan events
         if 'StarSystem' not in entry:
             if not system:
-                logger.warning("system is None, can't add StarSystem")
+                logger.warning("system is falsey, can't add StarSystem")
                 return "system is None, can't add StarSystem"
```
A-UNDERSCORE-D, line 1435: `is None` here is a problem as noted in the above line, it may or may not actually be None

plugins/eddn.py
```diff
@@ -1433,7 +1451,7 @@ def journal_entry(  # noqa: C901, CCR001
 
         if 'SystemAddress' not in entry:
             if not this.systemaddress:
-                logger.warning("this.systemaddress is None, can't add SystemAddress")
+                logger.warning("this.systemaddress is falsey, can't add SystemAddress")
                 return "this.systemaddress is None, can't add SystemAddress"
```
A-UNDERSCORE-D, line 1455: None -> falsey

Athanasius:
I'm still pondering tweaking this some to not send any message that's found to have empty strings just before queueing it up.

Ideally this can then set status text to alert the user: `Bad CodexEntry - please alert developer` because I *really* want to know which Journal events are tripping this up.

PR #1459, opened 2022-02-11
Reformat everything with black

A-UNDERSCORE-D:
This isnt a PR I expect to be merged; though it could be neat.

I wanted to get a feeling for what things would look like if we switched to using `black`

All of this code was reformatted using `black --safe`, which means the AST is exactly the same as the original.

### DIFF
--- a/EDMC.py
+++ b/EDMC.py
@@ -20,6 +20,7 @@ os.environ["EDMC_NO_UI"] = "1"
 # See EDMCLogging.py docs.
 # workaround for https://github.com/EDCD/EDMarketConnector/issues/568
 from EDMCLogging import edmclogger, logger, logging
+
 if TYPE_CHECKING:
     from logging import TRACE  # type: ignore # noqa: F401 # needed to make mypy happy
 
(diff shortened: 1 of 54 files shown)


Athanasius:
Looking at just EDMC.py for now:

1. Black's default of 88 columns is causing too much "separate line" for my tastes.  So we'll use `--line-length 120`.
2. Black has a maddening feature, not configurable, of minimising the number of backslashes within a string.  This means it will switch the enclosing from `"` (back) to `'` if the latter would need more blackslashes on in-string `'` versus `"`.  I can live with this, so long as I can convince PyCharm to act the same with its prompts for corrections, or turn such off with sufficient specificity.
3. I do wish we could get it to explicitly "if splitting function args, or list/array/set/dict contents, over more than one line, always put each on its own line" in some cases... but perhaps always fill lines in other cases.  At *least* make it not futz with some things like `argparse` -> `add_argument()` things where single thing per line makes a lot of sense.

Athanasius:
I was running manually with `black --line-length 120 --target-version py310 EDMC.py` for this, and you can see the result in https://github.com/EDCD/EDMarketConnector/tree/enhancement/ath/black-experiment

Rixxan:
Hey there!

While the idea to bring Black coding style to EDMC is a great idea, and one that I strongly support, this topic should be visited in a new PR due to the many changes in the code since this PR was opened.

A-UNDERSCORE-D:
Absolutely, this was a "hey this is what it would look like" branch that athan and I were chatting about at the time

Issue #1479, opened 2022-02-19
Prevent plugins' widgets from causing core code/UI issues


Athanasius:
When we released 5.3.0 it was found that the [Hutton Helper](https://github.com/aarronc/hutton-helper) plugin, if you were using Transparent theme, would cause:

1. Clicking 'OK' on 'Settings' after switching to Transparent theme doesn't close the Settings dialogue
2. If you already had Transparent theme selected, then **NO EDMC UI** would appear on startup.

Obviously we should add some try/except in key places to prevent this.

For case 1:

```python
Exception in Tkinter callback
Traceback (most recent call last):
  File "C:\Program Files (x86)\Python310-32\lib\tkinter\__init__.py", line 1921, in __call__
    return self.func(*args)
  File "C:\Users\Athan\Documents\Devel\EDMarketConnector\prefs.py", line 1266, in apply
    theme.apply(self.parent)
  File "C:\Users\Athan\Documents\Devel\EDMarketConnector\theme.py", line 326, in apply
    self._update_widget(widget)
  File "C:\Users\Athan\Documents\Devel\EDMarketConnector\theme.py", line 307, in _update_widget
    widget.configure(font=self.current['font'])
  File "C:\Users\Athan\AppData\Local\EDMarketConnector\plugins\Hutton-Helper\widgets.py", line 84, in configure
    if 'font' in kwargs:  # capture full font details
AttributeError: module 'collections' has no attribute 'Iterable'
```

That's in `theme.py:_Theme._update_widget()` where it's walking through all the widgets.

NB: This doesn't happen *every* time, which is weird.

Athanasius:
Ah, it not triggering was likely due to the plugin having an auto updater that had brought in fixed code.

OK, PR coming up with a try/except that *might* still leave things in a broken state theme-wise, but at least the UI is still working when it hits this.

Athanasius:
Closing as this was long since merged.

PR #1480, opened 2022-02-19
Attempt to prevent plugin widget issues from bringing down whole UI

Athanasius:
This adds a try/except around a bunch of .configure() calls on widgets. Yes, sure, this could then leave the widget in question in a broken state,but at least our entire UI won't fall flat on its face.

There may well be other places we should add such guards.  We should check for such before merging this.

If only we had plugins running in their own process with IPC for all UI poking, eh ?

Closing #1479

### DIFF
--- a/theme.py
+++ b/theme.py
@@ -14,6 +14,9 @@ from tkinter import ttk
 
 from config import config
 from ttkHyperlinkLabel import HyperlinkLabel
+from EDMCLogging import get_main_logger
+
+logger = get_main_logger()
 
 if __debug__:
     from traceback import print_exc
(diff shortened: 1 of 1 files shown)


theme.py
```diff
@@ -266,50 +269,64 @@ def _update_widget(self, widget):
...
+                    widget.configure(font=self.current['font'])
+
+            elif 'background' in widget.keys() or isinstance(widget, tk.Canvas):
+                # e.g. Frame, Canvas
+                if 'bg' not in attribs:
+                    widget.configure(background=self.current['background'],
+                                     highlightbackground=self.current['disabledforeground'])
+
+        except Exception:
+            logger.exception('Plugin widget issue ?')
```
A-UNDERSCORE-D, line 329:
```suggestion
            logger.exception(f'Plugin widget issue ? {widget=}')
```
May as well log whatever the widget was

A-UNDERSCORE-D (approved):

Issue #1504, opened 2022-03-15
New Odyssey Update 11 Journal file names don't sort the same.


Athanasius:
Whilst we correctly detect **new** Journal files once EDMC is running we have a startup issue if the player is using the Odyssey Update 11 client and still has old-name-style Journals as well.

I think this is due to us naively using `sorted()` on the list of file names.  Those extra `-` (and the `'T'`) separators are causing all the Odyssey Update 11 filenames to sort before the old-name-style ones, so we end up with the latest old-name-style one when doing:

```python
self.logfile = join(self.currentdir, logfiles[-1])
```

PR #1505, opened 2022-03-15
Fix 'newest journal file' determination for Odyssey Update 11

Athanasius:
Using os.path.getctime for the order, not the filename.  This addresses the changed Journal filenames in Odyssey Update 11, whilst keeping compatibility with the Horizons Update 11 old-name-style filenames.

Closing #1504


Athanasius:
Tested against both Odyssey and Horizons Update 11 clients, including an EDMC restart with Odyssey running the whole time.  Correctly picks up the newest file.

A-UNDERSCORE-D (approved):

PR #1558, opened 2022-05-25
Added fssbodysignals submission

spansh:
This has been completely drycoded as I don't have an environment which could run ED:MC at the moment, so this really should only be used as a starting point to start testing and evolving.

There's no guarantee this will even pass a compilation test


Athanasius:
Closing this as I've cloned the PR branch and am tweaking it myself.  See #1562 which upcoming PR will reference.

PR #1565, opened 2022-05-29
eddn: Location augmentations cross-check paranoia

### LINKED ISSUE #1557: EDDN: Add location cross-checks to all relevant messages
It's been observed that sometimes EDMC adds a wrong `StarPos` to an EDDN `scanbarycentre` message.

We can probably add a check to be sure the `ScanBarycentre` message's `StarSystem` and `SystemAddress` match where we think we are, so we only add `StarPos` when in that case.

As for root cause, either a game bug meaning we didn't get the FSDJump/Location event, an EDMC bug meaning we didn't track it correctly, or someone editing their Journal files/copying ones in and confusing the tracking.

Athanasius:
Closes #1557 

Add cross-checks to all the location augmentations.  This should work around any game bugs with not always seeing all of the relevent events.

### DIFF
--- a/plugins/eddn.py
+++ b/plugins/eddn.py
@@ -854,6 +854,12 @@ class EDDN:
         #######################################################################
         # Augmentations
         #######################################################################
+        # In this case should add StarPos, but only if the
+        # SystemAddress of where we think we are matches.
+        if this.systemaddress is None or this.systemaddress != entry['SystemAddress']:
+            logger.warning("SystemAddress isn't current location! Can't add augmentations!")
+            return 'Wrong System! Missed jump ?'
+
         ret = this.eddn.entry_augment_system_data(entry, system_name, system_starpos)
         if isinstance(ret, str):
             return ret
--- a/plugins/eddn.py
+++ b/plugins/eddn.py
@@ -946,7 +946,12 @@ class EDDN:
         #######################################################################
         # Augmentations
         #######################################################################
-        # General 'system' augmentations
+        # In this case should add StarPos, but only if the
+        # SystemAddress of where we think we are matches.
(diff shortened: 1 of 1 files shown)


Athanasius:
Yes, how I added the checks is somewhat anti-DRY, but I wanted to be sure both the check and any comments were 100% explicit and correct per event/schema.

That thing about the journal schema and it having a check for `SystemAddress` when that should always be in the event data?  I'd have to go look back through commit history to see how its original augmentation form got in there.  I'm confident the journal events I checked myself are a correct assertion that it will always be present on the relevent events.

plugins/eddn.py
```diff
@@ -959,15 +976,20 @@ def export_journal_codexentry(  # noqa: CCR001
...
-            # Only set BodyID if journal BodyName matches the Status.json one.
-            # This avoids binary body issues.
-            if this.status_body_name == this.body_name:
-                if this.body_id is not None and isinstance(this.body_id, int):
-                    entry['BodyID'] = this.body_id
-
-                else:
-                    logger.warning(f'this.body_id was not set properly: "{this.body_id}" ({type(this.body_id)})')
+            # In case Frontier add it in
+            if entry.get('BodyName', None) is None:
```
A-UNDERSCORE-D, line 980:
```suggestion
            if 'BodyName' not in entry:
```

A-UNDERSCORE-D (approved):
Only that one thing I noticed

plugins/eddn.py
Athanasius: Changed, and the very similar `'BodyId'` check below this.

Athanasius:
All these code paths just got tested by a fly around some systems, scanning etc.  No code errors, no EDDN issues.

PR #1605, opened 2022-07-02
Move back to latest `setuptools`, with workaround

Athanasius:
Whilst the actual `py2exe` build process now works with this workaround and latest `setuptools`, we're now hitting another issue with `infi.systray`'s use of `pkg_resources`, see https://github.com/py2exe/py2exe/issues/139

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -5,9 +5,7 @@ wheel
 # We can't rely on just picking this up from either the base (not venv),
 # or venv-init-time version.  Specify here so that dependabot will prod us
 # about new versions.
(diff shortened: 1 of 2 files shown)


Athanasius:
As commented on the py2exe issue, we'll need to wait for the next py2exe release now, as patching in the py2exe/hooks.py tweak on GitHub is un-necessary pain, and at this time the older setuptools is working for us.

Athanasius:
As I've tested this as now working I'm going to merge without review, which should also prompt dependabot to start poking us about newer setuptools in the future.

PR #1636, opened 2022-08-10
Handle NavRouteClear

### LINKED ISSUE #1634: NavRoute: Check if new `NavRouteClear` event affects us at all
Update 13 adds a new event `NavRouteClear`, triggered when the player *manually* clears their previously plotted route.

When this fires the `NavRoute.json` file's contents become something like:
```json
{ "timestamp":"2022-08-09T15:55:32Z", "event":"NavRouteClear", "Route":[

 ] }
```
We'll be passing the `NavRouteClear` on to plugins, but retaining the details from the last `NavRoute` in `monitor.state['NavRoute']`.  This is probably how we want it, but ensure this is documented.

Also, what happens if someone manages to plot a route and then clear it quickly enough such that we had not yet successfully read the file ?

Athanasius:
If the `NavRoute.json` contains a `NavClearRoute` event, then simply consider that a success and discontinue loading attempts.

Closes #1634

### DIFF
--- a/PLUGINS.md
+++ b/PLUGINS.md
@@ -698,6 +698,18 @@ New in version 5.1.1:
 `state` now has a `ShipLockerJSON` member containing the un-changed, loaded,
 JSON from the `ShipLockerJSON.json` file.
 
+New in version 5.4.2+:
+
+We now handle the 'Update 13' `NavRouteClear` event by detecting if that's what
+is in the `NavRoute.json` file.  If this is the case then we log that, **but
+do NOT clear `state['NavRoute']`**.  Plugins will get sent the Journal
+`NavRouteClear` event anyway, and there might be some value to them retaining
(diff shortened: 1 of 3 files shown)


Athanasius:
My initial testing on this was done via breakpointing on the `navroute` handling, cancelling the route whilst stopped in that, then continuing.  It definitely handles that case.

I then decided to actually test "Just hammer the route/clear route button in-game", as it's the *same* button you can.  That showed up the 'alert the player for no good reason' issue.

Throughout this I've occasionally been triggering a weird "`entry` in monitor.py code is None" issue, but it's far from consistent, even witth plotting/cancelling rapidly ~10 times.  I'm going to assume that's some heisenbug only applicable when running under PyCharm debugging.

A-UNDERSCORE-D (approved):

PR #1639, opened 2022-08-12
inara: ApproachSettlement: Only set marketID if present in event

Athanasius:
At least 'Ancient Ruins' generate an `ApproachSettlement` event without a `MarketID` in the event.  Not even a key with `null` value.

### DIFF
--- a/plugins/inara.py
+++ b/plugins/inara.py
@@ -673,7 +673,7 @@ def journal_entry(  # noqa: C901, CCR001
                         'starsystemBodyCoords': [entry['Latitude'], entry['Longitude']]
                     }
                     # Not present on, e.g. Ancient Ruins
-                    if market_id := entry.get('MarketID') is not None:
+                    if (market_id := entry.get('MarketID')) is not None:
                         to_send['marketID'] = market_id
 
                     new_add_event('setCommanderTravelLocation', entry['timestamp'], to_send)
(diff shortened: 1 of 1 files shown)


Athanasius:
Tested as working for an 'Ancient Ruins'.  I'm about to go poke a 'Thargoid Site' as well, so will check there are no further issues there.

Athanasius:
No further issues whilst playing around Thargoid stuff yesterday.

plugins/inara.py
```diff
@@ -669,10 +669,13 @@ def journal_entry(  # noqa: C901, CCR001
                     to_send = {
                         'starsystemName':       system,
                         'stationName':          entry['Name'],
-                        'marketID':             entry['MarketID'],
                         'starsystemBodyName':   entry['BodyName'],
                         'starsystemBodyCoords': [entry['Latitude'], entry['Longitude']]
                     }
+                    # Not present on, e.g. Ancient Ruins
+                    if market_id := entry.get('MarketID') is not None:
```
A-UNDERSCORE-D, line 676: This may cause market_id to be a bool?

A-UNDERSCORE-D (approved):

plugins/inara.py
Athanasius: Yes, yes it does.

f/x: Ath goes off grumbling about Python operator precedence

A-UNDERSCORE-D:
LGTM -- can't re review

Athanasius:
> LGTM -- can't re review

Well, that's not very helpful **GitHub**, is it ?

I did also test the paradigm in ipython to be sure ;) .

PR #1653, opened 2022-08-30
EDDN: Implement Journal-based `fcmaterials/1` schema

### LINKED ISSUE #1595: EDDN: Add FCMaterials support
NB: At time of opening this there isn't yet an EDDN schema, not even a work in progress.

1. It is supposed that the EDDN schema will be based on the Journal event/file format, as that's the easiest thing to do.
2. But relying on only the Journal data means a player will need to Disembark, walk to the Bartender, interact with the Bartender.  There *is* data for this in the CAPI `/market` endpoint as well, but the format is different.  EDDN schema README will outline the differences and necessary mutation.
3. As you get the Journal event/file every time you 'open' the Bartender we'll want to do the same `if this.fcmaterials != fcmaterials:` thing to not fire off 'duplicate' data to EDDN, e.g. the same as we already for for commodities.

Athanasius:
The EDDN schema is still a work in progress (see https://github.com/EDCD/EDDN/pull/195 ), unsure if we'll use the same one for both Journal and CAPI sourced data, or separate schemas.

But, so far, this works.

Closes #1595


A-UNDERSCORE-D (dismissed):

Athanasius:
@A-UNDERSCORE-D This is still draft, so I'll no doubt ask you to review again at some point.  It *should* now be 'done', but I'll be doing some more testing before the merge.

A-UNDERSCORE-D:
Sounds good, feel free to dismiss and then poke when you want it merged

Athanasius:
As per https://github.com/EDCD/EDDN/issues/193#issuecomment-1238102109 I gave this all another test just now, all smooth sailing.  This should be ready to merge.  @A-UNDERSCORE-D I did *ask* GitHub to re-request your review, whether it worked ....

A-UNDERSCORE-D (approved):

PR #1659, opened 2022-09-07
Implement a `ttk.Sizegrip` for any-theme window resizing

### LINKED ISSUE #478: v3.4.3.0 selecting dark theme breaks window resizing
After you select the dark theme you no longer can resize your edmc window.

Love the program, keep up the good work!

Athanasius:
Due to how some of the UI is set up this *still* only allows *horizontal* resizing, but that's no change.

Now there's a handle in the bottom-right corner, same row as the status text, which allows for resizing.

This entailed implementing a `ttk.Style` instance to ensure correct colouring in all themes.

**This will need testing on Linux**.

Closes #478

### DIFF
--- a/EDMarketConnector.py
+++ b/EDMarketConnector.py
@@ -551,9 +551,11 @@ class AppWindow(object):
         self.status.grid(row=0, column=0, sticky=tk.W)
 
         # Make the window resizable
-        master.resizable(True, False)
+        # Yes, we need to allow vertical resizing, else theme changing can
+        # cause the botttom of the UI to be irretrievably hidden.
+        master.resizable(True, True)
         ttk_grip = ttk.Sizegrip(self.status_row, name='size_grip')
-        ttk_grip.grid(row=0, column=1, sticky=tk.SE)
+        ttk_grip.grid(row=0, column=1, sticky=tk.E)
         #######################################################################
 
         for child in frame.winfo_children():
--- a/EDMarketConnector.py
+++ b/EDMarketConnector.py
@@ -526,19 +526,35 @@ class AppWindow(object):
                 else:
                     appitem.grid(columnspan=2, sticky=tk.EW)
 
+        #######################################################################
         # LANG: Update button in main window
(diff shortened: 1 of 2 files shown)


A-UNDERSCORE-D (approved):

Athanasius:
Using cygwin X stuff on Windows 10, then `ssh -X <host>` to my home server and running EDMarketConnector.py from this branch, and ... well it starts out OK.  But switching to Default theme and then back to Dark causes the status line to disappear.  It comes back if you switch to Default theme again.

So, yeah, this needs more checking on Linux, including if this is even an issue because of this branch or is a separate bug.

I *hope* this isn't merely because my Journal Directory setting is invalid under this setup (which is providing a status line message at least).

Athanasius:
1. The occasional "where did the status line go?" on switching theme (to either one) happens on my proper Linux/X11 setup as well.
2. The resized-to sizing isn't saved.  Probably need to hook into some (t)tk event to pick up the new size, and possibly adjust the current geometry saving, as I've a feeling that is coded to only remember the position.

A-UNDERSCORE-D:
Broken switch normal -> dark, looks like the height geometry isnt updated and something changes.
"working" switch dark -> normal, extra space at the bottom

After restarting, all behaves fine for both, handle works fine on both

Athanasius:
After finding that the current code *does* enploy `ttk.Style()` on Linux I suspect interaction with that (it loads the 'clam' theme) is probably the culprit.

So:

1. the only way to add this resize appears to be `ttk.SizeGrip()`.
2. With nothing else other than adding that the SizeGrip won't follow theme colouring.
3. It being `ttk` the only way to colour it is to tweak the Style.
4. Tweaking that Style appears to have side effects.

Which means, really, this whole thing needs to wait for, and be dependent on, a complete review and possible rewrite of *all* the styling and theming from first principles.

Athanasius:
Given the theme/style issues around this I'm going to close this PR for now, rather than have it clutter things.

PR #1678, opened 2022-09-30
Improve pytest coverage setup and usage

Athanasius:
Due to that EDDN branch causing me to keep thinking "I should add tests" I decided to look at our coverage.

That lead to fixing things to produce a coverage report that is actually correct on simply running `pytest`, but only if no test fails.

Then it became clear that our `sys.platform` branched code was a problem, and I found the `coverage-conditional-plugin` module to help with that.

It does help in the branch cases, and for other single lines, but currently is of no use for whole files.

This PR should attempt to `#pragma` up as much as possible before merging.

I'm torn on the semantics of the names of the rules versus their usage.  Citing a pragma means that if the test in its definition returns `True` that code section is *ignored* for coverage reporting.  Currently there's `sys-platform-not-win32 = "sys_platform != 'win32'"` which you would put on the conditional **FOR** win32.

We could switch the names, so that pragma rule is named `sys-platform-win32`, but then we might think it does a `sys.platform == 'win32'` test....

So, probably leave it as-is, and document it.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -15,8 +15,10 @@ omit = [ "tests/*", "venv/*", "dist.win32/*" ]
 plugins = [ "coverage_conditional_plugin" ]
 
 [tool.coverage.coverage_conditional_plugin.rules]
-# Yes, the sense of all of these is inverted, because else it ends up
-# inverted at *every* use.
+# NB: The name versus content of all of these are inverted because of the way
+#     they're used.  When a pragma cites one it causes that code block to
+#     **NOT** be considered for code coverage.
+#     See Contributin.md#test-coverage for more details.
 sys-platform-win32 = "sys_platform != 'win32'"
 sys-platform-not-win32 = "sys_platform == 'win32'"
 sys-platform-darwin = "sys_platform != 'darwin'"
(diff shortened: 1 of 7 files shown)


inorton:
Very cool! I might try that coverage plugin for my own stuff!

pyproject.toml
```diff
@@ -7,9 +7,25 @@ line_length       = 119
...
 [tool.coverage.run]
-omit = ["venv/*"] # when running pytest --cov, dont report coverage in venv directories
+omit = [ "tests/*", "venv/*", "dist.win32/*" ]
+plugins = [ "coverage_conditional_plugin" ]
+
+[tool.coverage.coverage_conditional_plugin.rules]
+# NB: The name versus content of all of these are inverted because of the way
+#     they're used.  When a pragma cites one it causes that code block to
+#     **NOT** be considered for code coverage.
+#     See Contributin.md#test-coverage for more details.
```
A-UNDERSCORE-D, line 21:
```suggestion
#     See Contributing.md#test-coverage for more details.
```

A-UNDERSCORE-D (approved):

PR #1703, opened 2022-11-18
Make runnable under 64-bit Python

Athanasius:
Given the `py2exe` deprecation of 32-bit packages I decided it was time to take another stab at getting our .exe files to run under 64-bit Python.

After some fiddling around with ctypes, and re-testing showing we **can** just use `tkinter.filedialog` now (last tested under Python 3.7.x in Oct 2020), the code now all works under 64-bit Python 3.10.8 (and also 3.11.0).

* `Build-exe-and-msi.py` has been adjusted to check for, and require, 64-bit, not 32-bit, Python.
* You need the `x64` version of the `WinSparkle.dll` (~~still need to adjust the github workflow for that~~ done).
* All the files that had `ctypes` in them have had all their code tested as working on 64-bit.
* For that "Where should the Output file location be?" code in `prefs.py`... I really did try to get the old ctypes code working under 64-bit python, but ultimately failed.  Upon re-checking `tkinter.filedialog` **without setting locale to UTF-8** it Just Worked anyway.  Indeed I couldn't reproduce the original error with Python 3.7.9 64-bit either.  It's possible that Windows 10 builds older than mine (21H2 / 19044.2311) might still have an issue with this.  It needs testing.

### DIFF
--- a/.github/workflows/windows-build.yml
+++ b/.github/workflows/windows-build.yml
@@ -34,7 +34,7 @@ jobs:
         run: |
           Invoke-Webrequest -UseBasicParsing https://github.com/vslavik/winsparkle/releases/download/v0.7.0/WinSparkle-0.7.0.zip -OutFile out.zip
           Expand-Archive out.zip
-          Move-Item 'out\WinSparkle-0.7.0\Release\*' '.\'
+          Move-Item 'out\WinSparkle-0.7.0\x64\Release\*' '.\'
 
       - name: Build EDMC
         run: |
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -82,7 +82,7 @@ repos:
     - id: LANG_comments
       name: 'LANG comments'
       language: system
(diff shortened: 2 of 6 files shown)


A-UNDERSCORE-D (approved):

A-UNDERSCORE-D:
Looks good; I dont expect the filedialog issue to crop up much

Athanasius:
With a tweak to the build script to allow *this* branch to build on 32-bit, that works.

1. Builds without issues.
2. EDMarketConnector.exe runs
3. If I had beforehand cleared `fdev_apikeys` from registry then the auth flow works without problems.  It's from the built .exe, so definitely via the `edmc://` method.
    `2022-12-23 15:52:46.155 UTC - DEBUG - 18852:14068:14068 protocol.WindowsProtocolHandler.worker:324: Message starts with edmc://auth`
4. The file dialog instances got changed to just tkinter, so no ctypes there any more.

So, in terms of merging this and continuing to build 32-bit for full releases I don't think there are any blockers.

PR #1716, opened 2022-11-27
inara: Send only Live galaxy data after Update 14

### LINKED ISSUE #1715: Inara: From Update 14 onwards send only data from Live galaxy
Ref: https://inara.cz/elite/board-thread/7049/463292/#463292

> For all app authors utilizing Inara API: As Inara will be supporting only Live data (so Odyssey and Horizons 4.0), there is recommended for any app using Inara API to update commander's data to send just Live data via API. You can simply implement a killswitch based on the journal's 'gameversion' property (or any other relevant property that may appear in the journal with the U14), to ensure just Live data will be sent.
> 
> On the game update 14 release day, I will disable all Inara API updates temporarily and will put it back on once there will be clear everything works as expected. Also, when you implement this data split in your apps, please let me know your app version containing the changes, so I can enforce that version for all API updates. This is relevant only for the apps updating CMDR's data, of course (other apps don't need to take any actions). 

So we need to add a check, and abort all Inara data processing if it's not the Live galaxy.

Given we *should* have 5.6.0 out *before* Update 14 we probably want to put a date/time component into the check.  I'd suggest (pseudo-code):
```
if datetime > '2022-11-29T09:00:00' and gameversion < 4.0:
	log('tell the user this is disabled')
    status('also tell the user here')
	return
```

**WHEN THIS IS RELEASED be sure to tell Artie which version is the minimum that is 'good' for this.**

Athanasius:
Closes #1715

### DIFF
--- a/plugins/inara.py
+++ b/plugins/inara.py
@@ -28,6 +28,7 @@ import time
 import tkinter as tk
 from collections import OrderedDict, defaultdict, deque
 from dataclasses import dataclass
+from datetime import datetime, timedelta, timezone
 from operator import itemgetter
 from threading import Lock, Thread
 from typing import TYPE_CHECKING, Any, Callable, Deque, Dict, List, Mapping, NamedTuple, Optional
@@ -44,6 +45,7 @@ import timeout_session
 from companion import CAPIData
 from config import applongname, appversion, config, debug_senders
 from EDMCLogging import get_main_logger
+from monitor import monitor
 from ttkHyperlinkLabel import HyperlinkLabel
 
 logger = get_main_logger()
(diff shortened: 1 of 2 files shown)


plugins/inara.py
```diff
@@ -361,6 +369,24 @@ def journal_entry(  # noqa: C901, CCR001
 
     :return: str - empty if no error, else error string.
     """
+    if not monitor.is_live_galaxy():
+        # This only applies after Update 14, which as of 2022-11-27 is scheduled
+        # for 2022-11-29, with the game servers presumably being down around
+        # 09:00
+        if datetime.now(timezone.utc) >= datetime.fromisoformat("2022-11-27T09:00:00+00:00"):
```
A-UNDERSCORE-D, line 376: I would move this into the other side, that way it's more generally useful

Athanasius: I actually moved it *out* of that side so as to keep the semantics of `monitor.is_live_galaxy()` cleaner.

Long-term it'll actually get removed.  We just need it for this release due to making it before Update 14 and needing it to only have effect *after*.

This is where we're thankful that we only (normally) process live, new, Journal files.  Anyone playing silly buggers with moving files around gets to keep the broken pieces.

A-UNDERSCORE-D (approved):

plugins/inara.py
A-UNDERSCORE-D: Works for me

Issue #1723, opened 2022-11-28
Get *all* python files to pass flake8 checks


Athanasius:
See https://github.com/EDCD/EDMarketConnector/pull/1712#issuecomment-1329474543 for why we need this to be the case.

Currently some lines in a *lot* of files don't pass, because they've never been part of a diff that was checked by flake8 (or have become non-passing due to flake8 changes since they were last 'seen').

Athanasius:
Reference for alternate run methods: https://github.com/PyCQA/flake8/issues/1760

Athanasius:
For reference, the flake8 changelogs seem to only be at https://flake8.pycqa.org/en/latest/release-notes/index.html

You have to spot the link to that site in the top-right of the github repo.

PR #1740, opened 2022-12-03
Get everything passing flake8 checks

Athanasius:
Because of the change in flake8 6.0.0 (`--diff` no longer exists, and it turns out it was always unreliable, hence it being removed), we need to move to just running it on whole files.

But some of the details about how to achieve that are ... awkward.   So, as per #1723 we need to get all of our code passing flake8 tests such that we can *then* adjust the github workflows, and pre-commit config, to always run on whole files.

**THEN** we can actuall upgrade to flake8 6.0.0+.

Now complete
---
I did endeavour to check code was still working as I went, file by file, so this *should* be good to merge.

### DIFF
--- a/loadout.py
+++ b/loadout.py
@@ -49,19 +49,11 @@ def export(data: companion.CAPIData, requested_filename: Optional[str] = None) -
     query_time = config.get_int('querytime', default=int(time.time()))
 
     # Write
-    #
-    #  When this is refactored into multi-line CHECK IT WORKS, avoiding the
-    #  brainfart we had with dangling commas in commodity.py:export() !!!
-    #
-    output_path = pathlib.Path(config.get_str('outdir'))
-    output_filename = pathlib.Path(
-        ship + '.' + time.strftime('%Y-%m-%dT%H.%M.%S', time.localtime(query_time)) + '.txt'
-    )
-    # Can't re-use `filename`, different type
-    file_name = output_path / output_filename
-    #
-    #  When this is refactored into multi-line CHECK IT WORKS, avoiding the
-    #  brainfart we had with dangling commas in commodity.py:export() !!!
-    #
-    with open(file_name, 'wt') as h:
+
+    with open(
+        pathlib.Path(config.get_str('outdir')) / pathlib.Path(
+            ship + '.' + time.strftime('%Y-%m-%dT%H.%M.%S', time.localtime(query_time)) + '.txt'
+        ),
+        'wt'
+    ) as h:
         h.write(string)
--- a/loadout.py
+++ b/loadout.py
@@ -1,48 +1,67 @@
-# Export ship loadout in Companion API json format
+"""Export ship loadout in Companion API json format."""
 
 import json
 import os
-from os.path import join
+import pathlib
 import re
 import time
+from os.path import join
+from typing import Optional
 
-from config import config
 import companion
 import util_ships
+from config import config
+from EDMCLogging import get_main_logger
+
+logger = get_main_logger()
 
 
-def export(data, filename=None):
+def export(data: companion.CAPIData, requested_filename: Optional[str] = None) -> None:
+    """
+    Write Ship Loadout in Companion API JSON format.
 
+    :param data: CAPI data containing ship loadout.
+    :param requested_filename: Name of file to write to.
+    """
     string = json.dumps(
         companion.ship(data), cls=companion.CAPIDataEncoder,
         ensure_ascii=False, indent=2, sort_keys=True, separators=(',', ': ')
     )  # pretty print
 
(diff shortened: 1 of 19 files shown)


Athanasius:
Still remaining:

- update.py
- util_ships.py
- plugins/eddb.py
- plugins/edsy.py
- scripts/killswitch_test.py
- scripts/pip_rev_deps.py

Athanasius:
Every file that a locale bare `flake8` pass checks is now clean in both flake8 and mypy.

If anything ever fails in a PR that wasn't locally, then that means `.flake8` `exclude=` likely needs adjusting.

loadout.py
```diff
@@ -1,48 +1,67 @@
...
     #
     #  When this is refactored into multi-line CHECK IT WORKS, avoiding the
     #  brainfart we had with dangling commas in commodity.py:export() !!!
     #
-    filename = join(config.get_str('outdir'), '%s.%s.txt' % (ship, time.strftime('%Y-%m-%dT%H.%M.%S', time.localtime(querytime))))
+    output_path = pathlib.Path(config.get_str('outdir'))
+    output_filename = pathlib.Path(
+        ship + '.' + time.strftime('%Y-%m-%dT%H.%M.%S', time.localtime(query_time)) + '.txt'
+    )
+    # Can't re-use `filename`, different type
```
A-UNDERSCORE-D, line 60: Still relevant?

theme.py
```diff
@@ -390,16 +410,16 @@ def apply(self, root):
...
-            GetWindowLongW = ctypes.windll.user32.GetWindowLongW
-            SetWindowLongW = ctypes.windll.user32.SetWindowLongW
+            GWL_EXSTYLE = -20  # noqa: N806 # ctypes
+            WS_EX_APPWINDOW = 0x00040000  # noqa: N806 # ctypes
+            WS_EX_LAYERED = 0x00080000  # noqa: N806 # ctypes
+            GetWindowLongW = ctypes.windll.user32.GetWindowLongW  # noqa: N806 # ctypes
+            SetWindowLongW = ctypes.windll.user32.SetWindowLongW  # noqa: N806 # ctypes
 
-            root.overrideredirect(theme and 1 or 0)
+            root.overrideredirect(theme and True or False)
```
A-UNDERSCORE-D, line 422:
```suggestion
            root.overrideredirect(bool(theme))
```

ttkHyperlinkLabel.py
```diff
@@ -1,26 +1,40 @@
...
 
 
-class HyperlinkLabel(sys.platform == 'darwin' and tk.Label or ttk.Label, object):
+# FIXME: Split this into multi-file module to separate the platforms
+class HyperlinkLabel(sys.platform == 'darwin' and tk.Label or ttk.Label, object):  # type: ignore
+    """Clickable label for HTTP links."""
 
-    def __init__(self, master=None, **kw):
+    # NB: do **NOT** try `**kw: Dict`, it causes more trouble than it's worth.
+    def __init__(self, master: Optional[tk.Tk] = None, **kw) -> None:
```
A-UNDERSCORE-D, line 37: kw: Any?

loadout.py
Athanasius: Oh, that was to *point out* why I was using "another temporary variable".  If not for type checking it would.  Might need a small update as I seem to have changed the original name too.

theme.py
Athanasius: Possibly.  Mostly I want to entirely change the "theme is an int with magic meanings for the values" at some point.

ttkHyperlinkLabel.py
Athanasius: That I did not try, but I'll be unsurprised if mypy still insists that then doesn't match with what typeshed thinks is going on with actual uses.  At that stage it might as well stay untyped ?

The 'make mypy happy' throughout this was mostly get satisfy the flake8 "must have at least 75% (default) types coverage".

loadout.py
Athanasius: Now I look at the code again, I've condensed all of that down into the `with open(...)`, no intermediate variables.  Tested as still working.

theme.py
Athanasius: I'm definitely going to follow up merging this with changing all the `0`, `1`, `2` for theme selection to using defined constants, and *not* doing any clever "oh, default is `0`, so let's treat this like a boolean" bullshit.

ttkHyperlinkLabel.py
Athanasius: I ought not respond to PR comments just before bed.

Yup, `Any` works, and at least makes it explicit.  Also used in the `__setitem__()` signature now.

PR #1744, opened 2022-12-05
eddn: Ensure CAPI-sourced messages have uploaderID set

Athanasius:
If no 'live' Journal events yet, plugins/eddn won't have had an event to set this from.


Athanasius:
This change is necessary because `EDDN.standard_header()` gets used and *that* assumes that `this.cmdr_name` has been set.

Issue #1714, opened 2022-11-26
Add killswitch functionality to all CAPI code


Athanasius:
We could really have made use of this for forthcoming Update 14, but unfortunately it doesn't exist.

- [x] Allow for completely turning off any CAPI requests.
- [x] Allow for turning off use of each CAPI endpoint individually.
- [x] Allow for turning off only export of any CAPI data.  This would be used to still allow to pull data for "File" > "Status" and plugin use, but would turn off its use for anything else, i.e. EDDN messages (which themselves largely can't tell if the data is CAPI or Journal sourced, i.e. EDDN commodity, shipyard and outfitting messages).

Athanasius:
One thing to note, if you use `capi.request./profile` then it'll disable the entire "usual set of queries", as it bails from `companion.py:Session.capi_station_queries()`.  Likewise if you kill `/market` then it won't get to trying the `/shipyard` query.

I'll see if I can change this, but I've a feeling it might entail some tortured logic.  If that's the case then I'd rather not.

PR #1759, opened 2022-12-16
companion: A start on adding CAPI killswitches

Athanasius:
Implement killswitch functionality for CAPI data retrieval and use as per #1714

### DIFF
--- a/companion.py
+++ b/companion.py
@@ -652,9 +652,10 @@ class Session(object):
     def start_frontier_auth(self, access_token: str) -> None:
         """Start an oAuth2 session."""
         logger.debug('Starting session')
-        self.requests_session = requests.Session()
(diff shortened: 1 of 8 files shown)


companion.py
```diff
@@ -593,7 +601,7 @@ def __init__(
     ):
         super().__init__(query_time=query_time, play_sound=play_sound, auto_update=auto_update)
         self.message: str = message  # User-friendly reason for failure.
-        self.exception: int = exception  # Exception that recipient should raise.
+        self.exception: BaseException = exception  # Exception that recipient should raise.
```
A-UNDERSCORE-D, line 604: Use Exception here rather than BaseException if possible. BaseException implies things like SystemExit

Athanasius: Yeah, I'll change that.  Blame the 'hint' from mypy.

Issue #847, opened 2021-01-15
Add "/fleetcarrier" cAPI endpoint data to EDMC plugin's API


norohind:
**Describe the enhancement**

Add information from https://companion.orerve.net/fleetcarrier to EDMC plugin's API as it done with "/profile" endpoint and "cmdr_data" function. 

**Expected behavior**

Call function with response of endpoint and beta state as it done with "cmdr_data" function.

**Considered workarounds**

It is possible to get current fleet carrier location by checking where your ships stored (by data from /profile endpoint). But it requires to store on carrier at least 1 ship. Also it doesn't provide any other information about carrier, just location.

**Additional context**
I can't imagine special conditions for automatically send request to this endpoint so i think there should be manual way to fetch data from this endpoint.

Athanasius:
Some notes for implementation:

1. We don't want to just sling out `/fleetcarrier` requests for every user.  The vast majority likely won't own an FC.
    1. There doesn't seem to be any indication in the Journal that the Cmdr owns an FC until one of the relevant events occurs, so we can't tell on Cmdr login and set a flag.
    2. Events to know we should start querying CAPI `/fleetcarrier` for the current Cmdr are going to be *any* of the events in section 11 of v28 Journal docs.
    3. We'll need to have a flag (might just be able to use a stored FC callsign) that we set when we first detect the Cmdr has an FC, and unset/set it False when we confirm a Decommission.  NB: There's no event for a Decommission completing.  This will need inferring from the CarrierDecommission event timestamp, plus the known time it takes, and/or when `/fleetcarrier` no longer returns valid data.
2. When would we actually fire off `/fleetcarrier` ?
    1. We'll want to do so after seeing relevant FC Journal events. However this probably needs to go through a 'queue' system so that we do *not* request `/fleetcarrier` for every event, which could mean several a minute.  So start a timer when we see one (and update this for each new event), when that expires, send the request.  Some care will be required in selecting this timer length to avoid the request *never* being sent for a 'fussy' Cmdr.
    2. In addition to this we'll want to periodically poll as things like the current status of Cargo can change without the Cmdr themselves causing the change.

Athanasius:
Re: point 1 - actually we can afford to sling out a single `/fleetcarrier` after initial authorization and then (re)set the flag based on that.  We can then later change it if we detect an FC being bought.

Athanasius:
[...]

PR #1773, opened 2022-12-22
Feature/847/capi fleetcarrier

aussig:
This PR implements #847 - calling of the CAPI endpoint for fleetcarrier data. 

This call is only triggered on `CarrierBuy`, `CarrierStats` and `CarrierTradeOrder` journal events, and is on a separate cooldown (currently 15 mins) to the main CAPI call.

The new plugin callback function is `capi_fleetcarrier()`.


plug.py
```diff
@@ -401,6 +401,32 @@ def notify_capidata(
     return error
 
 
+def notify_capi_fleetcarrierdata(
```
Athanasius, line 404: I don't think `is_beta` is actually necessary in this case.  The `.source_host` should denote it (because if there's ever a beta of the game it'll be `pts-companion.orerve.net`), and if somehow the beta detection has failed and the live host was used anyway, then the data will be for the live game, not beta.

EDMarketConnector.py
```diff
@@ -1489,6 +1565,11 @@ def crewroletext(role: str) -> str:
                 if not should_return:
                     self.w.after(int(SERVER_RETRY * 1000), self.capi_request_data)
 
+            if entry['event'] in ['CarrierBuy', 'CarrierStats', 'CarrierTradeOrder']:  # 'CargoTransfer' too?
```
Athanasius, line 1568: Re: `CargoTransfer` ... I'm conflicted on this because whilst it's a pro to trigger the query after cargo hold changes ... doing so on the first of possibly many would 'jump the gun' and only updated for the first... and that's assuming the CAPI data updates in a timely manner.

The latter also applies to `CarrierTradeOrder`.  I'm leaning towards requiring the player to relogin in the game (to get `CarrierStats`) as the simplest way to give clear player control for this, without complicating things with an extra UI button/config options.

PLUGINS.md
```diff
@@ -940,19 +940,52 @@ def cmdr_data(data, is_beta):
 | :-------- | :--------------: | :------------------------------------------------------------------------------------------------------- |
 | `data`    |     `CAPIData`   | `/profile` API response, with `/market` and `/shipyard` added under the keys `marketdata` and `shipdata` |
 | `is_beta` |      `bool`      | If the game is currently in beta                                                                         |
+
+If a plugin has a `capi_fleetcarrier()` function it gets called when the application has just fetched fresh Fleetcarrier data from Frontier's CAPI servers. This is done when `CarrierBuy`, `CarrierStats` or `CarrierTradeOrder` events are detected in the Player Journal. To avoid flooding Frontier's CAPI server, a throttle is applied to ensure a significant interval between requests (currently 15 mins).
+
```
Athanasius, line 945: Add a note here about how a single query can take a long time under some circumstances.  This means not only could a fleetcarrier update take a long time, *but it will also possibly delay all other CAPI queries*.

companion.py
```diff
@@ -952,6 +955,9 @@ def capi_station_queries(  # noqa: CCR001
                 if query.endpoint == self._CAPI_PATH_STATION:
                     capi_data = capi_station_queries(query.capi_host)
 
+                elif query.endpoint == self.FRONTIER_CAPI_PATH_FLEETCARRIER:
+                    capi_data = capi_single_query(query.capi_host, self.FRONTIER_CAPI_PATH_FLEETCARRIER)
```
Athanasius, line 959: This needs to specify a custom `timeout=...`, because the default 10s isn't going to work for the slow queries.

The query is threaded, so won't hold up the EDMC UI, but can hold up other CAPI queries, so I'm inclined to try this with 60s for now.

Athanasius: Long-term we might need to adjust `capi_query_worker()` to actually perform *these* (or future other 'not station' query set) queries in their own sub-thread.  But that can wait until we see how *this* shakes out in practice.

PLUGINS.md
Athanasius: Also, now I realise we might have lost some "don't allow 'Update' if already in progress" check when I made CAPI threaded.  But don't worry about that for this PR.

Athanasius (changes requested):
Mostly minor changes needed, but some further discussion about some things required.

EDMarketConnector.py
```diff
@@ -1082,6 +1084,56 @@ def capi_request_data(self, event=None) -> None:  # noqa: CCR001
...
+    def capi_request_fleetcarrier_data(self, event=None) -> None:
+        """
+        Perform CAPI fleetcarrier data retrieval and associated actions.
+
+        This is triggered by certain FleetCarrier journal events
+
+        :param event: Tk generated event details.
+        """
+        logger.trace_if('capi.worker', 'Begin')
+        should_return, new_data = killswitch.check_killswitch('capi.fleetcarrier', {})
```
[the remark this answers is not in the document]
A-UNDERSCORE-D: additionally if you're not going to pass data to be modified, you should use `_` for `new_data` to make that visible

EDMarketConnector.py
A-UNDERSCORE-D, line 1568:
```suggestion
            if entry['event'] in ('CarrierBuy', 'CarrierStats', 'CarrierTradeOrder'):  # 'CargoTransfer' too?
```

While its not likely to be changed / changable, good practice to use tuples here

A-UNDERSCORE-D (changes requested):

EDMarketConnector.py
Athanasius: Re: using `_` ... to be fair, that wasn't caught when *I* recently added the CAPI killswitches.  It'll need changing in general for those.

plug.py
```diff
@@ -401,6 +401,32 @@ def notify_capidata(
...
+    :param is_beta: whether the player is in a Beta universe.
+    :returns: Error message from the first plugin that returns one (if any)
+    """
+    error = None
+    for plugin in PLUGINS:
+        fc_data = plugin._get_func('capi_fleetcarrier')
+
+        if fc_data:
+            try:
+                newerror = fc_data(data, is_beta)
```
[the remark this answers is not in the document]
Athanasius: Yes, they're both still things, and we ought to put that `callable(...)` check on all the plugin calling code in plug.py.

FYI @aussig the deep copy is so that one plugin mutating the data doesn't affect a subsequent one.

EDMarketConnector.py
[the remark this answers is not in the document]
aussig: I'll implement both changes.

EDMarketConnector.py
```diff
@@ -1082,6 +1084,56 @@ def capi_request_data(self, event=None) -> None:  # noqa: CCR001
...
+            hotkeymgr.play_bad()
+            return
+
+        if not monitor.cmdr:
+            logger.trace_if('capi.worker', 'Aborting Query: Cmdr unknown')
+            # LANG: CAPI queries aborted because Cmdr name is unknown
+            self.status['text'] = _('CAPI query aborted: Cmdr name unknown')
+            return
+
+        if not monitor.mode:
```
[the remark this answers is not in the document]
aussig: I originally excluded the CQC and Multicrew checks as I agree they are irrelevant, but thought a check for a valid commander might be a good idea. Will remove that too though.

EDMarketConnector.py
```diff
@@ -1082,6 +1084,56 @@ def capi_request_data(self, event=None) -> None:  # noqa: CCR001
...
+
+        if monitor.state['GameVersion'] is None:
+            logger.trace_if('capi.worker', 'Aborting Query: GameVersion unknown')
+            # LANG: CAPI queries aborted because GameVersion unknown
+            self.status['text'] = _('CAPI query aborted: GameVersion unknown')
+            return
+
+        if not companion.session.retrying:
+            if time() < self.capi_fleetcarrier_query_holdoff_time:  # Was invoked while in cooldown
+                return
```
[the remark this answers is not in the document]
aussig: I'll implement.

EDMarketConnector.py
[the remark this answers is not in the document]
Athanasius: It's the `monitor.mode` check I was referring to.

You need the `monitor.cmdr` check, else it's going to fail anyway when it can't determine which Cmdr's credentials to use for the query.

EDMarketConnector.py
aussig: Ok, will remove that comment and also remove `CarrierTradeOrder` as a trigger.

plug.py
aussig: Yeah, was following the pattern of the other plugin callbacks, but agree everything is in the `source_host`. Will change.

PLUGINS.md
aussig: Will add some additional text.

EDMarketConnector.py
aussig: Want me to do that @Athanasius ?

Athanasius: Yes, it's the more correct form.  In general if A_D requests something and I haven't objected, take it as canon.  They know more Python than I do ;) .

plug.py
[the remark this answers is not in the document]
aussig: Will implement both.

EDMarketConnector.py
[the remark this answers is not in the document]
aussig: Ah ok, that makes more sense.

EDMarketConnector.py
aussig: Ok, tried this and it immediately throws an exception:

```
self.ship_label['text'] = _('Ship') + ':'  # LANG: 'Ship' label in main UI
                              ^
UnboundLocalError: cannot access local variable '_' where it is not associated with a value
```

This exception isn't thrown anywhere near any of the instances of 
```python
should_return, new_data = killswitch.check_killswitch('capi.request.fleetcarrier', {})
```
that I changed to:
```python
should_return, _= killswitch.check_killswitch('capi.request.fleetcarrier', {})
```

So, I suspect this is because `_()` is defined by i18n [as described here](https://stackoverflow.com/questions/5893163/what-is-the-purpose-of-the-single-underscore-variable-in-python/5893946#5893946)

That article suggests using double underscore `__` as an alternative for throwaway variables when used in conjunction with l18n.  @A-UNDERSCORE-D would that be acceptable?

Athanasius: Yup, if `__` works, then we'll have to use that (and I'm surprised where I changed it apparently hasn't blown up yet in this branch, but I've only just gotten to test-running it).

aussig: Ok, will stick with that.

aussig:
Thanks all, I've implemented the changes from the first round of reviews.

Athanasius:
I think everything in my Review has been address, however...

I just had a lovely edge-case thought:

1. Player with two accounts, each has an FC.
2. Login into Cmdr A, `/fleetcarrier` query issued, but it's a slow one.
3. Meantime, re-logs to Cmdr B.
4. Response from #2 arrives.

Now, the way it's currently architected `capi_fleetcarrier()` doesn't make any guarantees about the data being for "the current Cmdr", so technically this is OK.  But plugins would have to be aware of this and cater for it.

But, perhaps, we should record the `monitor.cmdr` at query time (I'd expand EDMCCAPIRequest to have this as a constructor arg and store it, to then be mirrored in the EDMCCAPIRespons), such that it's an extra property in CAPIData when passed to plugins.

aussig:
> But, perhaps, we should record the `monitor.cmdr` at query time (I'd expand EDMCCAPIRequest to have this as a constructor arg and store it, to then be mirrored in the EDMCCAPIRespons), such that it's an extra property in CAPIData when passed to plugins.

Ok, I'll try to get this done today, though it may have to wait to after the holidays as it's about to get very busy here!

Athanasius:
No rush, we're not planning on 5.8.0 any time soon.  And there's a bunch of 'small' issues in the milestone that I want to try and get cleared for it.

But, I'll probably end up merging my "get *everything* passing mypy checks" PR before you're done now, and then you'll definitely want to rebase, especially given you've applied changes in files irrelevant to your actual work (the plugins).

Athanasius (approved):

A-UNDERSCORE-D (approved):

Athanasius:
> If I inject this into the `CAPIData` itself, should I be wary of the property name? For example, `cmdr` might appear in some response data already or in the future, perhaps use something more distinct like `original_request_cmdr`?
> 
> ```python
> capi_data['original_request_cmdr'] = capi_response.cmdr
> ```
> 
> Or alternatively we could pass it as a separate parameter to the plugin callback function, keeping it out of the `CAPIData`?

I prefer having it in CAPIData.  We originally made that for slightly better type enforcing, but I think it's useful to have these additional properties too.

aussig:
Ok, will do. Shall I use the more lengthy `original_request_cmdr` or happy with `cmdr_name` perhaps?

Athanasius:
> Ok, will do. Shall I use the more lengthy `original_request_cmdr` or happy with `cmdr_name` perhaps?

Let's split the difference with `request_cmdr`.

aussig:
Oh FFS sorry, I had 'show whitespace' temporarily switched off in my diff viewer, so I didn't see a bunch of whitespace changes in `PLUGINS.py` that the editor applied automatically. 

I'll fix this in another commit.

aussig:
Ok, stupid whitespace chnages reverted in documentation.

The code changes to pass `request_cmdr` in `CAPIData` are ready for review again now.

companion.py
```diff
@@ -1070,7 +1076,8 @@ def fleetcarrier(
                 tk_response_event=tk_response_event,
                 query_time=query_time,
                 play_sound=play_sound,
-                auto_update=auto_update
+                auto_update=auto_update,
+                cmdr=monitor.cmdr
```
Athanasius, line 1080: This gets `cmdr` back into EDMarketConnector.py:AppWindow.capi_handle_response() ... but then you don't seem to have added anything to populate `capi_response.capi_data.request_cmdr` with it.

Athanasius: Overall I'd adjust `capi_single_query()` to store a copy of `monitor.cmdr` just before the actual `.get()` of the query, and then fill in `request_cmdr` *in CAPIData` when it creates that.  Do the latter by adjusting the CAPIData `__init__()` method.

i.e. it turns out there's no need to mess with EDMCAPIRequest/Response, as it goes into the CAPIData that is held in the latter.

Athanasius (changes requested):

EDMarketConnector.py
```diff
@@ -914,6 +916,9 @@ def set_labels(self):
 
     def login(self):
         """Initiate CAPI/Frontier login and set other necessary state."""
+        should_return: bool
+        new_data: Dict[str, Any] = {}
```
A-UNDERSCORE-D, line 920:
```suggestion
        new_data: dict[str, Any] = {}
```

EDMarketConnector.py
```diff
@@ -1005,6 +1010,9 @@ def capi_request_data(self, event=None) -> None:  # noqa: CCR001
         :param event: Tk generated event details.
         """
         logger.trace_if('capi.worker', 'Begin')
+        should_return: bool
+        new_data: Dict[str, Any] = {}
```
A-UNDERSCORE-D, line 1014:
```suggestion
        new_data: dict[str, Any] = {}
```

EDMarketConnector.py
```diff
@@ -1082,6 +1090,54 @@ def capi_request_data(self, event=None) -> None:  # noqa: CCR001
...
+        """
+        Perform CAPI fleetcarrier data retrieval and associated actions.
+
+        This is triggered by certain FleetCarrier journal events
+
+        :param event: Tk generated event details.
+        """
+        logger.trace_if('capi.worker', 'Begin')
+        should_return: bool
+        new_data: Dict[str, Any] = {}
```
A-UNDERSCORE-D, line 1103:
```suggestion
        new_data: dict[str, Any] = {}
```

EDMarketConnector.py
```diff
@@ -1106,8 +1162,32 @@ def capi_handle_response(self, event=None):  # noqa: C901, CCR001
...
                 raise ValueError(msg)
 
+            if capi_response.capi_data.source_endpoint == companion.session.FRONTIER_CAPI_PATH_FLEETCARRIER:
+                # Fleetcarrier CAPI response
+                # Validation
+                if 'name' not in capi_response.capi_data:
+                    # LANG: No data was returned for the fleetcarrier from the Frontier CAPI
+                    err = self.status['text'] = _('CAPI: No fleetcarrier data returned')
+
+                elif not capi_response.capi_data.get('name', {}).get('callsign'):
```
A-UNDERSCORE-D, line 1172: we have deep_get for this

PLUGINS.md
```diff
@@ -940,19 +940,53 @@ def cmdr_data(data, is_beta):
...
+    # Determining source galaxy for the data
+    if data.source_host == SERVER_LIVE:
+        ...
+
+    elif data.source_host == SERVER_BETA:
+        ...
+
+    elif data.source_host == SERVER_LEGACY:
+        ...
```
A-UNDERSCORE-D, line 966: For @Athanasius  more than anyone else

This is where I'd suggest a `match`

companion.py
```diff
@@ -570,25 +575,29 @@ def __init__(
         self, capi_host: str, endpoint: str,
         query_time: int,
         tk_response_event: Optional[str] = None,
-        play_sound: bool = False, auto_update: bool = False
+        play_sound: bool = False, auto_update: bool = False,
+        cmdr: Optional[str] = None
```
A-UNDERSCORE-D, line 579:
```suggestion
        cmdr: str | None = None
```

companion.py
```diff
@@ -570,25 +575,29 @@ def __init__(
...
 
 
 class EDMCCAPIResponse(EDMCCAPIReturn):
     """Encapsulates a response from CAPI quer(y|ies)."""
 
     def __init__(
             self, capi_data: CAPIData,
-            query_time: int, play_sound: bool = False, auto_update: bool = False
+            query_time: int, play_sound: bool = False, auto_update: bool = False,
+            cmdr: Optional[str] = None
```
A-UNDERSCORE-D, line 596:
```suggestion
            cmdr: str | None = None
```

companion.py
```diff
@@ -663,6 +674,9 @@ def login(self, cmdr: Optional[str] = None, is_beta: Optional[bool] = None) -> b
 
         :return: True if login succeeded, False if re-authorization initiated.
         """
+        should_return: bool
+        new_data: Dict[str, Any] = {}
```
A-UNDERSCORE-D, line 678:
```suggestion
        new_data: dict[str, Any]
```

companion.py
```diff
@@ -781,6 +795,9 @@ def capi_single_query(
             :return: The resulting CAPI data, of type CAPIData.
             """
             capi_data: CAPIData = CAPIData()
+            should_return: bool
+            new_data: Dict[str, Any] = {}
```
A-UNDERSCORE-D, line 799:
```suggestion
            new_data: dict[str, Any]
```

companion.py
```diff
@@ -1024,7 +1046,38 @@ def station(
...
                 query_time=query_time,
                 play_sound=play_sound,
-                auto_update=auto_update
+                auto_update=auto_update,
+                cmdr=monitor.cmdr
+            )
+        )
+
+    def fleetcarrier(
+            self, query_time: int, tk_response_event: Optional[str] = None,
```
A-UNDERSCORE-D, line 1055:
```suggestion
            self, query_time: int, tk_response_event: str | None  = None,
```

plug.py
```diff
@@ -401,6 +401,30 @@ def notify_capidata(
     return error
 
 
+def notify_capi_fleetcarrierdata(
+    data: companion.CAPIData
+) -> Optional[str]:
```
A-UNDERSCORE-D, line 406:
```suggestion
) -> str | None:
```

plug.py
```diff
@@ -401,6 +401,30 @@ def notify_capidata(
...
+    :returns: Error message from the first plugin that returns one (if any)
+    """
+    error = None
+    for plugin in PLUGINS:
+        fc_callback = plugin._get_func('capi_fleetcarrier')
+        if fc_callback is not None and callable(fc_callback):
+            try:
+                # Pass a copy of the CAPIData in case the callee modifies it
+                newerror = fc_callback(copy.deepcopy(data))
+                error = error or newerror
```
A-UNDERSCORE-D, line 420:
```suggestion
                error = error if error else newerror
```

plugins/eddb.py
```diff
@@ -168,6 +168,9 @@ def journal_entry(  # noqa: CCR001
     :param state: `monitor.state`
     :return: None if no error, else an error string.
     """
+    should_return: bool
+    new_entry: Dict[str, Any] = {}
```
A-UNDERSCORE-D, line 172:
```suggestion
    new_entry: dict[str, Any]
```

plugins/eddn.py
```diff
@@ -402,6 +402,9 @@ def send_message(self, msg: str) -> bool:
         :return: `True` for "now remove this message from the queue"
         """
         logger.trace_if("plugin.eddn.send", "Sending message")
+        should_return: bool
+        new_data: Dict[str, Any] = {}
```
A-UNDERSCORE-D, line 406:
```suggestion
        new_data: dict[str, Any]
```

plugins/eddn.py
```diff
@@ -635,6 +638,9 @@ def export_commodities(self, data: CAPIData, is_beta: bool) -> None:  # noqa: CC
         :param data: a dict containing the starport data
         :param is_beta: whether or not we're currently in beta mode
         """
+        should_return: bool
+        new_data: Dict[str, Any] = {}
```
A-UNDERSCORE-D, line 642:
```suggestion
        new_data: dict[str, Any]
```

There are more of these, but Im not going to cover each one :sweat_smile:

A-UNDERSCORE-D (changes requested):

companion.py
aussig: I'm setting `capi_data['request_cmdr']` inside the `EDMCCAPIResponse` instance (`companion.py` line 600), so it's there for every instance, whatever the CAPI request that triggered it.

EDMarketConnector.py
aussig: Might be beginning to regret incorporating this typing change into my PR ;)

companion.py
Athanasius: 1. That's wrong, as it shouldn't be **in the dictionary**.  It should be a property on `CAPIData` instead.
2. If we're doing it for `/fleetcarrier` there's no reason to not do it for all the CAPI queries, hence my suggestion to centralise it in `capi_single_query()`.

EDMarketConnector.py
Athanasius: Yes, tbh I'm about to merge my mypy PR, and that will have fixed/changed these things anyway.  So, when you rebase on top of that, please just drop any of your commits that are only that sort of thing and not about *your* code changes.

plugins/eddn.py
aussig: Ok, I've changed all that I originally added in this PR, but there are many more across the codebase and I don't want to expand the footprint of my PR any further - these should be fixed in a separate PR IMHO.

EDMarketConnector.py
aussig: `deep_get` is defined in the CLI interface `EDMC.py` and is currently only used there. Should I be importing that into `EDMarketConnector.py`?

Athanasius: Leave it for this PR.  If we end up doing that then `deep_get()` needs to end up in `utils/deep_get.py` or somesuch, and that's not a relevant change for this work.

plug.py
aussig: I was following the same pattern as the existing function `notify_capidata`.  Should I change the return type for that function too?  Extending that question further, the pattern `-> Optional[str]:` is used for many functions across the codebase, perhaps there should be a separate PR to review those.

plug.py
aussig: Again following the same pattern as `notify_capidata()`.  Should that function change too?

Athanasius: Yes, just ignore @A-UNDERSCORE-D for these things in this PR.  We'll do such general cleanup in other PRs.

Athanasius:
I've been mulling over how the FC query is triggered, and just realised that triggering it *immediately* on `CarrrierStats` has the potential to delay the "Automatically update on docking" *startup* set of queries.

So, we probably want to make `CarrierStats` cause an `after` call to trigger it, rather than it being immediate.  The specific idea being to ensure it will get triggered *after* the 'station' queries on login.  We do *not* want to delay those at all.

Then there's a whole thing with possibly wanting to run any `/fleetcarrier` query in its own separate thread, so as to never get in the way of the 'station' queries, but that's infrastructure changes not appropriate for *this* PR.

aussig:
It's already using an `after` call: `EDMarketConnector.py` line 1588:

```python
self.w.after(int(SERVER_RETRY * 1000), self.capi_request_fleetcarrier_data)
```

But do you mean somehow ensuring this is also after the 'station' queries?  How would I ensure those had happened though...

Athanasius:
> It's already using an `after` call: `EDMarketConnector.py` line 1588:
> 
> ```python
> self.w.after(int(SERVER_RETRY * 1000), self.capi_request_fleetcarrier_data)
> ```
> 
> But do you mean somehow ensuring this is also after the 'station' queries? How would I ensure those had happened though...

Ah, I'd missed that you were using the retry timer for that, it will probably suffice.  To be after the station queries the delay just needs to be long enough.

However, another thought.  There could be FC owners with large cargo/orders data that don't run any plugin that makes use of the data.  Delaying their station queries for a long time due to a `/fleetcarrier` query taking a long time is bad UX.  So, please add an option, which will default *off*, to enable/disable the FC queries.  I would place this:

1. On the Configuration tab.
2. Above the 'Hotkey' section.

Athanasius (approved):

aussig:
> However, another thought. 

You're a demanding boss ;)   Looking at the current layout of the Configuration panel, it's at the limit of available vertical space.  A couple of options here:

1. We could try to reduce vertical space used on the existing elements. A possible candidate are the horizontal separators which are currently quite tall.
2. We could introduce a scrollbar to the pane.

In both cases, I think I could make a case for this UI stuff to be separated out into a separate issue and therefore PR?

Athanasius:
I want it in this PR so that once this is merged it's in a state where I'd be happy for it to be in a release.  Else if we merge this PR and other things mean we need to make a release we get into a messy "well, either FC owners have to put up with this, or we need to temporarily disable triggering code just for this release, but then remember to undo that after the release branch is forked off, and ensure neither accidentally merges something that changes that state, and ...".

I really don't want to make a release where (some) FC owners are suddenly in a situation where the automatic docking updates of station data just don't happen in a timely manner.  Keep in mind that all those "is it the same system/station as journal says?" checks are upon the received data being processed back in the main thread.  If the player has undocked by then due to a slow FC query then the entirety of the station queries will have been a waste.

Athanasius:
WRT "Settings is already too tall"... you obviously don't have as many disabled plugins as I do.  Those cause the Plugins tab to be taller than any other in my setup, maybe around 55-60% of my 2560x1440 monitor height.

Please just add this option as requested and we can consider changes to accommodate smaller screens in another PR.

aussig:
Ok, I'll get onto it...

EDMarketConnector.py
```diff
@@ -1582,7 +1582,7 @@ def crewroletext(role: str) -> str:
                 if not should_return:
                     self.w.after(int(SERVER_RETRY * 1000), self.capi_request_data)
 
-            if entry['event'] in ('CarrierBuy', 'CarrierStats'):
+            if entry['event'] in ('CarrierBuy', 'CarrierStats') and config.get_int('capi_fleetcarrier') == 1:
```
Athanasius, line 1585: It'll end up *stored* as in int on Windows, but you should be using `get_bool()` and sending it a `bool` value in `config.set()`.

prefs.py
```diff
@@ -470,6 +470,25 @@ def __setup_config_tab(self, notebook: ttk.Notebook) -> None:  # noqa: CCR001
                 state=tk.NORMAL if config.get_str('journaldir') else tk.DISABLED
             ).grid(column=2, pady=self.PADY, sticky=tk.EW, row=row.get())
 
+        # CAPI settings
+        self.capi_fleetcarrier = tk.IntVar(value=config.get_int('capi_fleetcarrier'))
```
Athanasius, line 474: So this should be `tk.BooleanVar(...)`

prefs.py
```diff
@@ -1244,6 +1263,8 @@ def apply(self) -> None:
         else:
             config.set('journaldir', logdir)
 
+        config.set('capi_fleetcarrier', int(self.capi_fleetcarrier.get()))
```
Athanasius, line 1266: And you should just be able to lose the `int(...)` wrapping here.

Athanasius (changes requested):

Athanasius (approved):

Athanasius:
OK, I'm going to merge this and then fix up some fiddly bits, including making `Session.dump_capi_data()` happy when it's `/fleetcarrier` output.

PR #1779, opened 2022-12-22
Get all code passing `pre-commit run --all-files mypy`

Athanasius:
With mypy if you run it literally on a single file it might not take into account the proper types from other files, especially as we have `follow_imports = skip` in `.mypy.ini`.

Using `pre-commit run --all-files mypy` to have it consider all our files at once turned up extra things that needed fixing.

Once this is merged we can look at moving to the default of `follow_imports = normal`, or even `error`.


EDMarketConnector.py
```diff
@@ -16,7 +16,7 @@
 from os import chdir, environ
 from os.path import dirname, join
 from time import localtime, strftime, time
-from typing import TYPE_CHECKING, Literal, Optional, Tuple, Union
+from typing import TYPE_CHECKING, Any, Dict, Literal, Optional, Tuple, Union
```
A-UNDERSCORE-D, line 19: if possible, swapping all `Dict` to `dict` could be nice, though note that this breaks 3.8 for those that wanted to use it

config/__init__.py
```diff
@@ -307,7 +307,10 @@ def _suppress_call(
 
         return None
 
-    def get(self, key: str, default: Union[list, str, bool, int] = None) -> Union[list, str, bool, int]:
+    def get(
+        self, key: str,
+        default: Union[list, str, bool, int, None] = None
+    ) -> Union[list, str, bool, int, None]:
```
A-UNDERSCORE-D, line 313:
```suggestion
        default: list | str | bool | int | None = None
    ) -> list | str | bool | int | None:
```

Or something that defines a TypeVar that can be all of those, then accept it as the default and use it as the returntype, which would be something like

```py
T = typing.TypeVar('T', list, str, bool, int, None)
def get(self, key: str, default: T) -> T:
    ...
```
the returntype here would then become defined by default, so it may break things elsewhere

docs/Releasing.md
```diff
@@ -166,7 +166,7 @@ that it's actually included in the installer.
         <File KeyPath="yes" Source="SourceDir\\<file name>" />
    </Component>
    ```
-
+ | None
```
A-UNDERSCORE-D, line 169: paste mistake?

hotkey/windows.py
```diff
@@ -0,0 +1,359 @@
...
+            text += u'Ctrl+'
+
+        if modifiers & MOD_ALT:
+            text += u'Alt+'
+
+        if modifiers & MOD_SHIFT:
+            text += u'⇧+'
+
+        if VK_NUMPAD0 <= keycode <= VK_DIVIDE:
+            text += u'№'
```
A-UNDERSCORE-D, line 335:
```suggestion
        if modifiers & MOD_WIN:
            text += '❖+'

        if modifiers & MOD_CONTROL:
            text += 'Ctrl+'

        if modifiers & MOD_ALT:
            text += 'Alt+'

        if modifiers & MOD_SHIFT:
            text += '⇧+'

        if VK_NUMPAD0 <= keycode <= VK_DIVIDE:
            text += '№'
```

Unicode strings do nothing in py3

plugins/eddn.py
```diff
@@ -635,6 +635,8 @@ def export_commodities(self, data: CAPIData, is_beta: bool) -> None:  # noqa: CC
         :param data: a dict containing the starport data
         :param is_beta: whether or not we're currently in beta mode
         """
+        should_return: bool
+        new_data: Dict[str, Any]
```
A-UNDERSCORE-D, line 639:
```suggestion
        new_data: dict[str, Any]
```

plugins/eddn.py
```diff
@@ -766,6 +768,8 @@ def export_outfitting(self, data: CAPIData, is_beta: bool) -> None:
         :param data: dict containing the outfitting data
         :param is_beta: whether or not we're currently in beta mode
         """
+        should_return: bool
+        new_data: Dict[str, Any]
```
A-UNDERSCORE-D, line 772:
```suggestion
        new_data: dict[str, Any]
```

plugins/eddn.py
```diff
@@ -832,6 +836,8 @@ def export_shipyard(self, data: CAPIData, is_beta: bool) -> None:
         :param data: dict containing the shipyard data
         :param is_beta: whether or not we are in beta mode
         """
+        should_return: bool
+        new_data: Dict[str, Any]
```
A-UNDERSCORE-D, line 840:
```suggestion
        new_data: dict[str, Any]
```

stats.py
```diff
@@ -478,7 +480,7 @@ def addpagespacer(self, parent) -> None:
         self.addpagerow(parent, [''])
 
     def addpagerow(
-        self, parent: tk.Frame, content: Sequence[str], align: Optional[str] = None, with_copy: bool = False
+        self, parent: ttk.Frame, content: Sequence[str], align: Optional[str] = None, with_copy: bool = False
```
A-UNDERSCORE-D, line 483:
```suggestion
        self, parent: ttk.Frame, content: Sequence[str], align: str | None = None, with_copy: bool = False
```

stats.py
```diff
@@ -462,7 +464,7 @@ def addpage(self, parent, header: List[str] = None, align: Optional[str] = None)
 
         return page
 
-    def addpageheader(self, parent: tk.Frame, header: Sequence[str], align: Optional[str] = None) -> None:
+    def addpageheader(self, parent: ttk.Frame, header: Sequence[str], align: Optional[str] = None) -> None:
```
A-UNDERSCORE-D, line 467:
```suggestion
    def addpageheader(self, parent: ttk.Frame, header: Sequence[str], align: str | None = None) -> None:
```

A-UNDERSCORE-D (changes requested):

EDMarketConnector.py
Athanasius: I've been slinging `A | B` around recently, so we're actually Python 3.10 minimum now.  So, yeah, I'll try to remember to avoid typing.Dict and typing.List etc.

config/__init__.py
Athanasius: I'll go with the Python 3.10-esque `|` version, because:

```python
getT = TypeVar('getT', list, str, bool, int, None)

    def get(
        self, key: str,
        default: getT = None
    ) -> getT:
```
yields:
```python
0$ mypy --follow-imports normal config/__init__.py
config\__init__.py:315: error: Incompatible default for argument "default" (default has type "None", argument has type "List[Any]")  [assignment]
config\__init__.py:315: note: PEP 484 prohibits implicit Optional. Accordingly, mypy has changed its default to no_implicit_optional=True
config\__init__.py:315: note: Use https://github.com/hauntsaninja/no_implicit_optional to automatically upgrade your codebase
config\__init__.py:315: error: Incompatible default for argument "default" (default has type "None", argument has type "str")  [assignment]
config\__init__.py:315: error: Incompatible default for argument "default" (default has type "None", argument has type "bool")  [assignment]
config\__init__.py:315: error: Incompatible default for argument "default" (default has type "None", argument has type "int")  [assignment]
config\__init__.py:332: error: Cannot determine type of "a_str"  [has-type]
config\__init__.py:335: error: Cannot determine type of "a_bool"  [has-type]
config\__init__.py:338: error: Cannot determine type of "an_int"  [has-type]
Found 7 errors in 1 file (checked 1 source file)
```
It's the default value on `default` that causes that first line.  And in this case we do want to set that.

docs/Releasing.md
Athanasius: Yeah, has to have been, fixed.

hotkey/windows.py
Athanasius: Legacy code ahoy !   Fixed in this file.  I'm going to hope that *when* I get around to trying `black` again it would take care of this sort of thing ?

Athanasius:
Changes pushed for all resolved comments.

hotkey/windows.py
A-UNDERSCORE-D:
```
--- test.py     2022-12-23 12:26:01.724569 +0000
+++ test.py     2022-12-23 12:26:07.943035 +0000
@@ -1 +1 @@
-u"test"
+"test"
would reformat test.py

All done! ✨ 🍰 ✨
1 file would be reformatted.
```
yep!

Athanasius: OK, I do intend another stab at black, and trying really hard to not get angry with it messing anything up ;).

PR #1781, opened 2022-12-23
Make mypy happy on linux

A-UNDERSCORE-D:
Some simple-ish changes to make mypy stop being upset that linux exists.

Tested that EDMC does still launch (on linux)

- config/linux.py Fix mypy complaints
- tests/config: Disable mypy on _old_config.py
- build: add "dumb" platform check for mypy
- myNotebook.py: Make platform check simpler
- hotkey/linux.py: Update stub
- hotkeys/windows.py: Add explicit platform check
- EDMarketConnector.py: Guard platform specific func
- protocol.py: Make mypy happy

### DIFF
--- a/protocol.py
+++ b/protocol.py
@@ -121,9 +121,13 @@ elif (config.auth_force_edmc_protocol
           and not is_wine
           and not config.auth_force_localserver
       )):
+    # This could be false if you use auth_force_edmc_protocol, but then you get to keep the pieces
+    assert sys.platform == 'win32'
     # spell-checker: words HBRUSH HICON WPARAM wstring WNDCLASS HMENU HGLOBAL
     from ctypes import windll  # type: ignore
-    from ctypes import POINTER, WINFUNCTYPE, Structure, byref, c_long, c_void_p, create_unicode_buffer, wstring_at
+    from ctypes import (  # type: ignore
+        POINTER, WINFUNCTYPE, Structure, byref, c_long, c_void_p, create_unicode_buffer, wstring_at
+    )
     from ctypes.wintypes import (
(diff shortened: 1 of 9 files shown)


A-UNDERSCORE-D:
Done messing this now and is good for a review

protocol.py
```diff
@@ -121,9 +121,13 @@ def handleEvent_withReplyEvent_(self, event, replyEvent) -> None:  # noqa: N802
           and not is_wine
           and not config.auth_force_localserver
       )):
+    # This could be false if you use auth_force_edmc_protocol, but then you get to keep the pieces
+    assert sys.platform == 'win32'
     # spell-checker: words HBRUSH HICON WPARAM wstring WNDCLASS HMENU HGLOBAL
     from ctypes import windll  # type: ignore
-    from ctypes import POINTER, WINFUNCTYPE, Structure, byref, c_long, c_void_p, create_unicode_buffer, wstring_at
+    from ctypes import (  # type: ignore
```
Athanasius, line 128: Why *this* `# type: ignore` ?

A-UNDERSCORE-D: because it seems that mypy ignores the assert above for the imports, and thus decides that those imports dont exist (WINFUNCTYPE for the second one, windll for the first)

Athanasius: Ah, fair enough.

protocol.py
```diff
@@ -121,9 +121,13 @@ def handleEvent_withReplyEvent_(self, event, replyEvent) -> None:  # noqa: N802
           and not is_wine
           and not config.auth_force_localserver
       )):
+    # This could be false if you use auth_force_edmc_protocol, but then you get to keep the pieces
```
Athanasius, line 124: Hmmm, tbh we should change the args parsing section to enforce that then.

PR #1791, opened 2022-12-31
UI: Properly name main widgets & put each plugin in its own Frame

Athanasius:
Given talk about adding the ability to change the UI ordering of third-party plugins I started looking at the totality of plugins in the main UI.  Most of them had no `name=...` in their creation so ended up with non-useful tkinter names.  This PR ensures each one has a pertinent name.

Then I experimented with creating a per-plugin frame for all its UI to be in, rather than it being directly in the main UI Frame.  This appears to have no ill effects on the three plugins with UI I have running (aussi's BGS-Tally, Hourly Income and FCMS).  The one piece of fallout was to switch to `nametowidget()` in some of the core plugin code where it needs to access system/station widgets or set status text.

There is no sanctioned reason for a third-party plugin to be trying to poke at/get information about the main UI widgets, so this change causing them to be one step down in the family tree *shouldn't* be a breaking one for them.

Specifically:
```python
	some_widget = self.parent.children['somewidgetname']
```
would have to change to:
```python
	some_widget = self.parent.parent.children['somewidgetname']
```
But, actually the correct way to do this is:
```python
from config import appname

def some_function():
	some_widget = this.parent.nametowidget(f".{appname.lower()}.somewidgetname")
```
i.e. look up the widget by its full tk name.  This method doesn't require knowing how many generations down you are.

### DIFF
--- a/EDMarketConnector.py
+++ b/EDMarketConnector.py
@@ -578,8 +578,9 @@ class AppWindow(object):
                 plugin_sep.grid(columnspan=2, sticky=tk.EW)
                 ui_row = frame.grid_size()[1]
                 plugin_frame.grid(
-                    row=ui_row, columnspan=2, sticky=tk.EW
+                    row=ui_row, columnspan=2, sticky=tk.NSEW
                 )
+                # plugin_frame.columnconfigure(1, weight=1)
                 if isinstance(appitem, tuple) and len(appitem) == 2:
                     ui_row = frame.grid_size()[1]
                     appitem[0].grid(row=ui_row, column=0, sticky=tk.W)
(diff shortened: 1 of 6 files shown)


A-UNDERSCORE-D (approved):

A-UNDERSCORE-D:
I like this, decouples stuff a bunch more

Athanasius:
I tested every plugin listed on the Wiki page, one by one (sometimes EDMCOverlay also active).

EDMC-Canonn highlighted an issue where its own code to adjust wrapping of text (to cause vertical resizing instead of horizontal) produced difference results.  However [61ad75d](https://github.com/EDCD/EDMarketConnector/pull/1791/commits/61ad75d52a44f3b26c4a93a61ea02e6900451ddd) and [7bfeb5e](https://github.com/EDCD/EDMarketConnector/pull/1791/commits/7bfeb5ec621a965a17164fb753934b5268715f84) fixed this.

No other issues observed, although I didn't check each plugin in detail, and skipped ones where no UI is shown until some ingame/other action is taken.

PR #1798, opened 2023-01-09
Move core plugin 'location' tracking into monitor.py

### LINKED ISSUE #1042: Move EDDB, EDSM, Inara system/station tracking into monitor.py
The necessary might already be tracked in monitor.state.  Ensure it is, and then use *that* rather than all the extra checking in these three plugins.

Athanasius:
As well as being duplicate code in the core plugins, there's also a bug related to plugins not receiving events during Journal catch up.  This means that current release plugins/eddn.py code doesn't properly know where it is if you relog on a body surface, meaning the CodexEntry handling then doesn't have the data for cross-checks.

Fix this by extending monitor.state tracking to include the necessary data, and relying on that in the EDDN plugin.

NB: Still need to check the *other* core plugins to update their tracking to reflect they can just rely on `journal_entry()`'s `state` now.

Closes #1042

### DIFF
--- a/EDMC.py
+++ b/EDMC.py
@@ -326,10 +326,12 @@ sys.path: {sys.path}'''
             raise companion.CmdrError()
 
         elif (
-            data['lastSystem']['name'] != monitor.state['SystemName'] or
-            ((data['commander']['docked'] and data['lastStarport']['name'] or None) != monitor.station) or
-            data['ship']['id'] != monitor.state['ShipID'] or
-            data['ship']['name'].lower() != monitor.state['ShipType']
+            data['lastSystem']['name'] != monitor.state['SystemName']
+            or (
+                (data['commander']['docked'] and data['lastStarport']['name'] or None) != monitor.state['StationName']
+            )
+            or data['ship']['id'] != monitor.state['ShipID']
+            or data['ship']['name'].lower() != monitor.state['ShipType']
         ):
             raise companion.ServerLagging()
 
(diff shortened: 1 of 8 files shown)


Athanasius:
@A-UNDERSCORE-D Please take particular care when reviewing this (still draft atm).  It's not just surface code correctness that needs verifying, but also that I've not actually broken any of the EDDN tracking/checks in the process.

I *think* it's correct after this morning's commits, where I was looking at this PR's monitor.py and plugins/eddn.py versus `develop`'s current plugins/eddn.py (as of f8a2e6916d8fcd91c30dac0bd20b0257f2e151bf ).

Athanasius:
That should be all the relevant code changed in both monitor.py and the core plugins.

I've only done very basic testing (with the "move latest journal out, run EDMC, move file back in" trick) to be sure System/Station fields are still getting populated.

A-UNDERSCORE-D (approved):
Im reasonably happy that you havent broken behavior; however some playtesting might be appropriate if theres anyone that wants to do that pre-beta?

Athanasius:
> Im reasonably happy that you havent broken behavior; however some playtesting might be appropriate if theres anyone that wants to do that pre-beta?

We already had a little testing from @slippycheeze on Discord -> <https://discord.com/channels/164411426939600896/251742967927144449/1061984192088526959>

But, yes, I should probably knuckle down, enable the EDDN tracking UI, and do some testing.

PR #1800, opened 2023-01-10
.desktop: Add a file, and adjust icon filename to suit

Athanasius:
There is one issue with the Icon, it won't be in any path that will be searched.  e.g. the flathub does:
```yaml
install -Dm644 edmarketconnector/EDMarketConnector.png ${FLATPAK_DEST}/share/icons/hicolor/512x512/apps/${FLATPAK_ID}.png
```

So that's an *additional* manual issue for a (non-flathub) Linux user wanting everything to Just Work.  The other being they're going to have to create an `edmarketconnector` shell script in their PATH for this to even work:

1. It will need to know the path to `EDMarketConnector.py`.
2. It will **NEED** to `cd` to that path, else the program won't function correctly.

This is starting to sound more and more like we need a `scripts/linux-setup` to take care of these things.


Athanasius:
@der-eismann heads up that the flatpak will need slightly adjusting for the rename of the .png file in this PR.

Athanasius:
> @der-eismann heads up that the flatpak will need slightly adjusting for the rename of the .png file in this PR.

Also, this imports (slightly edited) the .desktop file.  It *should* still suit the flatpak, let me know if anything is in contention with its specific needs.  I removed the `X-Desktop-File-Install-Version=0.21` because I couldn't find any reference/documentation for that.

der-eismann:
Thanks for the notice! 
> I removed the X-Desktop-File-Install-Version=0.21 because I couldn't find any reference/documentation for that.

Actually I don't know where I got this from, probably some desktop-file-generator.

PR #1802, opened 2023-01-11
Create a custom .zip archive for releases

### LINKED ISSUE #1801: Running from source also needs FDevIDs files
Ensure any documentation for running from source points out/instructs that if *not* using a git clone the user will need to ensure the FDevIDs files (commodity.csv and rare_commodity.csv) files are available.

This only affects some niche 'Output' functionality, but is still necessary.

Alternatively, see if we can get GitHub to include these in the generated archives.  It includes the directories of the sub-modules, but not anything they *contain*.

Athanasius:
Because the GitHub automatic .zip (and .tar.gz) archive is based purely on the files in *this* repo it doesn't include the two necessary files from the FDevIDs sub-module.

So, this PR adds an extra build step to make such an archive *with* those files.  Actually I felt it easiest to just include all the FDevIDs files.  The resulting .zip file is then an extra build artifact which is included in subsequent steps, i.e. hashes.sum and release draft creation.

I've left the most recent successful draft release at https://github.com/EDCD/EDMarketConnector/releases/tag/untagged-d68b90c527b0e582d027 (likely only viewable by collaborators).

We should be able to extend this to a custom .tar.gz as well.

Close #1801

### DIFF
--- a/.github/workflows/windows-build.yml
+++ b/.github/workflows/windows-build.yml
@@ -71,7 +71,7 @@ jobs:
   release:
     name: Release new version
     runs-on: ubuntu-latest
-    needs: test
+    needs: windows_build
     if: "${{ github.event_name != 'workflow_dispatch' }}"
 
     steps:
--- a/.github/workflows/windows-build.yml
+++ b/.github/workflows/windows-build.yml
@@ -7,13 +7,6 @@ on:
       - "Release/*"
   workflow_dispatch:
 
-env:
-  # If this value gets too long it can cause issues with Windows
-  # command-line length and zip/tar will fail
-  # And all the exclusions need to be prefixed with the 'root'
-  # directory.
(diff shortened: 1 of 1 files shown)


Athanasius:
TODO: Rather than that looooooooooooong exclude list, which can cause issues if it gets too long (zip fails with command arg parsing error), perhaps have an intermediate step to, carefully, remove all but only the files that are still necessary.

Will likely need to leave .git/ in place though, so that would have to still be excluded.

Athanasius:
TODO: Once this is merged for an actual release we should update [the wiki guide](https://github.com/EDCD/EDMarketConnector/wiki/Running-from-source#obtain-a-copy-of-the-application-source) to specify using these custom archives, not the automatic ones.

A-UNDERSCORE-D (approved):

Athanasius:
We might as well do the work to have a custom .tar.gz archive as well.

And then there's the suggest to use something like [repo-file-sync-action](https://github.com/marketplace/actions/repo-file-sync-action) *instead of* FDevIDs as a sub-module.  i.e. go back to having the two .csv files in the main repo and use that action to have PRs bring in newer versions as they're committed upstream.

Athanasius:
Right, tar archive creation finally working, as per the last build https://github.com/EDCD/EDMarketConnector/actions/runs/3902861020/jobs/6666497920 which created https://github.com/EDCD/EDMarketConnector/releases/tag/untagged-b41f5c83a227f648c94d

Both of those archives seem to yield a working application.  Tested with a python 3.10.9 venv I had laying around on my home Linux server.  They run, no readily apparent errors, main UI comes up.  I don't have journal file location correct, so not much else is testable there.

Athanasius:
I'm going to merge this now before I start doing things like "'out' is a crap name for the WinSparkle stuff, and why not clean up the directory and zip file immediately after the files are in place?".

PR #1808, opened 2023-01-12
Migrate all ctypes code to use pywin32 instead, if possible

### LINKED ISSUE #1805: 64-bit: Properly prototype *all* the used functions
LCU No Fool Like One reported seeing:
```python
exception ignored on calling ctypes callback function: <function EDLogs.game_running.<locals>.callback at 0x0000020E9C9D0EA0>
  File "monitor.pyc", line 2041, in callback
  File "monitor.pyc", line 2034, in WindowTitle
ctypes.ArgumentError: argument 1: OverflowError: int too long to convert
```
presumably on the 5.7.1-alpha1 64-bit pre-release.  That makes `File "monitor.pyc", line 2034, in WindowTitle` to be the line:
```python
length = GetWindowTextLength(h) + 1
```
That function hasn't been properly prototyped, so will be returning `int`, instead of the proper type.  I note that this is also actually [GetWindowTextLengthW](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-getwindowtextlengthw).

So, we should really just properly prototype *all* of the ctypes functions we use to be sure of no issues going forwards.

Athanasius:
This started out as properly prototyping the ctypes code in `monitor.py`, so that nothing gets confused by return types on 64-bit python/windows...

... but after struggling for far too long with `GetWindowTextW()`, due to its *second* (out of three) parameter being 'out'... and realising that ctypes 'helpfully' returns all 'out' as the function return, and that means you need to define a function to assign to the `.errcheck` property in order to actually check result (return) ... I came across how to do the same thing with pywin32 modules instead.

And now I've managed to eliminate all the ctypes code from `monitor.py`.  I'll look into all the other ctypes code as well.

This does mean we'll start having `pywin32` as a PIP-level dependency ... and I *am* wondering if py2exe will do the right thing with all the separate little modules it provides.  They *are* explicit imports in our code, so here's hoping.

Close #1805

### DIFF
--- a/EDMarketConnector.py
+++ b/EDMarketConnector.py
@@ -252,7 +252,6 @@ if __name__ == '__main__':  # noqa: C901
                 import win32process
                 from ctypes.wintypes import BOOL, HWND, INT, LPARAM, LPCWSTR, LPWSTR
 
-                EnumWindows = ctypes.windll.user32.EnumWindows  # noqa: N806
                 GetClassName = ctypes.windll.user32.GetClassNameW  # noqa: N806
                 GetClassName.argtypes = [HWND, LPWSTR, ctypes.c_int]
                 GetWindowText = ctypes.windll.user32.GetWindowTextW  # noqa: N806
@@ -278,7 +277,6 @@ if __name__ == '__main__':  # noqa: C901
 
                     return None
 
-                @ctypes.WINFUNCTYPE(BOOL, HWND, LPARAM)
                 def enumwindowsproc(window_handle, l_param):  # noqa: CCR001
                     """
                     Determine if any window for the Application exists.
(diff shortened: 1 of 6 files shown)


Athanasius:
**NB: I'm not 100% sure that:
```python
handle = win32api.OpenProcess(win32con.PROCESS_TERMINATE, False, process_id)
```
will definitely only return a handle for the current user's processes.  It *might* do so for other users' processes if the current one is a Windows Administrator ?

There's an alternative method of getting process info using the `wmi` module which can tell you the owner of it, but in testing it is `v -e- rrrrr-yyyyyy     s-lllll-ooooooo- wwww`.  We end up calling `EDLogs.game_running()` about once a second because we poll the current journal file, rather than trying to do anything clever with a file watcher, so that's not at all viable.

Athanasius:
I also note that the whole of the *old* code only finds the `Elite - Dangerous*` process/window if it's running in the visible session.  So it'll find such that was `runas`'d, but not if you actually switch user, run the process, then switch back to the one running EDMarketConnector.

The same as also true of the new code in this PR.

Athanasius:
I double checked both old and new code through `runas`'ing steam and then the game (that alt is my Steam one), and indeed that process check *does not actually determine if the game is running as the current user, only if it's in the current user session*.

Issue #2031, opened 2023-07-22
Remove references to EDDB


Rixxan:
**Describe the enhancement**

Update dependencies and remove EDDB references. EDDB has been shut down, and thus there is no point in continuing to include EDDB files in EDMC. 

**Expected behavior**

EDDB code should no longer be included in the project, and the default EDDB plugin should be removed.

**Considered workarounds**

Currently, the EDDB code sits in place and while it does not stop the program from functioning, it does add excess complexity which is not needed at this time.

**Additional context**

This will necessitate a Minor version change, and is not compatible with a simple patch change.

PR #2032, opened 2023-07-22
#2031 Remove EDDB Code

Rixxan:
Removes EDDB Code from EDMC, and updates dependency files to latest working versions.


C1701D:
LGTM

PR #2078, opened 2023-09-18
Fix Inara URLs for station & system search

der-eismann:
Looks like Inara changed their URL format. When visiting it for the first time, I had to choose for which game I'm coming. After choosing Elite, the old links work again. But these seem to be the correct ones.


Rixxan (approved):
I like it!

PR #2085, opened 2023-10-30
Updated the Italian translation

GLWine:
Added and corrected missing lines of code for the Italian language.
Change made to Inara.py and EDSM.py files:
Fixed the section of code that integrates the Show API Key text translation call that was not written in such a way as to be actually translated.

### DIFF
--- a/L10n/it.strings
+++ b/L10n/it.strings
@@ -1,8 +1,14 @@
 /* Language name */
 "!Language" = "Italiano";
 
+/* companion.py: Frontier CAPI didn't respond; In files: companion.py:226; */
+"Error: Frontier CAPI didn't respond" = "Errore: Frontier CAPI non ha risposto"
+
 /* companion.py: Frontier CAPI data doesn't agree with latest Journal game location; In files: companion.py:245; */
 "Error: Frontier server is lagging" = "Errore: il server Frontier sta laggando";
+
+/* companion.py: Commander is docked at an EDO settlement, got out and back in, we forgot the station; In files: companion.py:261; */
+"Docked but unknown station: EDO Settlement?" = "Stazione attraccata ma sconosciuta: insediamento EDO?";
 
 /* companion.py: Generic "something went wrong with Frontier Auth" error; In files: companion.py:271; */
 "Error: Invalid Credentials" = "Errore: Credenziali non valide";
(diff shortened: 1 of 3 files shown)


Athanasius:
@Rixxan and @C1701D need to get up to speed with the OneSky workflow.

Translators should see https://github.com/EDCD/EDMarketConnector/issues/24 .

Merging this PR would just get undone the next time OneSky translations are merged in (special-casing Italian because of this PR would just cause long-term issues, the updated translations **really** need to be on OneSky unless the active developers decide to stop using that).

Now I check, I see that GLWine *has* been attempting update on OneSky as well.

So, yes, let's sort out the new developers having access to OneSky.  I *do* have a secondary account with all the admin rights (from when github auth was broken) which we can probably get transferred across.  I would highly suggest doing something like making a new gmail/hotmail/whatever account **so that it could be transferred more easily if needs be in the future** and getting it migrated to that.

Athanasius:
Here's the latest OneSky download of *all* the translations in their current state:
[EDMarketConnector-Tanslations-OneSky-20231030.zip](https://github.com/EDCD/EDMarketConnector/files/13208364/EDMarketConnector-Tanslations-OneSky-20231030.zip)

GLWine:
I apologize, but the part about email and GitHub is not entirely clear to me. If you can explain better. For me, I still understand the choice not to integrate the pool, however the part about the two files EDSM.py and Inara.py. 

In OneSky I made comments about some blocked translations, which perhaps could be changed.

Athanasius:
> I apologize, but the part about email and GitHub is not entirely clear to me. If you can explain better. For me, I still understand the choice not to integrate the pool, however the part about the two files EDSM.py and Inara.py.
> 
> In OneSky I made comments about some blocked translations, which perhaps could be changed.

That was for @Rixxan and @C1701D , you don't need to worry about it.

GLWine:
> > I apologize, but the part about email and GitHub is not entirely clear to me. If you can explain better. For me, I still understand the choice not to integrate the pool, however the part about the two files EDSM.py and Inara.py.
> > In OneSky I made comments about some blocked translations, which perhaps could be changed.
> 
> That was for @Rixxan and @C1701D , you don't need to worry about it.

At this point I close the pull, and redo one of just the Inara.py and EDSM.py files. So as to at least correct the missing hook for the translation.

PR #2099, opened 2023-11-17
Add spansh as a URL provider

### LINKED ISSUE #1490: Spansh: Implement URL providers
spansh.co.uk is a good source of data.  Some users might prefer it for the System/Station providers.

```
Spansh: both of those are trivial to work with https://spansh.co.uk/station/<market_id> https://spansh.co.uk/system/<id64>
```

C1701D:
# Description
This adds spansh as a URL provider for system and station URLs. It does so by adding a core plugin called ``spansh_core`` with said functionalities. The reason for that name is to make sure that custom plugins called ``spansh`` that users may already have installed don't result in either of these two plugins breaking. The ``spansh_core`` plugins is also designed to be expandable in the future if additional features should be added.

Closes #1490 

## Type of change
- Adds a spansh plugin as a URL provider for systems and stations.

# How Has This Been Tested?
Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration
- Multiple install runs.
- Tested Plugin and core functionality in all relevant in-game situations.
- All PyTest tests pass.

### DIFF
--- a/plugins/spansh_core.py
+++ b/plugins/spansh_core.py
@@ -1,0 +1,195 @@
+"""
+spansh_core.py - Spansh URL provider.
+
+Copyright (c) EDCD, All Rights Reserved
+Licensed under the GNU General Public License.
+See LICENSE file.
+
+This is an EDMC 'core' plugin.
+All EDMC plugins are *dynamically* loaded at run-time.
+
+We build for Windows using `py2exe`.
+`py2exe` can't possibly know about anything in the dynamically loaded core plugins.
+
+Thus, you **MUST** check if any imports you add in this file are only
+referenced in this file (or only in any other core plugin), and if so...
+
+    YOU MUST ENSURE THAT PERTINENT ADJUSTMENTS ARE MADE IN
+    `build.py` TO ENSURE THE FILES ARE ACTUALLY PRESENT
+    IN AN END-USER INSTALLATION ON WINDOWS.
+"""
+from __future__ import annotations
+
+import tkinter as tk
+from typing import Any, cast
+import requests
+from companion import CAPIData
+from config import appname, config
+from EDMCLogging import get_main_logger
+
+logger = get_main_logger()
+
+
+class This:
+    """Holds module globals."""
+
+    def __init__(self):
+        self.parent: tk.Tk
+        self.shutting_down = False  # Plugin is shutting down.
+        self.system_link: tk.Widget = None  # type: ignore
+        self.system_name: str | None = None  # type: ignore
+        self.system_address: str | None = None  # type: ignore
+        self.system_population: int | None = None
+        self.station_link: tk.Widget = None  # type: ignore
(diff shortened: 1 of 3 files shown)


plugins/spansh_core.py
```diff
@@ -0,0 +1,195 @@
...
+    """Holds module globals."""
+
+    def __init__(self):
+        self.parent: tk.Tk
+        self.shutting_down = False  # Plugin is shutting down.
+        self.system_link: tk.Widget = None  # type: ignore
+        self.system_name: str | None = None  # type: ignore
+        self.system_address: str | None = None  # type: ignore
+        self.system_population: int | None = None
+        self.station_link: tk.Widget = None  # type: ignore
```
Rixxan, line 43: If assigning to None, should this be `tk.Widget | None`?

plugins/spansh_core.py
```diff
@@ -0,0 +1,195 @@
...
+logger = get_main_logger()
+
+
+class This:
+    """Holds module globals."""
+
+    def __init__(self):
+        self.parent: tk.Tk
+        self.shutting_down = False  # Plugin is shutting down.
+        self.system_link: tk.Widget = None  # type: ignore
```
Rixxan, line 39: If assigning to None, should this be `tk.Widget | None`?

plugins/spansh_core.py
```diff
@@ -0,0 +1,195 @@
...
+        self.system_name: str | None = None  # type: ignore
+        self.system_address: str | None = None  # type: ignore
```
Rixxan, line 41: Do we need the `#type: ignore` here? Does MyPy complain without it?

plugins/spansh_core.py
```diff
@@ -0,0 +1,195 @@
...
+
+this = This()
+STATION_UNDOCKED: str = '×'  # "Station" name to display when not docked = U+00D7
+
+
+def plugin_start3(plugin_dir: str) -> str:
+    """
+    Start the plugin.
+
+    :param plugin_dir: NAme of directory this was loaded from.
```
Rixxan, line 57: Nit: NAme

Rixxan (commented):
No blocking comments, but a few nits about type hinting and S/G

plugins/spansh_core.py
C1701D: ``self.station_link`` should only be ``None`` while the main window is being created, once ``plugin_app()`` gets called ``self.station_link`` becomes a ``tk.Widget`` and shouldn't become ``None`` again.

plugins/spansh_core.py
C1701D: ``self.system_link`` should only be ``None`` while the main window is being created, once ``plugin_app()`` gets called ``self.system_link`` becomes a ``tk.Widget`` and shouldn't become ``None`` again.

PR #2109, opened 2023-12-13
[555] Move Default Registry Creation

### LINKED ISSUE #555: [Installer] Create 'defaults' Registry keys at install time
Check if there's a way to have the MSI installer create certain Registry keys, with default values, if and only if they don't already exist.

Then we don't need to be trying to set things like WinSparkle options in the EDMC code (search config.py for WinSparkle).

Rixxan:
# Description
Attempts to move the WinSparkle key creation from the Windows config itself to the InnoSetup INI file. This is done in such a way as to try and reduce some overhead as well as create these values if and only if they don't already exist. 

## Type of change
- Feature Enhancement

# How Has This Been Tested?
- Tests of installation, operation, and building on the developer's machine
- All PyTest tests pass
- Manual edits of registry values to test survivability

Closes #555

### DIFF
--- a/build.py
+++ b/build.py
@@ -19,12 +19,16 @@ from config import (
     copyright,
     git_shorthash_from_head,
     _static_appversion,
+    update_interval
 )
 
 
 def iss_build(template_path: str, output_file: str) -> None:
     """Build the .iss file needed for building the installer EXE."""
-    sub_vals = {"appver": _static_appversion}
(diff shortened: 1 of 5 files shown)


C1701D (approved):
Confirmed that the new code is functionally equivalent to or better (as in it doesn't override the check for updates setting if the user wants to permanently disable it) than the old code.

PR #2110, opened 2023-12-14
Broken Plugins Popup

C1701D:
# Description
Adds a popup at startup about plugins that failed to load, presumably due to a wrong folder structure. This popup will come up at every EDMC start, unlike the "Plugin without Python 3.x support" one that only pops up every 24h. This PR also adds a list of those Plugins under their own entry in the Settings - Plugins tab.

## Type of change
- Adds a Popup at startup about Plugins that failed to load.
- Adds entry about broken Plugins in the Settings - Plugins tab.
- Adds corresponding translation entries.
- Includes a minor fix for try/except if coriolis-data is missing.

# How Has This Been Tested?
Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration
- All PyTest tests pass.
- `find_localised_strings.py` doesn't complain.
- Multiple install runs.
- Tested feature with multiple "broken" plugins.

### DIFF
--- a/EDMarketConnector.py
+++ b/EDMarketConnector.py
@@ -2325,6 +2325,7 @@ sys.path: {sys.path}'''
 
     root.wm_attributes('-alpha', ui_transparency / 100)
     # Display message box about plugins without Python 3.x support
+    # Display message box about plugins that failed to load
     root.after(0, messagebox_broken_plugins)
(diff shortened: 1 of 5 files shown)


Rixxan (approved):
Changes look good and appear to operate as intended.

PR #2111, opened 2023-12-15
[1801] Add Warning for FDevIDs

Rixxan:
# Description
Adds a popup at startup about missing FDevIDs files on startup to EDMC. This popup will come up at every EDMC start, unlike the "Plugin without Python 3.x support" one that only pops up every 24 hours. This includes an option to open the wiki to view the instructions for installing the Git Submodule. 

This should only be an issue encountered if running from the source, as the packaged/distributed installs will include this information and files by default. The popup is designed so that it will only pop up once even if both fdevid files are missing or only one is missing. 

## Type of change
- Adds a Popup at startup about Plugins that failed to load.
- Adds corresponding translation entries.

# How Has This Been Tested?
- All PyTest tests pass.
- `find_localised_strings.py` doesn't complain.
- Tested feature with missing and found FDEVIDs

This is a measure related to #1801 but does not close. Further research should be done for later versions to determine how to avoid the need for this entirely.

### DIFF
--- a/EDMarketConnector.py
+++ b/EDMarketConnector.py
@@ -2319,11 +2319,11 @@ sys.path: {sys.path}'''
             config.set('plugins_not_py3_last', int(time()))
 
     def check_fdev_ids():
-        """Display message about missing FDEVID files"""
+        """Display message about missing FDEVID files."""
         fdev_files = {'commodity.csv', 'rare_commodity.csv'}
         for file in fdev_files:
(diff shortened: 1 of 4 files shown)


C1701D (approved):
Looks good, all functions using the FDevID files have been tested with different combinations of files available.

Rixxan:
Confirming expected behavior w/ the changes C17 made :)

PR #2117, opened 2023-12-24
[Minor] Update Python Version, Deprecate Old Function, Other Minor Changes

Rixxan:
# Description
Implements a few minor code changes and one deprecated function which was goes-nowhere-does-nothing. 

Full notes: 
- Updates the Pylintrc ignore file
- Updates the Python version we test against to the latest 3.11 version
- Clarifies a few LANG functions to specify what we return (shuts up some linters)
- Updates dependencies to later versions
- Clarifies some conditional logic
- Officially deprecates openurl() in favor of just using webbrowser.open. 

## Type of change
- Minor Edits and Function Maintenance


C1701D (approved):
LGTM

Issue #1827, opened 2023-01-23
Source Tarball from Release Does Not Display Git Hash


dvdmuckle:
This is with at least Version 5.8.0, and also happens with 5.8.1, tested on Fedora 37.

The new tarball provided with the intention of running from source, and on Linux, will show the version as "5.8.0+UNKNOWN."
![image](https://user-images.githubusercontent.com/8870292/214176917-7c5645a7-3eb1-4396-8ce4-42a5824ffdb0.png)
While this isn't a bug per say, it doesn't really align with the feature for the Windows version, which will show the Git hash as it's running "frozen" and thus using the included `.gitversion` file. This also affects the new Flatpak version of EDMC, as 1. `git` is not included in the Flatpak runtime and 2. There's a PR for Flatpak to switch to this new tarball, which doesn't include the `.git` folder in it.

Pertinent log:
```
'fatal: not a git repository (or any parent up to mount point /)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).' doesn't look like a valid git short hash, forcing to None
2023-01-23 23:41:23.323 UTC - INFO - 34582:140044549572416:34582 EDMarketConnector.__main__:2110: Startup v5.8.1+UNKNOWN : Running on Python v3.11.1 (main, Jan  6 2023, 00:00:00) [GCC 12.2.1 20221121 (Red Hat 12.2.1-4)]
```

One possible solution for this is to include the `.gitversion` file in the tarball, and always check if this file exists, instead of just when the app is running in a "frozen" state. If that sounds agreeable, I can submit a PR with the changes to make that work.

Rixxan:
I'd welcome a PR for this! We should probably only rely on the .gitversion file if the exact hash can't be found (as in, check for git status and then fall back to hash) and might want to have some indication of relying on the githash, but it seems like a good idea.

PR #2118, opened 2023-12-24
Add .gitversion file to Linux archive

dvdmuckle:
Also renames a couple things in the GHA workflow file for building after a release.


dvdmuckle:
I need to do some additional tests but this is how I'm looking to address #1827

Rixxan:
This all looks good so far. Let me know when your testing is complete and I'll do my own passes through it :)

dvdmuckle:
Finished my testing! I undid the renames I mentioned in the PR comment, as I didn't realize renaming GHA workflows changes the link to them.

dvdmuckle:
Realized from the comment you left on #1827 that the checks should be done in the other order (IE, check for `git rev-parse --short HEAD`, then fall back to `.gitversion` if we don't get a valid git sha back). I can change the behavior here if you want.

Rixxan:
Ideally, yes the preference should be the git HEAD, and then the .gitversion file, and then an error out to "unknown".

dvdmuckle:
@Rixxan Done!

Rixxan (approved):
Changes look good to me! Thanks for contributing this.

PR #2119, opened 2023-12-24
[Fix] Add Required Dupe Imports

Rixxan:
Fixes a bug introduced in 724dd2ce6af07c8be10a4a7eabb72b91d2d2fe2d

These imports are indeed duplicates, however, they are required for the "an instance of EDMC is already running" popup to function as intended.


C1701D (approved):
LGTM, still doesn't work if no journal files exist, working as intended.

PR #2139, opened 2024-01-04
[635] Remove Ordered Dict

### LINKED ISSUE #635: Remove collections.OrderedDict spam
For 90% of the uses, OrderedDict was used to preserve insertion order. In those cases for python3.7+ the builtin dict type does this for us.

Rixxan:
# Description
Before Python 3.7, Dictionaries did not maintain their order by default. However, in the modern age of Python, Dictionaries do maintain their order and this import and complication is largely unnecessary. This removes all orderedicts from the project. 

This is technically a 3.11 milestone task, however can be done as a drop-in replacement patch in the 5.10.x line.

## Type of change
- Old Code Rewrite

Closes #635

### DIFF
--- a/companion.py
+++ b/companion.py
@@ -28,7 +28,7 @@ import urllib.parse
 import webbrowser
 from email.utils import parsedate
 from queue import Queue
-from typing import TYPE_CHECKING, Any, Mapping, OrderedDict, TypeVar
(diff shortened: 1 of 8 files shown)


C1701D (approved):
One comment that may no longer be needed, rest looks good.

PR #2142, opened 2024-01-04
[2127] Cleanup Wix and Manifest Files

### LINKED ISSUE #2127: Remove Old Wix Files
**Describe the enhancement**

Remove files which are no longer in use by EDMC and are holdovers from the prior Wix build system

Since the handover to Inno Installer from the custom-rolled Wix setup, a number of files are still included but abandoned in place and are no longer required or used anywhere in the program. Where practical, we should remove these files and no longer bundle them.

Rixxan:
# Description
Since the handover to Inno Setup from Wix, some files are no longer called by anything and can be safely removed. Additionally, in the interests of cleanup, some files can be moved from top level to a resources folder. As these files are not actually included in the program (only called buy the installer) it's safe to move these files. Also moves the .xcf (basically the image template file) to resources to not clutter up top level.

Finally, add the icon to the EDMC.exe file. Because everyone deserves to look nice. 

## Type of change
- Cleanup

Closes #2127


C1701D (approved):
LGTM

PR #2144, opened 2024-01-14
Enhancement/662/edsm rate limit

### LINKED ISSUE #662: [Paranoia] Make EDSM plugin aware of any API limits
I don't *think* the player API on EDSM actually implements rate limiting.

Presumably if Anthor ever did it would be the same 'leaky bucket' he uses on the non-player endpoints.

So, out of paranoia, let's make the EDSM plugin aware of those headers and do *something* about them, even if it's only log if the bucket gets near empty.

i.e. on the systems API:

```
curl --verbose 'https://www.edsm.net/api-v1/systems?systemName=Sol'
...
< x-rate-limit-limit: 360
< x-rate-limit-remaining: 359
< x-rate-limit-reset: 10
...
```

Rixxan:
# Description
Currently, EDSM does not implement a hard rate limit on user API calls. Even with the soft rate limit provided by the API, it should be exceptionally hard to reach that limit. However, we should respect that limit.  

Tested during a few hours of gameplay, and no errors from EDSM were reported or recorded. 

## Type of change
- New Feature

Closes #662

### DIFF
--- a/EDMC.py
+++ b/EDMC.py
@@ -1,5 +1,4 @@
 #!/usr/bin/env python3
-# flake8: noqa TAE001
 """
 EDMC.py - Command-line interface. Requires prior setup through the GUI.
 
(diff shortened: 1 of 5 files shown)


C1701D (approved):
LGTM

PR #2154, opened 2024-02-08
[#2077] Ensure Providers Are Valid

### LINKED ISSUE #2077: Station name not shown in UI before change in station provider
**Please check the [Known Issues](https://github.com/EDCD/EDMarketConnector/issues/618) in case this has already been reported.**

**Please also check if the issue is covered in our [Troubleshooting Guide](https://github.com/EDCD/EDMarketConnector/wiki/Troubleshooting).** It might be something with a known work around, or where a third party (such as EDSM) is causing logging that is harmless.

 - Version: 5.9.5
 - Game Version: 'Odyssey'
 - OS: Fedora 38
 - OS Locale: German
 - Please attach **BOTH** log files, by dragging and dropping them into this input: Actually I can't find any of the log files, need to work on that as well.
 - 
[EDMarketConnector-debug.log](https://github.com/EDCD/EDMarketConnector/files/12653181/EDMarketConnector-debug.log)

**Describe the bug**
~I tried to finally update the Flatpak version of EDMC from 5.8.1 to 5.9.5, but when testing it, it seems to not be able to read the currently docked station from the journal anymore.
This is while being docked at a station. Leaving and re-docking also doesn't help. When I go back to 5.8.1 it works fine again.~

I updated the Flatpak version from 5.8.1 to 5.9.5, but when testing it, it seems to not be able to read the currently docked station from the journal anymore. While debugging, I found that in my config I still had the line
```
station_provider = eddb
```

Interestingly removing the line also didn't fix the issue. Only when I enter the configuration menu and select Inara as the station provider and confirm the change, the station is shown in the UI again. Changing it back to EDSM afterwards also works fine. But it seems that the selection of the default provider is failing somehow.

**To Reproduce**
1. Remove line with `station_provider` from `EDMarketConnector.ini`
2. Start EDMC
3. See no Station name in UI
4. Enter configuration menu, set station provider to Inara
5. Station name is now visible

**Expected behavior**
Station name shown in UI

**Screenshots**
Bildschirmfoto vom 2023-09-18 22-08-30

Rixxan:
# Description
In rare instances, if an invalid System, Station, or Plugin provider is set in the configuration (such as EDDB in Station or System provider), this can result in the program breaking. This alleviates this issue by checking if the set configuration values are valid at startup. If not, reset them to the default values and throw an error to the user. 

This fix also includes the required LANG changes to allow localization. 

image

## Type of change
- Bug Fix

Closes #2077

### DIFF
--- a/EDMarketConnector.py
+++ b/EDMarketConnector.py
@@ -2162,6 +2162,7 @@ def validate_providers():
         popup_text
     )
 
+
 # Run the app
 if __name__ == "__main__":  # noqa: C901
     logger.info(f'Startup v{appversion()} : Running on Python v{sys.version}')
--- a/EDMarketConnector.py
+++ b/EDMarketConnector.py
@@ -807,6 +807,9 @@ class AppWindow:
         self.w.bind_all('<<CompanionAuthEvent>>', self.auth)  # cAPI auth
         self.w.bind_all('<<Quit>>', self.onexit)  # Updater
 
+        # Check for Valid Providers
+        validate_providers()
+
         # Start a protocol handler to handle cAPI registration. Requires main loop to be running.
         self.w.after_idle(lambda: protocol.protocolhandler.start(self.w))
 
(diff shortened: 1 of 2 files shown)


C1701D (approved):
LGMT

PR #2164, opened 2024-02-25
[1181] Save Coriolis URL override when changing languages

### LINKED ISSUE #1181: Changing language will cause non-auto (default) Coriolis override to become auto
This is due to the dropdown for Coriolis override setting having the translated word as the label, and that being all you get back.  But after a language change the translation will be different, not match up, and the code re-sets it to auto.

Proposed solution:
Store the translations when constructing the preferences tab and reference those not `_('Auto')` and the like in the back-conversion.

C1701D:
# Description
Currently changing languages may cause Coriolis URL overrides to revert back to auto (default) if the current selection has a different display string between the old and new language. This PR fixes that by saving the old display strings when creating the prefs window and checking against those when closing it. 

## Type of change
- Bug Fix

## How Has This Been Tested?
- Switching between languages with and without changing the override setting at the same time.
- Manually editing Registry to ensure original fallback to default still works.

Closes #1181


Rixxan (approved):
LGTM!

PR #2165, opened 2024-02-25
[2163] Update Inara plugin to better handle Thargoid Combat events

### LINKED ISSUE #2163: Update Inara plugin to use ``isThargoid`` flag for Interdictions
**Describe the enhancement**
In May 2023 a new parameter called ``isThargoid`` was added to the ``Interdicted`` and ``EscapeInterdiction`` journal events. These events now no longer contain the ``Faction`` parameter for Thargoid Interdictions. This causes the Inara plugin to throw Exceptions when trying to handle those events as it expects either a ``Factrion``, ``Power`` or ``Interdictor``/``Interdicted`` parameter to be present.

**Proposed soluten**
Check if the ``isThargoid`` flag is present and set when handling those events and then use it instead of the other parameters to set ``opponentName``.

C1701D:
# Description
In May 2023 a new parameter called ``isThargoid`` was added to the ``Interdicted`` and ``EscapeInterdiction`` journal events which caused issues with the way the Inara plugin handled Interdiction events. Additionally Thargoid caused deaths didn't show up in Inara's Combat Log.
This PR:
- Updates the Inara plugin to check for that ``isThargoid`` flag and set the ``opponentName`` in Thargoid related Inara API updates to ``Thargoid`` if it is set.
- Adds a check to the ``Died`` event to use the Killer's ship's name as ``opponentName`` if the Killer's name is not present in the journal entry (as is the case for Thargoid caused deaths).
- Rewrites the ``EscapeInterdiction`` path to match the other Interdiction paths.
- Updates the Interdiction related Paranoia checks to also work in cases of missing and not just empty parameters.

## Type of change
- Match journal update
- Bug Fixes
- Minor Code Rewrites

## How Has This Been Tested?
- Testing multiple Interdiction source (Player, NPC and Thargoids) and outcome (interdicted and escaped) combinations in game.
- Checking that getting interdicted and destroyed by Thargoids shows up properly in Inara's Combat Log.

Closes #2163

### DIFF
--- a/plugins/inara.py
+++ b/plugins/inara.py
@@ -1008,7 +1008,8 @@ def journal_entry(  # noqa: C901, CCR001
             opponent_name_issue = 'opponentName' not in data or data['opponentName'] == ""
             wing_opponent_names_issue = 'wingOpponentNames' not in data or data['wingOpponentNames'] == []
             if opponent_name_issue and wing_opponent_names_issue:
-                logger.warning('Dropping addCommanderCombatDeath message because opponentName and wingOpponentNames came out as ""')
+                logger.warning('Dropping addCommanderCombatDeath message'
+                               'because opponentName and wingOpponentNames came out as ""')
 
             else:
                 new_add_event('addCommanderCombatDeath', entry['timestamp'], data)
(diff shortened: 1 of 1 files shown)


Rixxan (approved):
Good changes here, everything seems in order.

Issue #1654, opened 2022-08-31
Inara: Add `shipMaxJumpRange` and `shipCargoCapacity` to `setCommanderShip`


Athanasius:
> minor update for Inara API `setCommanderShip` event - there will be possible to set `shipMaxJumpRange` and `shipCargoCapacity`, for the respective properties in the journal's `Loadout` event (`MaxJumpRange` and  `CargoCapacity`). Both are optional and those values will be displayed on Inara in the next site update, so no rush if you will wish to implement it.

PR #2180, opened 2024-03-18
Add Inara jump range and cargo capacity

dvdmuckle:
This is an attempt to address #1654. I need to test this but I'll leave a comment when I've confirmed this is working ass intended.


Rixxan:
How's the testing going? This looks good from my eyes :)

dvdmuckle:
So, to test this I tried buying a stock Sidewinder at a station, and then checking my Inara profile to see if the jump range was there. Unfortunately, I couldn't find the jump range there. I tried just testing again with a normal play session that included switching ships, and I added some good ol' `print` commands to the cases I added in this PR, but I couldn't find those `print`s.

dvdmuckle:
I need to better understand when this journal event is emitted so I can figure out if it's an issue with how I'm testing this change, or with the code itself. Or, I may just be expecting Inara to update with info faster than it actually is.

dvdmuckle:
I just did some more testing, as far as I can tell this is correct. I couldn't figure out why my `prints` (that I also tried changing to `logger.info` and `logger.debug` with the correct log levels set) didn't show up, but as you said, this _looks_ fine.

Rixxan:
@dvdmuckle during your testing, did Inara in fact pick up on the details you were sending it eventually?

dvdmuckle:
@Rixxan that's something that I'm also not 100% sure on, as I tried testing with buying a new ship and seeing what showed up on Inara from there. I don't believe the updated info showed up immediately, but I'm also not sure how quickly info like that is supposed to show up on Inara.

Rixxan:
This is still on deck to test more as soon as I figure out how.

Rixxan:
Looking through this, I don't believe this is correct. Although it looks logical as to how you would expect Elite to send the data, it's not accurate to how the game sends events.

Testing has shown that shipMaxJumpRange is only valid for the setCommanderShip event type, which are only available for (so far) two events in game that I have found (loadout changes). This can be generated on a ship swap or on manually changing the loadout, but is not included in other instances where setCommanderShip is called such as on login. 

As such, the code as presented here will never be received from the game data and can't be processed. I'm currently working on an updated branch. 

I'm going to close this PR as rebuilding the PR would essentially be the same as opening a new one so I'll opt for that.

dvdmuckle:
Sounds good, I'll try implementing this elsewhere in the code.

Issue #2171, opened 2024-03-06
Python 3.9 error: TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'


chennin:
[Running from source wiki](https://github.com/EDCD/EDMarketConnector/wiki/Running-from-source) currently states (Dec 18, 2023 revision) that EDMC "will no longer work under Python 3.8 or earlier". However, EDMC does not run on 3.9 either.

```python
Traceback (most recent call last):
  File "/data/./EDMarketConnector.py", line 61, in <module>
    import killswitch
  File "/data/killswitch.py", line 20, in <module>
    import config
  File "/data/config/__init__.py", line 486, in <module>
    config = get_config()
  File "/data/config/__init__.py", line 480, in get_config
    from .linux import LinuxConfig
  File "/data/config/linux.py", line 17, in <module>
    class LinuxConfig(AbstractConfig):
  File "/data/config/linux.py", line 25, in LinuxConfig
    def __init__(self, filename: str | None = None) -> None:
TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'
```

Repro'd by checking out the code and `git checkout stable` (tag: Release/5.10.1), then `podman run -it --rm -v  $PWD:/data python:3.9 bash`, then in the container: `cd /data; pip install -r requirements.txt; python ./EDMarketConnector.py`.

Also reported on [ED Community Discord](https://discord.com/channels/89778537522802688/89782152853479424/1214828688219049994).

`str | None` syntax was apparently added in 3.10, though I don't have a source other than StackOverflow.

Athanasius:
Yeah, that's just a wiki update being necessary.  The thing to note really is the 'targeted against'.  That means no actual testing of prior versions, so there should be no real expectation of it working.

bgol:
Yes, added with 3.10: [Allow writing union types as X | Y](https://peps.python.org/pep-0604/)

Rixxan:
@chennin please try the following: 

Change the imports and files in config/linux.py to include:

```python
"""
linux.py - Linux config implementation.

Copyright (c) EDCD, All Rights Reserved
Licensed under the GNU General Public License.
See LICENSE file.
"""
from __future__ import annotations

import os
import pathlib
import sys
from configparser import ConfigParser
from config import AbstractConfig, appname, logger

assert sys.platform == 'linux'
``` 

I think we just forgot to add a future annotation there.

chennin:
[...]

Issue #1471, opened 2022-02-16
Consider packaging PIL (Python Image Library) for plugins


Athanasius:
EDMC-Screenshot has this as a dependency, but it includes architecture, and python-version, specific files.

They keep on having to update to include multiple relevant versions of PIL (so that they can support prior versions of EDMC as well).

We could hopefully make this a lot simpler by just packaging the ~4MB PIL ourselves.  Hopefully there will be no issues to doing so with the GH build.

PR #2183, opened 2024-03-23
[1133] Add ContextMenu Globally

### LINKED ISSUE #1133: Context menu in Options input text boxes
**Describe the enhancement**

Currently EDMC doesn't have context menu in Options screen input text boxes. You can't right click for example in Inara API text box and paste the key.

**Expected behavior**

Fully functional context (right mouse click) menu in all text input boxed.

**Considered workarounds**

N/A

**Additional context**

N/A

Rixxan:
# Description
Adds a context menu including Cut, Copy, and Paste options to all nb.Entry fields, primarily those in the config menu. 

fcjGMblQ11

This has been tested to ensure that any secrets cannot be extracted by using this system. This also leverages existing nb systems, so legacy MacOS functionality will be maintained (though they will not get this new feature).

Additionally, contains a fix to a ctypes prototype function that could break on some source installs running using 64-bit Python (not currently officially supported) and updates some developmental dependencies. 

Includes a fix for a rare bug when running from source under #1805 

Finally, this PR also adds the Python Image Library (Pillow) (#1471, https://github.com/NoFoolLikeOne/EDMC-Screenshot/issues/51)

## Type of change
- Enhancement
- New Package

## How Has This Been Tested?
- Tested on developer's machine, and tested to ensure that functionality cannot be leveraged to extract secrets
- Tested on multiple Entry fields with various input and output.

Resolves #1133, #1471, #2171

Related to #1805

### DIFF
--- a/myNotebook.py
+++ b/myNotebook.py
@@ -77,7 +77,7 @@ class Label(tk.Label):
     """Custom tk.Label class to fix some display issues."""
 
     def __init__(self, master: ttk.Frame | None = None, **kw):
-        # This format chosen over `sys.platform in (...)` as mypy and friends dont understand that
+        # This format chosen over `sys.platform in (...)` as mypy and friends don't understand that
         if sys.platform in ('darwin', 'win32'):
             kw['foreground'] = kw.pop('foreground', PAGEFG)
             kw['background'] = kw.pop('background', PAGEBG)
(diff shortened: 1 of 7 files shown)


myNotebook.py
```diff
@@ -87,15 +87,63 @@ def __init__(self, master: ttk.Frame | None = None, **kw):
...
+
+    def paste(self) -> None:
+        """Paste the selected Entry text."""
+        if self.selection_present():
+            self.delete(tk.SEL_FIRST, tk.SEL_LAST)
+        try:
+            text = self.clipboard_get()
+            self.insert(tk.INSERT, text)
+        except tk.TclError:
+            pass  # No text in clipboard or clipboard is not text
```
C1701D, line 135: Shouldn't the deletion part be done after checking if the clipboard contains pastable text? Otherwise it would just delete the text without replacing anything which doesn't seem to be how paste should operate.

Rixxan:
@nofoollikeone I don't know if you still maintain EDMC-Screenshot or not, but this seems to have been a request that your plugin in particular used, and it seemed to simply be the best solution for our purposes here so hopefully it helps!

C1701D (approved):
LGTM

Issue #1812, opened 2023-01-16
Rename the `_()` translation function


Athanasius:
An unfortunate side-effect of forcing `_()` into `builtins` for translation of strings is that it means the `_` for things like "I don't care about this return value" clashes.  In general it was a **REALLY** bad choice given the uses of `_` normally.
```python
        should_return: bool
        new_data: dict[str, Any]
        should_return, _ = killswitch.check_killswitch('capi.auth', {})
```
Here I've changed `new_data` in the function call to `_`.  After this the use of `_("string to be translated")` breaks:
```python
Traceback (most recent call last):
  File "C:\Program Files\Python311\Lib\tkinter\__init__.py", line 1948, in __call__
    return self.func(*args)
           ^^^^^^^^^^^^^^^^
  File "C:\Users\Athan\Documents\Devel\EDMarketConnector\EDMarketConnector.py", line 1128, in capi_request_data
    self.status['text'] = _('Fetching data...')
                          ^^^^^^^^^^^^^^^^^^^^^
TypeError: 'dict' object is not callable
```
Proposal:
Change the translation `_()` to `_t()` - it's only one extra character.

Issues:
*minor* - PLUGINS.md mentions manually setting this up (does that code even *work* ?), that needs updating.  But as plugins would have to manually set things up in this manner nothing will break by changing this in core code.

Follow-ups:
Identify any places where `_` as "throw this away" would be appropriate and update them.  There's at least some killswitch related code.
Check over all the developer documentation to be sure it's up to date with the change.

Rixxan:
Honestly, I'm looking at just dropping forcing the builtin entirely. It confuses type checking, it makes things look weird, and it's not very clear where it's set or what it's doing. 

It'd be easier to just add a dummy function to the l10n _Translation class and accept one more direct import.

I don't think it's too much of a design sacrifice to do something along these lines: 

Add this to l10n's _Translations class
```py
    def tl(self, *args) -> str:
        """
        Dummy loader for the translate function. Used as shorthand.
        """
        return self.translate(*args)
```
and in files where we need translations
```py
from l10n import Translations as Tr
...
def somefunc():
        foo = Tr.tl("My String Here")
...
```
instead of the traditional
```py
def somefunc():
        foo = _("My String Here")
```

Rixxan:
Mocked up what would need to be done in https://github.com/HullSeals/EDMarketConnector/tree/enhancement/1812/rename_underscore_translate
Tasks:
[...]

PR #2189, opened 2024-04-01
[2188] Support Translation Language Override

### LINKED ISSUE #2188: Support Translation Language Override
**Describe the enhancement**

As a plugin author, I have a requirement to temporarily override the EDMC language with a different language for translating certain strings.

**Expected behavior**

The `l10n.Translations.translate()` function should take a parameter for overriding the language.

**Considered workarounds**

With assistance from @Rixxan, I have written a function in my plugin that does this, but think this could be a useful enhancement to contribute to the wider EDMC project, so available to all plugin authors. 

**Additional context**

I intend to submit a PR to add this functionality.

aussig:
## Description
The default `l10n.Translations.translate()` function only works with the language currently selected by the user and loaded in `l10n.Translations.install()`. This PR adds an additional optional `lang` parameter to `translate()`  which, when passed will temporarily load an alternative language and translate the passed string using that language.
 
## Type of change
Enhancement

## Implementation notes
* Some flake8 issues were flagged with the pre-existing imports. These were fixed.
* Some flake8 issues were flagged with the Darwin imports. These were not fixed as Darwin will be removed in 2186.
* The existing `context` parameter, passed to the function, was not well documented so this has been rewritten. Also, this parameter was modified inside the function, to strip it down to just the plugin name, and then used. IMHO this creates difficult to read code so I have introduced a new `plugin_name` variable to the existing code, to take this value, and the original `context` parameter is now left untouched. 
* I noted the comment "# TODO: There is probably a better way to go about this now." against the above extraction of the plugin name from `context` and would be happy to include a rework of this line if you can let me know what's the latest thinking on the approach.
* Apologies for the automatic trailing space removals from `PLUGINS.md`, I didn't spot these on commit but they don't seem unreasonable as a cleanup?

 
## How has this been tested?
By running EDMC in various languages and by modifying a plugin to use the new optional parameter.

Resolves #2188


Rixxan:
This in the pipe for review after #1812, @aussig this PR may require minor updates and edits after 1812 is merged due to potential conflicts.

aussig:
No problem @Rixxan, I'll await the merge on that. I took the opportunity to update this PR to the latest `develop`, resolving one conflict with imports caused by the removal of Darwin, I've used the latest imports from `develop` (I note you are ignoring Flake8 rule I001 - import alpha ordering - so I did the same)

Rixxan:
@aussig This is ready for a re-review given the new translation system. Just ping me when it's been given a once-over :)

aussig:
@Rixxan I've merged in the latest changes to `develop`, adding in the optional `lang` parameter to the new `tl()`.

Rixxan:
Love it! Great work. Seems to test fine on my end.

Issue #2199, opened 2024-04-12
The program just stopped opening, the reasons are unknown to me


Zapelmon:
[EDMarketConnector.log](https://github.com/EDCD/EDMarketConnector/files/14961660/EDMarketConnector.log)
[EDMarketConnector-debug.log](https://github.com/EDCD/EDMarketConnector/files/14961661/EDMarketConnector-debug.log)

 I decided to update the BioScan plugin, I updated, everything was fine, I later shut down my computer, and behold when I tried to start the program again, it wouldn't open.... reinstalling, cleaning everything did not help. (including uninstalling all plugins, etc.)

**To Reproduce**
1. Idk, it just happened
**Expected behavior**
The program will open...

**Additional context**
The program as far as I know was not run as administrator, and it is in Windows Defender exception.

Athanasius:
This looks like a simple case of "more exception cases, possibly a catch-all one, need adding to the code".

However that's only the necessary patch to get past the issue of "couldn't reach the github servers to retrieve killswitches".  So, @Zapelmon you might want to double-check local firewall configuration for EDMarketConnect.exe and/or raw.github.com (or other github.com hosts).

Zapelmon:
You'll laugh, but, it fixed itself XD. Thanks for the reply of course, and it's good that it was resolved so easily. But I still haven't figured out the cause, it had no problems with the firewall.

PR #2205, opened 2024-04-15
[2199] Better Exceptions on URL Errors

Rixxan:
In the event of a connection issue to github, there are a number of errors and exceptions that can be thrown. A number of exception types are accounted for. Some are not. 

This PR changes to the general requests exception to handle, and prevent EDSM from crashing if the killswtich file cannot be contacted for whatever reason. 

Also includes a minor fix to the EntryMenu class which will allow greater plugin compatibility with existing plugins which use an unclear nb.Entry. 

Discovered by #2199


C1701D (approved):
LGTM

PR #2211, opened 2024-04-26
[2203] Implement Total Exception Handling

### LINKED ISSUE #2203: Catch all exceptions at top level for better UX and bug reporting.
Given issues like https://github.com/EDCD/EDMarketConnector/issues/2199 , where failure to catch all relevant exceptions in the killswitch code can lead to the program crashing on startup, looking to the user like it's not working at all, it would be an improvement to User Experience (UX), not to mention gathering bug reports, to have a top level try/except to catch all unhandled exceptions.

1. There's already a try/except around the tkinter `root.mainloop()` call, in order to catch `KeyboardInterrupt` and attempt a clean shutdown.  This is where a catch-all `except` would go.
2. The body of this accept should:
    1. Log the exception, so there's some useful information to go on about the cause/source.
    2. Pop up a **blocking** (the program is in a BAD state and going DOWN, and you're outside of the tkinter mainloop now anyway), alert to the user about having encountered a fatal bug, and to please go make a bug report.  Given being outside the tkinter mainloop you're going to have to create a wholly new top-level window for this.  Something basic with the message and a single 'OK' button to close it.
    3. Close the program **with no attempt to clean up anything**.  There's no telling what bad state the program might be in.
3. Importantly such a general exception catch **will not stop any core or plugin code from catching and handling its own exceptions**.  It's just the place of last resort to handle any errors.

There are some choices for the exact nature of this general exception catch:

1. `except:` - completely bare, which catches literally every exception, including `KeyboardInterrup`, `SystemExit` and some others.
2. `except Exception as e:` - Will *not* catch those "very top level, system" exceptions, and is likely the behaviour you want.  Also means you can do something with `e` if needs be (but remember you can just `logger.exception(...)`, no need to reference it).
3. `except BaseException as e:` - will likely give much the same behaviour as with `Exception`.

To hammer it home, the point of this is:

1. Better User Experience - not having the program just disappear or look like it never started.
2. Encourage users who experience such a crash to make a bug report so the culprit code can be improved.

Rixxan:
# Description
This PR adds a high-level error handler for critical errors that are not properly handled inside of EDMC. This handles elements that can lead to the program crashing on startup and includes errors where the program may look like it's not working at all. 

This improves the user's experience and also encourages users to report errors to us on GitHub so that we can discover and fix errors more often - or at least direct the user on how to fix the error. Also includes the relevant translation information. 

Additionally, simplifies the "An EDMarketConnector process was already running" popup to utilize tkinter.messagebox elements, which significantly reduces the amount of overhead needed to show the prompt and also makes the prompt more clear that it is an error. 

# Example Images
image
image

# Type of change
- New Feature

# How Tested
Tested with both built and source versions of EDMC, with both intentionally broken and known working builds to generate errors.

# Notes
Relies upon #2209 for the updated translation system. Until that PR is merged in, this will appear with a larger diff than it really has. 

Resolves #2203

### DIFF
--- a/EDMarketConnector.py
+++ b/EDMarketConnector.py
@@ -11,9 +11,11 @@ from __future__ import annotations
 import argparse
 import html
 import locale
+import os
 import pathlib
 import queue
 import re
+import signal
 import subprocess
 import sys
 import threading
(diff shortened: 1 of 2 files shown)


C1701D (approved):
LGTM

PR #2213, opened 2024-04-27
[2166] Update Watchdog Type Hints

Rixxan:
# Description
Watchdog 4.0.0 included an update to the public API where FileSystemEvent, and subclasses, are now dataclasses, and their repr() has changed.

As such, (insofar as I can tell) the FileCreatedEvent is now a FileSystemEvent. MyPy also suggests that this is a FileSystemEvent as well. This type hint change should have no practical impact on the code, only how items that are calling this function expect data from it. 

# Type of Change
* Minor Fix

# How Has This Been Tested?
* MyPy tests, boot tests, and minor gameplay time to ensure no new errors are generated.

Replacement for #2166

### DIFF
--- a/monitor.py
+++ b/monitor.py
@@ -42,7 +42,7 @@ if sys.platform == 'win32':
     import ctypes
     from ctypes.wintypes import BOOL, HWND, LPARAM, LPWSTR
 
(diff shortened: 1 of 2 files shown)


C1701D (approved):
LGTM

PR #2214, opened 2024-04-28
[1801] FDEV ID Local Updater

Rixxan:
# Description

This PR removes the FDevID Submodule from the project and instead hands over responsibility for the two required files to EDMC locally. The check is now performed on a build of the EDMC, as well as on every check for updates using the internal updater. 

If an updated FDEVID file is found, the updates are pulled from EDCD's GitHub repo and applied automatically. No user input is required to keep these data files up to date. The logging output of the function can be suppressed, so as to not interfere with EDMC's console-only variant. 

This retires and removes the check for the FDevID Pop-up on the GUI version of EDMC, as well as the associated translations. 

NOTE: This updater does not ask if the user wants to update the files, it updates the files. This may be behavior we want to consider in greater detail. 

# Type of Change
Feature Enhancement

# How Tested
Tested on both built and source versions of EDMC under a variety of conditions including: 

1. Both files missing
2. One file missing
3. One file inaccurate
4. Both files inaccurate
5. Both files up to date
6. Simulated GitHub outage (Bad URL)

Related to #1801, #2111. May cause minor conflicts with #2209 due to removal versus rename in the function.


Rixxan:
Marking as draft until #2209 is complete due to potential conflicts

C1701D (approved):
LGTM

PR #2217, opened 2024-05-01
[#1462] Remove Deprecated load_module() Function

### LINKED ISSUE #1462: `load_module()` being deprecated for Python 3.12
```python
2021-12-10 10:53:37.148 - INFO - 6264:1884:1884 plug.Plugin.__init__:47: loading plugin "coriolis" from "C:\Users\Athan\Documents\Devel\EDMarketConnector\plugins\coriolis.py"
<frozen importlib._bootstrap>:283: DeprecationWarning: the load_module() method is deprecated and slated for removal in Python 3.12; use exec_module() instead
```

Rixxan:
# Description

importlib's `load_module()` function has been deprecated since Python 3.4, and has been toyed with being removed in the past. To replace it, the importlib `exec_module()` function has been introduced as a more "modern" method of importing modules, such as plugins. 

This is not a drop-in replacement, however. Some work needs to be done to resolve issues with some plugins and imports, particularly more complicated plugins which use dataclasses, decorators, and other more advanced functions. Previous attempts (such as 816d041) tried to simply work around this, however were not granular enough to suffice. 

This fixes a bug with the last attempt at fixing this issue by ensuring that `sys.modules` is aware of the imported module and can process it correctly in its namespace. Only once this is done does the program attempt to load the plugin. This also adds additional error checking and validation to ensure that a valid plugin is being sent to the program.

This will remove one of the major blocking events for us to move to Python 3.12 or beyond.,

# Type of Change
- Feature Enhancement

# How Tested

Tested on a number of third party and bundled plugins, including: 

- DistanceCalc
- EDMC-Canonn 7.3.1
- EDMC-Canonn 7.3.2
- HabZone-master (to ensure Plugins without Python 3 support still don't load)
- HuttonHelper 3.0.4
- Norohind's Spanch Router
- EDMC PluginTest

# Notes

The previous attempt was almost correct - but we needed to ensure that the system was aware of the module in order to handle it correctly. Additional testing of 3rd party plugins to ensure no plugins break with this is recommended. 

Resolves #1462


plug.py
```diff
@@ -66,19 +66,23 @@ def __init__(self, name: str, loadfile: str | None, plugin_logger: logging.Logge
...
-                if getattr(module, 'plugin_start3', None):
-                    newname = module.plugin_start3(os.path.dirname(loadfile))
-                    self.name = str(newname) if newname else self.name
-                    self.module = module
-                elif getattr(module, 'plugin_start', None):
-                    logger.warning(f'plugin {name} needs migrating\n')
-                    PLUGINS_not_py3.append(self)
+                spec = importlib.util.spec_from_file_location(filename, loadfile)
+
+                if spec is not None and spec.loader is not None:  # Check if spec and spec.loader are not None
```
Athanasius, line 71: This comment literally repeats exactly what the code states.  I can't imagine any replacement would be any less obvious, so just remove it ?

C1701D (approved):
Tested with multiple plugins, none broke, LGTM.

Athanasius:
> Tested with multiple plugins, none broke, LGTM.

Just to be sure... that selection of plugins definitely covered the ones that both provide and make use of a module themselves, right ?  `edmcoverlay` is one such, but I've no direct experience with it myself.  I think perhaps it, and some other plugins only have a top-level `__init__.py` in order to 'hack' them being earlier in the load order.

I'm struggling to find any other actual examples.  The idea is that some plugins might (also) provide a library for use by other plugins.  Was that case tested ?  Although, tbh, if it's that rare, I'd want to check they worked as intended on the old loading code as well....

PR #2224, opened 2024-05-02
[806] Add System Profiler Utility

### LINKED ISSUE #806: Small utility to report on OS version and other useful information
Given issues like https://github.com/EDCD/EDMarketConnector/issues/798 it might be useful to have a small utility to gather key points of information about a user's system:

1. OS version.
2. Hardware (if there's an easy, non-intrusive way to get at this, else we'd need to ask a user to run some third-party util).
3. Locale + encoding.
4. Check that the expected journals location actually contains journal files.
5. Check expected log destinations can be accessed (create and truncate files).

Anything else we can think of.  This utility would need to be as paranoid/bare bones as possible, e.g. it would be no good using our logging in it if our logging can cause issues.

Rixxan:
# Description

When receiving bug reports, there is a number of details about the user's environment that need collected. These include the version of EDMC, the OS version and architecture, the game version, and other diagnostic details. 

This PR adds a standalone system utility to gather all of this information in one place, as well as adding links to the utility from the Help menu of EDMC. The profiler uses only standard library tools we already bundle with EDMC, and minimal external utilities also present in EDMC. Some aspects of the system profiler are imported, such as config and Monitor, so any issues with those systems may not be diagnosable by this tool (though that will tell us things too!) 

The Profiler can be run as an EXE bundled with default distributions, or by the console if running from source with the `--out-console` argument. 

This PR also significantly reorders and cleans up the default bug report template, removing old guidance that has not been "new" for years. 

Finally, this PR optimizes several of the repository images, reducing image file sizes by 25%

# Type of Change
- Feature Enhancement

# Screenshots
327233323-57fae736-6cab-44ca-bc36-6987a15c909a

# How Tested

Tested under a number of conditions to ensure uniform output, including intentionally triggering failures in the ability to pull EDMC or Journal data. Tested in compiled and source builds. 

# Notes

Resolves #806 

Uses the new translation system


C1701D (approved):
LGTM

PR #2225, opened 2024-05-02
[1268] Handover to Path.is_reserved()

### LINKED ISSUE #1268: util_ships.py:ship_file_name() could use `pathlib.PureWindowsPath.is_reserved()`
Rather than trying to maintain a list of 'reserved' file names ourselves.

Rixxan:
# Description

This PR adds additional robust checking to make sure that a ship name is not in the Windows Reserved Filename list. This removes the older self-maintained version, and dumps the responsibility for that on Python's Pathlib module. (Why replicate existing work?)

This slightly renames how ship names can be output, however, remains safe for use in Windows and Linux (Linux will accept any filename). 

# Type of Change
- Feature Enhancement

# How Tested

Tested on a number of ship names using Windows reserved naming schemes. (Hello con, the Adder!)

# Notes

Resolves #1268

### DIFF
--- a/util_ships.py
+++ b/util_ships.py
@@ -5,18 +5,21 @@ Copyright (c) EDCD, All Rights Reserved
 Licensed under the GNU General Public License.
 See LICENSE file.
 """
+from pathlib import Path
 from edmc_data import ship_name_map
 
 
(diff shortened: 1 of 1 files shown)


C1701D (approved):
LGTM

PR #2239, opened 2024-05-25
UI rewrite in wxPython

ElSaico:
# Description
This rips off the current Tkinter UI and replaces it with a wxPython implementation. Some points on its favor:

- [Much better documentation](https://docs.wxpython.org/index.html)
- A rich library of widgets (see above)
- Handles most of the platform-dependant stuff, while exceptions are documented
- Both APIs are similar enough to make the inevitable rewrite of the plugins' UI code easier
- Built-in i18n support via gettext, eliminating the need for `l10n.py` - and allowing the use of tools such as [Poedit](https://poedit.net)
- A single approach to theming, instead of the current kludge to support non-ttk widgets
- Tools like [wxFormBuilder](https://github.com/wxFormBuilder/wxFormBuilder), which even generates Python code, can be used as a visual aid

# Example Images
TBA

# Type of Change
A rather large rewrite.

# How Tested
TBA

# Notes
wxPython's [Config](https://docs.wxpython.org/config_overview.html) is remarkably similar to ours, down to the use of registry on Windows and INI files on Linux.
This could mean even less code to maintain on EDMC's side, but its lack of a `get_list` equivalent impedes a direct port.


ElSaico:
Closing this as a practical approach to ttk themes was found on #2241.

PR #2241, opened 2024-05-27
Ttk-based theming

ElSaico:
# Description
This replaces all Tk widgets on the main screen with their Ttk equivalents and makes `theme.apply()` use actual Ttk themes, stored in JSON files. From there, it generates the colors to be used by `_update_widget()` - just like it currently does for the Default theme.
Also excludes Ttk widgets from `theme.register()`, as they no longer need it.

Plugin widgets are not affected, which allows for a release that prepares the room for yanking most of `theme.py` on a future major version.

# Type of Change
UI rendering.

# How Tested
Works for the basic widgets; an example plugin will be built to demonstrate all of them (which should also aid plugin developers from then on).

# Notes
Some issues emerged in its current state:
- Although it's explicitly declared, the Transparent theme does not set its font on `HyperlinkLabel`
- What to make of the "Normal text" and "Highlighted text" settings?
- Ttk widgets on *all* windows are affected, most notably Settings; is that acceptable - or maybe even desirable?

### DIFF
--- a/EDMarketConnector.py
+++ b/EDMarketConnector.py
@@ -400,7 +400,6 @@ if TYPE_CHECKING:
     # isort: on
 
 
-import json
 import tkinter as tk
 import tkinter.filedialog
 import tkinter.font
@@ -412,7 +411,6 @@ import prefs
 import protocol
 import stats
 import td
-from commodity import COMMODITY_CSV
 from dashboard import dashboard
 from edmc_data import ship_name_map
 from hotkey import hotkeymgr
@@ -446,6 +444,7 @@ class AppWindow:
         self.minimizing = False
         self.w.rowconfigure(0, weight=1)
         self.w.columnconfigure(0, weight=1)
+        theme.initialize()
 
         # companion needs to be able to send <<CAPIResponse>> events
         companion.session.set_tk_master(self.w)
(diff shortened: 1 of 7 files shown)


ElSaico:
`HyperlinkLabel` is the last major roadblock - effectively a `Label` which emulates an `active` state.
The latter is done with event bindings and toggling between two fonts, depending on whether the text must be underlined or not. All of it a bunch of Python-only state that bypasses the theming system, and can turn into a very ugly kludge if not approached correctly.

A trivial solution would be to make it subclass `ttk.Button` instead, which it actually behaves like, but it breaks plugins such as EDMC-Canonn and Hutton Helper that set `Label`-specific options...

ElSaico:
Seems like the HyperlinkLabel issue is pretty much solved, but there are some weird bugs upon switching themes:

- Starting out on Default, change to Dark/Transparent and then back (or vice-versa) leaves the following widgets stuck with the latter's colors:
  - all non-ttk
  - ttk.Label (main window only)
  - ttk.Entry/Spinbox/Combobox (text and main window only)
- Starting out on Transparent, change to Default/Dark and then back (or vice-versa) leaves the following widgets stuck with the latter's font:
  - ttk.Label/Entry/Spinbox/Combobox (settings window only)

ElSaico:
Problem: while yeeting the alternative title bar for the Windows native dark one would rid us of the former's gnarly minimize/restore handling, the Transparent theme doesn't work on the latter - as its background color needs to be the exact same as the rest of the window (and worse, it changes upon deselecting).

The only pre-Windows 11 way to change a title bar's color (at least without invoking a mountain of undocumented Win32 calls that'd defeat its purpose anyway) is through the Windows App SDK, a set of WinRT APIs that work all the way back to version 1809.
There are bindings thanks to [PyWinRT](https://github.com/pywinrt/pywinrt), but it lacks a crucial piece: [a Win32 interop API](https://learn.microsoft.com/en-us/windows/windows-app-sdk/api/win32/_uiinterop/) that [converts HWNDs from existing windows to the SDK's native WindowId](https://learn.microsoft.com/en-us/windows/windows-app-sdk/api/win32/winrt-microsoft.ui.interop/nf-winrt-microsoft-ui-interop-getwindowidfromwindow). I made a request (pywinrt/pywinrt#61) and the head maintainer is so prolific that it might not take much long to get it implemented, so there's still hope.

...also, this SDK might open the door for cleaner alternatives of our Win32 API sprawl in the future - even beyond the ongoing move to `pywin32`.

PR #2244, opened 2024-05-29
[#1469] Add Translation Push Check

### LINKED ISSUE #1469: Translations: Push/PR-merge check for any changed/additional phrases
Basically if we change `L10n/en.template` then we need to remember to upload to OneSky.  This should ignore the *comments*, because only the actual strings matter.  So, have such a check 'fail' if there are relevant changes and then I might remember to upload to OneSky.

Rixxan:
# Description

This PR adds a check to pushes to ensure that there are no missing translations, unused translations, or other out-of-date strings in the translation system. This check runs on every push. 

# Type of Change

Feature Enhancement

# How Tested

Tested by breaking and fixing translations and testing output on pushes in a test branch. And screaming at bash. Lots of screaming at bash. See: 
https://github.com/HullSeals/EDMarketConnector/actions/runs/9278648668/job/25529983265
https://github.com/HullSeals/EDMarketConnector/actions/runs/9278611138/job/25529891228

# Notes
Resolves #1469

### DIFF
--- a/.github/workflows/push-checks.yml
+++ b/.github/workflows/push-checks.yml
@@ -72,8 +72,4 @@ jobs:
 
     - name: translation checks
       run: |
-          output=$(python ./scripts/find_localised_strings.py --compare-lang L10n/en.template --directory . --ignore coriolis-data 2>&1)
-          if [ -n "$output" ]; then
-              echo $output
-              exit 1
-          fi
+          python ./scripts/find_localised_strings.py --compare-lang L10n/en.template --directory . --ignore coriolis-data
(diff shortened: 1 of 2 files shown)


Rixxan:
> If it doesn't already ... why not make `scripts/find_localised_strings.py` have different exit status that can be used ? 0 for no issues found, 1 for some other state, 2 if there _is_ a different other state, etc ?
> 
> Sure, you want to use the output if not empty here, but there might be other uses cases that would benefit from that more programmatic way to test the result of running the script.

Good idea. Implemented.

C1701D (approved):
LGTM

PR #2255, opened 2024-06-09
Enabled DeprecationWarning by default and fixed references

ElSaico:
# Description
Our use of `DeprecationWarning` falls into two traps described in [the `warnings` library documentation](https://docs.python.org/3/library/warnings.html):

1. Since Python 3.7 this particular warning is not shown by default, unless triggered directly in `__main__` (or the `-W` argument is set on the interpreter)
2. Calls to `warnings.warn()` display their own lines, which defeats the purpose of `DeprecationWarning` - the `stacklevel` exists for that, pointing one level down to the plugin function calling them.

# Type of Change
A convenience for plugin developers which would otherwise miss that they are using deprecated functions.

# How Tested
Running the latest version of some plugins showed immediate results - the use of `myNotebook.Label` being particularly common.

# Notes

### DIFF
--- a/config/__init__.py
+++ b/config/__init__.py
@@ -328,6 +328,7 @@ class AbstractConfig(abc.ABC):
         :raises OSError: On Windows, if a Registry error occurs.
         :return: The data or the default.
         """
+        # DEPRECATED: Migrate to specific type getters. Will remove in 6.0 or later.
         warnings.warn('get is Deprecated. use the specific getter for your type', DeprecationWarning, stacklevel=2)
 
         if (a_list := self._suppress_call(self.get_list, ValueError, key, default=None)) is not None:
@@ -386,6 +387,7 @@ class AbstractConfig(abc.ABC):
         See get_int for its replacement.
         :raises OSError: On Windows, if a Registry error occurs.
         """
+        # DEPRECATED: Migrate to get_int. Will remove in 6.0 or later.
         warnings.warn('getint is Deprecated. Use get_int instead', DeprecationWarning, stacklevel=2)
 
         return self.get_int(key, default=default)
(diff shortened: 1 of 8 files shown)


config/__init__.py
```diff
@@ -489,5 +486,9 @@ def get_update_feed() -> str:
     return 'https://raw.githubusercontent.com/EDCD/EDMarketConnector/releases/edmarketconnector.xml'
 
 
-# WARNING: update_feed is deprecated, and will be removed in 6.0 or later. Please migrate to get_update_feed()
```
Rixxan, line 492: This appears to be incorrect. If I try and call update_feed, I receive the following error:

```
Exception in thread update worker:
  File "C:\Users\david\AppData\Local\Programs\Python\Python311\Lib\threading.py", line 1045, in _bootstrap_inner
    self.run()
  File "C:\Users\david\AppData\Local\Programs\Python\Python311\Lib\threading.py", line 982, in run
  File "D:\Users\david\Documents\GitHub\EDMarketConnector\update.py", line 233, in worker
    newversion = self.check_appcast()
                 ^^^^^^^^^^^^^^^^^^^^
  File "D:\Users\david\Documents\GitHub\EDMarketConnector\update.py", line 184, in check_appcast
    request = requests.get(get_update_feed, timeout=10)
                           ^^^^^^^^^^^^^^^
NameError: name 'get_update_feed' is not defined. Did you mean: 'update_feed'?
```
This should instead throw a dep. warning.

Rixxan (changes requested):
In general looks good, but one particular thing stands out on initial testing.

config/__init__.py
ElSaico: Looks like `get_update_feed` instead to me...

Rixxan: Update, I am dumb.

ElSaico: A canonical event for every developer out there: spend a long time chasing something that was caused by a very silly mistake that you somehow overlooked more than once.

Rixxan (approved):
Error was due to my mistake, not PRs. Also replaces infi w/ SST which clears up the initial dep errors I see. 

I like it :)

C1701D (approved):
LGTM

PR #2259, opened 2024-06-10
[2215] Move Logging Directory

### LINKED ISSUE #2215: Follow xdg base directory specification for log files (instead of `$TMPDIR`)
**Describe the enhancement**

Would it be possible to change the log file directory location (currently `$TMPDIR`) to use the  [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html)? Or at least have an option to change it in the configuration file, or at runtime (via a flag, not an env variable) if the log directory info is needed before configurations can be read.

It's not clear to me how transient those log files are meant to be (since by default they reside in `/tmp` which would get cleared every reboot at least), so either `$XDG_STATE_HOME` (more persistent, but not generally a candidate for backups) or `$XDG_CACHE_HOME` (closest to `/tmp`) would fit it seems.

For what it's worth, since presumably EDMC must be reading the environment to begin with somewhere, something like `env.get("TMPDIR", env.get("XDG_CACHE_HOME", "/tmp")` could perhaps do this semi-hackily with minimal churn (where it is assumed `env` is a dictionary of the environment). So any users who had explicitly set `$TMPDIR` and also have whichever XDG directory is chosen (between state and cache) set wouldn't see changes in behaviour.

**Expected behavior**
More or less explained above.

**Considered workarounds**

Just export `TMPDIR=$XDG_CACHE_HOME`, though I'd like to avoid this if possible (more variable pollution).

Alternatively launch EDMC with `TMPDIR=... ./EDMark...` every time.

**Additional context**

I mostly found it strange how the config and plugin locations properly respected the XDG spec but not the log directory, thus this ask.

Rixxan:
# Description
This PR moves the default logging directory away from $TEMPDIR or %TEMP% and to the current app data directory (Linux: `~/.local/share/edmarketconnector/logs`, Windows: `%localappdata%/edmarketconnector/logs`). 

This change is done to better follow the XDG Base Directory specifications, and to unify where files exist to fewer locations. (From up to 3 locations down to 1 for log files, plugin data, and other app files). 

# Type of Change
Enhancement

# How Tested
Tested with various builds and install locations with both installers and user-frozen exe's to ensure logging refers correctly without errors. 

# Notes
Resolves #2215

### DIFF
--- a/EDMCLogging.py
+++ b/EDMCLogging.py
@@ -41,7 +41,6 @@ import logging
 import logging.handlers
 import os
 import pathlib
-import tempfile
 import warnings
 from contextlib import suppress
 from fnmatch import fnmatch
@@ -51,7 +50,6 @@ from threading import get_native_id as thread_native_id
 from time import gmtime
 from traceback import print_exc
 from typing import TYPE_CHECKING, cast
-
 import config as config_mod
 from config import appcmdname, appname, config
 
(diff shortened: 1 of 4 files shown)


Rixxan:
Marking as draft until 5.11.2 is out the door.

PR #2260, opened 2024-06-10
[2114] Pathlib Handover

### LINKED ISSUE #2114: Retire os.path in favor of pathlib
Pathlib and os.path are largely similar, and we use both fluently throughout EDMC. 

Pathlib is a newer library since Python 3.4+ (I think).

Since 3.6, [pip 519](https://www.python.org/dev/peps/pep-0519/) allows pathlib Paths to be passed as inputs to the functions to os.path. Why keep two different parallel dependencies around?

Rixxan:
# Description

This PR hands over the majority of os.path related functionality to the more modern pathlib environment. This change is designed to unify filesystem operations and not split EDMC across two different libraries. Every effort has been made to maintain compatibility, with the hopes that this is a seamless handover.

# Type of Change

Enhancement

# How Tested

Tested to ensure as seamless a transition as possible is performed. Tested in builds and installer runs. 

# Notes

Resolves #2114

### DIFF
--- a/EDMC.py
+++ b/EDMC.py
@@ -14,6 +14,7 @@ import locale
 import os
 import queue
 import sys
+from pathlib import Path
 from time import sleep, time
 from typing import TYPE_CHECKING, Any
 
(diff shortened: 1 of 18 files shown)


Rixxan:
Marking as draft until 5.11.2 is out the door.

Issue #1310, opened 2021-10-17
EDDN: Set `horizons` and `odyssey` on `navroute` if allowed


Athanasius:
We're discussing allowing these on the `navroute` schema before it goes live.

This issue is to ensure we get it correct upon 5.2.0 release.

PR #2276, opened 2024-07-13
pass Horizons/Odyssey flags in NavRoute

HansAcker:
# Description
The old workaround appears to be not needed anymore. EDDN accepts NavRoute with Horizons/Odyssey flags now.
EDDiscovery, EDO Materials Helper and others also set these flags.

# Type of Change
Removed workaround code from when EDDN did not accept these flags in NavRoute events.

# How Tested
I built and installed a release, produced NavRoute events and watched them come through the EDDN gateway.

# Notes
Mentioned in #1310

### DIFF
--- a/plugins/eddn.py
+++ b/plugins/eddn.py
@@ -1416,15 +1416,6 @@ class EDDN:
         #######################################################################
         # Elisions
         #######################################################################
-        # WORKAROUND WIP EDDN schema | 2021-10-17: This will reject with the Odyssey or Horizons flags present
-        if 'odyssey' in entry:
-            del entry['odyssey']
-
-        if 'horizons' in entry:
-            del entry['horizons']
-
-        # END WORKAROUND
-
         # In case Frontier ever add any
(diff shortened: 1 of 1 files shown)


Rixxan:
Hey @HansAcker! Thanks for submitting. 

Just wanting to check, this change you're suggesting, is compliant with https://github.com/EDCD/EDDN/blob/master/docs/Developers.md#horizons-and-odyssey-flags? Notably, we're only adding these keys when we're 100% certain and not adding Odyssey: False (Odyssey only exists if True)?

HansAcker:
As far as I can tell, it is. A NavRoute event always arrives after the game variant was determined, if I understand it right. I tested a modified build in both Odyssey and Horizons (live, not legacy) and the routes always had the correct flags on them.

It's probably of no great consequence to leave them out of route events, either.

Rixxan:
As far as I can tell this is correct and seems to be working. Looks good, let's give it a shot :)

PR #2359, opened 2025-01-22
Fix linux-setup.sh

jimmon89:
$HOME/bin should be $HOME/.local/bin

# Description

Fix the path for the default user binary directory on linux

# Type of Change

Bug Fix

# How Tested

script automatically finds the user binary directory with no manual intervention required

### DIFF
--- a/scripts/linux-setup.sh
+++ b/scripts/linux-setup.sh
@@ -13,14 +13,14 @@
 #######################################################
 # Determine where edmarketconnector.sh needs to go
 #######################################################
-# Really we need this to be "${HOME}/bin", so check that is in $PATH
-if [[ ":$PATH:" != *":$HOME/bin:"* ]];
+# Really we need this to be "${HOME}/.local/bin", so check that is in $PATH
+if [[ ":$PATH:" != *":$HOME/.local/bin:"* ]];
 then
-        echo "You need to have '${HOME}/bin' in your PATH"
+        echo "You need to have '${HOME}/.local/bin' in your PATH"
         echo "Please fix this (might require relogging) and try again"
         exit 1
 fi
-EDMC_BIN_PATH="${HOME}/bin"
+EDMC_BIN_PATH="${HOME}/.local/bin"
 if [ ! -d "${EDMC_BIN_PATH}" ];
 then
         echo "'${EDMC_BIN_PATH}' must exist and be a directory!"


Athanasius:
Sure, https://www.freedesktop.org/software/systemd/man/latest/file-hierarchy.html and https://specifications.freedesktop.org/basedir-spec/latest/ cite `$HOME/local/bin/`, not `$HOME/bin/`, but it would have been nice to cite them in this PR as the reason for the change.

`$HOME/bin/` was a thing long before such specifications, so isn't per se a **bug**.

Also, is this solving an actual *problem*, you've not cited an open EDMC issue, which is required as per https://github.com/EDCD/EDMarketConnector/blob/main/Contributing.md#work-on-issues

klightspeed:
It looks like `~/.local/bin` was introduced in the XDG base directory spec 0.8, and before that in 2014 in the systemd file hierarchy standard. Some things before that (e.g. Python 2.6 with PEP 370) treated `~/.local` as having the same kind of structure as `/usr/local`, and also had `~/.local/bin`

If you're going to put the launcher script into `~/.local/bin`, then you probably shouldn't error out that the directory isn't in the $PATH, since it likely wont exist unless e.g. you have pip installed something that includes a CLI script on a system where the environment isn't marked as being externally managed.

I do wonder how many people use this convenience script.

Athanasius:
> If you're going to put the launcher script into `~/.local/bin`, then you probably shouldn't error out that the directory isn't in the $PATH, since it likely wont exist unless e.g. you have pip installed something that includes a CLI script on a system where the environment isn't marked as being externally managed.

Except that the whole reason for this check is because other things assume that the launcher script **will** be in `$PATH`.  It's right there in the comment in https://github.com/EDCD/EDMarketConnector/blob/main/scripts/linux-setup.sh#L10-L11

1. a .desktop file doesn't support `~` or `$HOME`.
2. We're not enforcing the install being in a hard coded path
3. Therefore we must rely on `$PATH`.
4. Therefore there must be a script called `edmarketconnector` somewhere in `$PATH` in order for use of https://github.com/EDCD/EDMarketConnector/blob/main/io.edcd.EDMarketConnector.desktop to cause its invocation (the `Exec=edmarketconnector` line).

I'm not objecting to changing to `$HOME/.local/bin/` per se, but I am curious as to the actual **problem** with `$HOME/bin/`, if any.

Rixxan:
Until additional documentation is provided specifying this as an actual problem, and lacking an actual issue being opened to discuss this, this change cannot be accepted right now. 

If consensus is derived that a change should be made, this PR would be welcomed back, but without such consensus not change is warranted at this time.

PR #2372, opened 2025-02-21
Add code to track PP2.0 rank, merits, etc

### LINKED ISSUE #2371: Implement Powerplay 2.0 into `state`
**Describe the enhancement**

Presently, it doesn't look like information on Powerplay 2.0 is available to plugins through the `state` interface. This information, such as what power you're pledged to and how many merits you have, is part of the Startup event and is emitted on logging into the game. Having this information would allow plugin developers to craft a couple plugins around Powerplay 2.0, the simplest of which could display how many merits you have at launch, so players can compare to how many merits they have in game (merit earning is not presently reported as an event sadly) when constructed reports on PP2.0 operations.

**Expected behavior**

Add to the `state` interface a subsection where one can query for Powerplay 2.0 information, as documented here: https://elite-journal.readthedocs.io/en/latest/Startup/#powerplay

**Considered workarounds**

I can't think of any workarounds to this at the moment, but I can take a crack at implementing this and submitting a PR.

**Additional context**

If it wasn't obvious, I'm looking to craft a plugin that will report the number of merits one has when they start the game. It doesn't look like that's currently possible with EDMC.

dvdmuckle:
# Description

Adds some code to `monitor.py` as well as the Inara plugin to track Powerplay 2.0 rank, merits, and more

# Example Images

# Type of Change

New Feature

# How Tested

Ran EDMC with some debug code to make sure that `monitor.py` would pick up on the PP event. Inara also tested and confirmed working, I was able to refresh my profile and see the new up to date rank and merit count on starting the game.

# Notes

Fixes #2371

### DIFF
--- a/PLUGINS.md
+++ b/PLUGINS.md
@@ -688,6 +688,7 @@ Content of `state` (updated to the current journal entry):
 | `StationName`[3]      |       `Optional[str]`       | Name of the station we're docked at, if applicable                                                              |
 | `MarketID`[3]         |       `Optional[str]`       | MarketID of the station we're docked at, if applicable                                                          |
 | `StationType`[3]      |       `Optional[str]`       | Type of the station we're docked at, if applicable                                                              |
+| `Powerplay`           |           `dict`            | `dict` of information on Powerplay
 
 [1] - Contents of `NavRoute` not changed if a `NavRouteClear` event is seen,
 but plugins will see the `NavRouteClear` event.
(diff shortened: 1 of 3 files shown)


dvdmuckle:
Realizing now, I need to update `plugins.md` with this information, yes?

Rixxan:
@dvdmuckle If you are able to update plugins.md with the new information, it'd be greatly appreciated!

dvdmuckle:
@Rixxan can do! Have a bit of a busy day today, unsure if I'll get to it today or tomorrow.

dvdmuckle:
Alright, added the docs. I optimistically guessed this would be in 5.13, let me know if you think it might be in a sooner/later version so I can change things.

Rixxan:
This is fantastic. Approved and welcomed!

PR #2414, opened 2025-04-23
Support dynamic plugin install/enable/disable

mkizub:
# Description
This PR allows plugin installation from zip archive, allows dynamically disable and enable plugins.
Dynamicall add/remove plugin's frame from main EDMC window, plugin settings tab from settings Notebook.

I assume that when this feature will be added to EDMC, plugin developers will test and update plugin shutdown
procedure, if plugin requires it.

# Example Images
image

# Type of Change
New Feature

# How Tested
Tested with several valid plugins, with python2 and broken plugins.

# Notes
No translation is provided for 'Install Plugin' button and 'running'/'broken' (plugin status label).

### DIFF
--- a/EDMarketConnector.py
+++ b/EDMarketConnector.py
@@ -440,6 +440,7 @@ class AppWindow:
         self.minimizing = False
         self.w.rowconfigure(0, weight=1)
         self.w.columnconfigure(0, weight=1)
+        self.plugin_id = 0
 
         # companion needs to be able to send <<CAPIResponse>> events
         companion.session.set_tk_master(self.w)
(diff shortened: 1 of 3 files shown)


mkizub:
Well, all pylint/type warnings are fixed.
Only translation check left.
But if I run 
**python scripts/find_localised_strings.py --lang=L10n/en.template**
it changes en.template in every line, because format has changed. Looks like it was not run for years.
For instance,

mkizub:
1. Without EDMC functionality to start/stop plugins dynamically - plugin developers will have no way to test correct plugin shutdown.
2. We may add a warning, if the plugin (we want to stop) does not have plugin_stop() - notify our user that plugin does not supports clean shutdown, and we recommend to restart EDMC.
3. Plugin dependancy is not clearly defined anyway. When you load plugins now - you make no dependancy check at all. And full-featured dependancy tree will be too complex to implement, even using explicit declaration from plugins (i.e. plugin A declares that it depends on plugin B) - because of versioning, because currently you may run several versions of the same plugins (if they reside in different directories), etc. I belive that plugin dependancy is out of scope for this PR. When (and if) plugin dependancy will be implemented - the code to shutdown plugin will be updated acordingly.

mkizub:
EDMC is hopeless. If you don't accept changes of one line length, I can't believe you'll accept this PR. Farewell.

PR #2466, opened 2025-07-24
Added support for panthermkii

vrza:
Added support for `panthermkii`, as the missing hash map key breaks journal processing.

```
  File "monitor.pyc", line 725, in parse_entry
KeyError: 'panthermkii'
```


Rixxan:
Thank you for the PR! I think this is handled by 76c3929. Closing as dupe - will push here in the next bit.

Issue #2486, opened 2025-09-13
RuntimeError: main thread is not in main loop and failing to read log files


GrLoopBack:
**Please complete the following information:**

[//]: # (You can gather most of this information with the EDMC System Profiler)
 
 - Version: ['5.12.1`]
 - Game Version: ['Odyssey']
 - OS: [Linux, Kubuntu 25.04]
 - OS Locale: [English]

**Describe the bug**

EDMC wont read the log files without me "reloading" the directory - Browse, enter same dir, OK. 

There are also several other errors:

```
2025-09-13 09:23:23.245 UTC - INFO - 19101:130903605726080:19101 EDMarketConnector.__main__:2080: Startup v5.12.1+fc00839d : Running on Python v3.11.9 (main, May 22 2025, 16:37:34) [GCC 14.2.0]
2025-09-13 09:23:23.246 UTC - INFO - 19101:130903605726080:19101 protocol.LinuxProtocolHandler.__init__:319: Web server listening on http://localhost:39453/auth
E:D Market Connector 5.12.1+fc00839d
2025-09-13 09:23:23.247 UTC - INFO - 19101:130903605726080:19101 killswitch.fetch_kill_switches:346: Attempting to fetch kill switches
2025-09-13 09:23:24.254 UTC - INFO - 19101:130903605726080:19101 killswitch.parse_kill_switches:419: Kill switches last updated 18 October 2021
2025-09-13 09:23:24.282 UTC - INFO - 19101:130903605726080:19101 EDMarketConnector.__main__:2209: Overriding tkinter default font to None at size 10
2025-09-13 09:23:24.283 UTC - INFO - 19101:130903605726080:19101 plug.Plugin.__init__:66: loading plugin "coriolis" from "/home/testing/ed_stuff/EDMarketConnector/plugins/coriolis.py"
2025-09-13 09:23:24.284 UTC - INFO - 19101:130903605726080:19101 plug.Plugin.__init__:66: loading plugin "eddn" from "/home/testing/ed_stuff/EDMarketConnector/plugins/eddn.py"
2025-09-13 09:23:24.285 UTC - INFO - 19101:130903605726080:19101 plug.Plugin.__init__:66: loading plugin "edsm" from "/home/testing/ed_stuff/EDMarketConnector/plugins/edsm.py"
[...]

PR #2499, opened 2025-10-03
[Bugfix] Fix race starting journal monitor

klightspeed:
# Description
Fixes a race identified where long plugin initialization times can result in the following error being thrown in the journal worker thread, requiring a preferences reload to have journals continuing to be processed:

```
Exception in thread Journal worker:
  File "/home/testing/.pyenv/versions/3.11.9/lib/python3.11/threading.py", line 1045, in _bootstrap_inner
    self.run()
  File "/home/testing/.pyenv/versions/3.11.9/lib/python3.11/threading.py", line 982, in run
  File "/home/testing/gitstuff/EDMarketConnector/monitor.py", line 461, in worker
    self.root.event_generate('<<JournalEvent>>', when="tail")
  File "/home/testing/.pyenv/versions/3.11.9/lib/python3.11/tkinter/__init__.py", line 1933, in event_generate
    self.tk.call(args)
RuntimeError: main thread is not in main loop
```

# Type of Change
Bug fix

# How Tested
Tested locally to ensure journal events start being processed after entering main loop.
Original reporter appears to report that the issue no longer occurs with this fix.

# Notes
Discussed in EDMC channel on EDCD discord.
Fixes: #2486

### DIFF
--- a/EDMarketConnector.py
+++ b/EDMarketConnector.py
@@ -754,7 +754,7 @@ class AppWindow:
         config.delete('username', suppress=True)
         config.delete('password', suppress=True)
         config.delete('logdir', suppress=True)
-        self.postprefs(False)  # Companion login happens in callback from monitor
+        self.w.after_idle(lambda: self.postprefs(False))  # Companion login happens in callback from monitor
         self.toggle_suit_row(visible=False)
         if args.start_min:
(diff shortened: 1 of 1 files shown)


Rixxan:
Thank you for the report and fix.

PR #2501, opened 2025-10-04
[Bugfix] Fix EDDN queue not processing at startup

### LINKED ISSUE #2484: EDDN queue processing does not actually start on app launch
From [`EDDNSender.__init__`](https://github.com/EDCD/EDMarketConnector/blob/main/plugins/eddn.py#L154):
```python
        # Initiate retry/send-now timer
        logger.trace_if(
            "plugin.eddn.send",
            f"First queue run scheduled for {self.eddn.REPLAY_STARTUP_DELAY}ms from now"
        )
        if not os.getenv("EDMC_NO_UI"):
            self.eddn.parent.after(self.eddn.REPLAY_STARTUP_DELAY, self.queue_check_and_send, True)
```
`queue_check_and_send` actually does not get called here. The problem is that `parent` doesn't live long enough for the scheduled event to execute. In fact, even if you pass ms=0 to `after`, it won't fix the problem.

---

`parent` throughout the plugin code is annotated as `tk.Tk`:
- [`This.__init__`](https://github.com/EDCD/EDMarketConnector/blob/main/plugins/eddn.py#L100):
```python
        self.parent: tk.Tk
```
- [`EDDN.__init__`](https://github.com/EDCD/EDMarketConnector/blob/main/plugins/eddn.py#L593):
```python
    def __init__(self, parent: tk.Tk):
        self.parent: tk.Tk = parent
```
- [`plugin_app`](https://github.com/EDCD/EDMarketConnector/blob/main/plugins/eddn.py#L2005):
```python
def plugin_app(parent: tk.Tk) -> tk.Frame | None:
````

This can be misleading, since delayed events for `tk.Tk` do indeed always execute after calling `.mainloop()`. However, if we trace the `parent` passed into `plugin_app`, we arrive at the [`get_app`](https://github.com/EDCD/EDMarketConnector/blob/main/plug.py#L103) function in `plug.py`:
```python
    def get_app(self, parent: tk.Frame) -> tk.Frame | None:
        """
        If the plugin provides mainwindow content create and return it.

        :param parent: the parent frame for this entry.
        :returns: None, a tk Widget, or a pair of tk.Widgets
        """
        plugin_app = self._get_func('plugin_app')
        if plugin_app:
            try:
                appitem = plugin_app(parent)
                if appitem is None:
                    return None
```

Here we see that `parent` is actually a `tk.Frame`. And if we dig a bit deeper to see where `get_app` is called, we find this code snippet in `EDMarketConnector.py`, [`AppWindow.__init__`](https://github.com/EDCD/EDMarketConnector/blob/main/EDMarketConnector.py#L531):
```python
        for plugin in plug.PLUGINS:
            # Per plugin separator
            plugin_sep = tk.Frame(
                frame, highlightthickness=1, name=f"plugin_hr_{plugin_no + 1}"
            )
            # Per plugin frame, for it to use as its parent for own widgets
            plugin_frame = tk.Frame(
                frame,
                name=f"plugin_{plugin_no + 1}"
            )
            appitem = plugin.get_app(plugin_frame)
            if appitem:
                # <...>
            else:
                # This plugin didn't provide any UI, so drop the frames
                plugin_frame.destroy()
                plugin_sep.destroy()
```
Because EDDN plugin doesn't normally have a UI, its parent frame gets destroyed before `.mainloop()` is called, **and thus the scheduled `queue_check_and_send` call is lost**. It will be called again only when the player docks on any station - meaning that queued messages from explorers can remain unsent for months, even with frequent and active gameplay. Actually, this is exactly how I discovered this issue.

klightspeed:
# Description
Workaround the `.after()` queue being cleared (or invalidated) when the widget that call was made through is destroyed after that call is made, which in this case resulted in the EDDN queue not being processed until some other action re-triggered processing of the EDDN queue.

# Type of Change
Bug Fix

# How Tested
rinkulu in #2484 confirmed this small fix resulted in the EDDN queue being processed on startup

# Notes
Fixes #2484

### DIFF
--- a/plugins/eddn.py
+++ b/plugins/eddn.py
@@ -183,7 +183,7 @@ class EDDNSender:
             f"First queue run scheduled for {self.eddn.REPLAY_STARTUP_DELAY}ms from now"
         )
         if not os.getenv("EDMC_NO_UI"):
-            self.eddn.parent.after(self.eddn.REPLAY_STARTUP_DELAY, self.queue_check_and_send, True)
+            self.eddn.parent.winfo_toplevel().after(self.eddn.REPLAY_STARTUP_DELAY, self.queue_check_and_send, True)
 
     def sqlite_queue_v1(self) -> sqlite3.Connection:
         """


Rixxan:
This patch (and the associated conversation in #2484) are really clever. Thanks for the PR!

Issue #2500, opened 2025-10-04
T11 issue


alexzk1:
**Please complete the following information:**

[//]: # (You can gather most of this information with the EDMC System Profiler)
 
 - Version: [5.13.1+0a6ce0d8.DIRTY]
 - Game Version: ['Odyssey']
 - OS: [Linux]
 - OS Locale: [English]

I see error in debug log:
```
Traceback (most recent call last):
  File "/mnt/sda2/alex/Games/EliteTools/EDMarketConnector/monitor.py", line 725, in parse_entry
    ship = ship_name_map[entry["Ship"]]
           ~~~~~~~~~~~~~^^^^^^^^^^^^^^^
KeyError: 'lakonminer'
```
Relevant part of json:
```
2025-10-04 10:29:15.533 UTC - DEBUG - 288939:140558097966784:288957 monitor.EDLogs.parse_entry:1871: Invalid journal entry:
b'{ "timestamp":"2025-10-03T22:35:02Z", "event":"Loadout", "Ship":"lakonminer", "ShipID":120, "ShipName":"RHEA II", "ShipIdent":"MS-02S", "ModulesValue":43148638, "HullHealth":1.000000, "UnladenMass":612.349976, "CargoCapacity":322, "MaxJumpRange":35.383110, "FuelCapacity":{ "Main":32.000000, "Reserve":0.600000 },
```
It seems like you have the list of all ships in the game and this new one does not fit.

klightspeed:
Could you please test https://github.com/klightspeed/EDMarketConnector/tree/loadout-fuelcap and see if it fixes the issue with an unknown ship causing the `Loadout` event to be rejected?

alexzk1:
> Could you please test https://github.com/klightspeed/EDMarketConnector/tree/loadout-fuelcap and see if it fixes the issue with an unknown ship causing the `Loadout` event to be rejected?

I took your 2 branches locally already `git merge klightspeed/fix-journal-startup-race` and `git merge klightspeed/develop-corsair`.
Will see how it goes. My problem was that my plugin does not take current carrier load, but there is frontier's bug too on "transfer". This needs couple days if it was no-load because of earlier crash into edmc.
Any way I think those 2 branches must be added ASAP.

alexzk1:
@klightspeed With that 2 branches it was no error in log already.
I just added 3rd `git merge  klightspeed/loadout-fuelcap`
[...]

PR #2502, opened 2025-10-04
[Bugfix] Prefer loadout fuel capacity if valid

klightspeed:
# Description
Uses the `FuelCapacity` from the `Loadout` event in the stored SLEF, falling back to calculating the main capacity and taking the reserve capacity if the ship is known.
Fixes a key error when processing the `Loadout` event from the journal when the ship in the loadout is not in the `ship_name_map`

# Type of Change
Bug fix

# Notes
Fixes errors such as those in #2500 when new ships are added to Elite Dangerous, until the new data for these ships are added to EDMC.

### DIFF
--- a/monitor.py
+++ b/monitor.py
@@ -719,11 +719,18 @@ class EDLogs(FileSystemEventHandler):
                 data_dict = {}
                 for module in entry['Modules']:
                     if module.get('Slot') == 'FuelTank':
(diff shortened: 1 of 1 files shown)


Rixxan:
This may need revisited for the implementation based on previous changes coming in 6.0. This may be a good candidate for a 6.1 feature.

PR #2533, opened 2025-11-22
Correctly decode lists in Linux .ini file, treat all config keys as lowercase

dvdmuckle:
# Description

Previous versions of the ini to toml import/export would not correctly decode lists, meaning things like Inara and EDSM API keys would not carry over correctly. This addresses that by in part using some decoding logic from 5.13.3. Additionally, config keys across both the .ini and Windows (kinda) are case insensitive. The changes here not only make sure config keys are exported to the toml as lowercase, but also convert any gets and sets to lowercase.

# Type of Change

Bug fix/enhancement.

# How Tested

Run on Linux, making sure the .ini export works, flipping configs back and forth and closing EDMC, and checking the config.toml to make sure the config took correctly. This needs to be tested on Windows, but I don't have an environment up for that.

# Notes

### DIFF
--- a/config/__init__.py
+++ b/config/__init__.py
@@ -43,7 +43,8 @@ import sys
 import tomllib
 import tomli_w
 from time import gmtime
-from typing import Any, Callable, Type, TypeVar
+from typing import Any, TypeVar
+from collections.abc import Callable
 import semantic_version
 from constants import GITVERSION_FILE, applongname, appname
 
@@ -474,7 +475,7 @@ class Config:
     @staticmethod
     def _suppress_call(
         func: Callable[..., _T],
-        exceptions: Type[BaseException] | list[Type[BaseException]] = Exception,
+        exceptions: type[BaseException] | list[type[BaseException]] = Exception,
         *args: Any,
         **kwargs: Any,
     ) -> _T | None:
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -59,7 +59,7 @@ jobs:
         # your codebase is analyzed, see https://docs.github.com/en/code-security/code-scanning/creating-an-advanced-setup-for-code-scanning/codeql-code-scanning-for-compiled-languages
     steps:
     - name: Checkout repository
-      uses: actions/checkout@v5
+      uses: actions/checkout@v6
 
     # Initializes the CodeQL tools for scanning.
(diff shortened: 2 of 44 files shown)


config/__init__.py
```diff
@@ -275,16 +275,16 @@ def _load(self):
...
     def get(self, key: str, default=None):
         """Return raw stored value."""
-        return self.settings.get(key, default)
+        return self.settings.get(key.lower(), default)
 
     def get_str(self, key: str, default="") -> str:
         """Return string value."""
-        val = self.get(key, default)
+        val = self.get(key.lower(), default)
         return str(val) if val is not None else default
```
kitlaan: We may want to revert to the original 5.13.x `get_str()` behavior.

When testing BGS-Tally, a default config shows a changed (read: wrong) behavior.

The notes here are for Windows, but I believe Linux should have the same behavior...

Test Case:
1. no prior BGST settings (in windows registry)
2. launch EDMC; copy newly created config.toml so as to have a reference
3. close EDMC, to make sure config.toml has been flushed
4. launch EDMC; check BGS-Tally conf tab
5. see that none of the things default-on are checked


For reference, the newly created toml (using default BGST values) will have:
```
bgst_showzeroactivity = "CheckStates.STATE_ON"
```
And the config UI shows an "unchecked" state for "Show Systems with Zero Activity".

If you click on the checkbox, the toml will now show:
```
bgst_showzeroactivity = "Yes"
```
Upon restarting EDMC, the checkbox will show "checked" properly.

dvdmuckle: I was also thinking we do something like the old Linux behavior and always set things as strings. I _think_ that would solve the issue of enums being set, instead of being converted to `Yes`. Outside of the `# type: ignore` in your suggestion, what else is different with what we currently have?

kitlaan:
Testing w/ latest change on Windows looks good.

* toml shows bool values as 0/1 now; they were migrated from `winreg` properly.
* BGS-Tally shows "Yes" and "No" strings now, and the loaded-with-migration config UI shows the proper toggles set.

dvdmuckle:
@Rixxan I think this is ready for a review!

Rixxan:
Looks good to me, I'll fix the linter gripes this weekend and get the RC out.

dvdmuckle:
Excellent, thank you!

PR #2540, opened 2025-11-29
Add localisation of stations names

### LINKED ISSUE #2404: Display localised stationname for `System Colonisation Ship`
**Describe the enhancement**

Frontier is using a localised version for the stationname of `System Colonisation Ship`. The key `StationName` now is `$EXT_PANEL_ColonisationShip:#index=1;` (the part after the `:` might be different) and there is a new key called `StationName_Localised` which does contain the actual name `System Colonisation Ship` (for EN, it uses localised names for the game language).

**Expected behavior**

The EDMC window should display the localised name of the station.

**Considered workarounds**

None.

**Additional context**

For EDDN messages the station name should NOT use the localised version.

inutt:
# Description
Adds localisation of station names, by converting strings like "$EXT_PANEL_ColonisationShip; Okawara Horizons" to "System Colonisation Ship Okawara Horizons". If the ship has an index parameter but no name, the index will be used in place of the name.

# Example Images
edmc

# Type of Change
Enhancement, fixes #2404.

# How Tested
Spent the evening flying back and forth to a colonisation ship, and tried forcing several values by temporarily overriding the 'value' variable in the new localising function.

# Notes
Currently only system colonisation ships appear to need localising, but other strings may need to be added in future.

The new code also depends on station "provider" plugins actually calling it - the Inara and spansh plugins call the common `station_name_setter` function in `common_coreutils.py`, but the EDSM plugin does its own thing and needed modifying to call the new code.

I'm unsure of the meaning of the index parameter in a StationName field. I'm assuming it's there to disambiguate stations that would otherwise have the same name, but it might mean something else.

### DIFF
--- a/L10n/en.template
+++ b/L10n/en.template
@@ -839,3 +839,9 @@
 
 /* edastro_core.py: Settings>EDAstro - Label on checkbox for 'send data'; */
 "Send data to EDAstro" = "Send data to EDAstro";
+
+/* common_coreutils.py: Colonisation ship name, with title; In files: common_coreutils.py:81:81 */
+"System Colonisation Ship {NAME}" = "System Colonisation Ship {NAME}";
+
+/* common_coreutils.py: Unidentified colonisation ship; In files: common_coreutils.py:84:84 */
+"System Colonisation Ship" = "System Colonisation Ship";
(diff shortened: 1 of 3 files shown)


Rixxan:
Thank you for the submission! I'll look at this one soon - I haven't had the time to actually find one of these ships in-game and check it out but this has been on my list of very annoying bugs I want to get fixed :)

PR #2542, opened 2025-11-29
Fix preference dialog not detecting tab change events

### LINKED ISSUE #2541: Preferences dialog does not show journal or output paths
 
 - Version: 6.0.0-beta2+48dafaf3.DIRTY
 - OS: Linux 6.17.9-arch1-1
 - OS Locale: en_GB

**Describe the bug**

The preferences dialog doesn't show the path to journals or output directory unless the value is changed

**To Reproduce**
1. Run the above version (6beta2)
2. Open File->Settings, Configuration tab
3. Notice journal file location entry box is empty, even if the value is set.

**Expected behavior**

Currently configured directory would be shown.

**Additional context**

This seems to be due to the new scrollable tabs.

Previously, the entry fields were updated when the NotebookTabChanged event fired, set in the `PreferencesDialog`'s `__init__` function:
```python
notebook.bind('<<NotebookTabChanged>>', self.tabchanged)  # Recompute on tab change
```
This event is only generated by the `ttk.Notebook` object within the new scrollable tabbed notebook, not the scrollable tabbed notebook itself.

**Please Confirm the Following...**

[//]: # (Add any other context about the problem here.)
- [x] I have checked the [Known Issues](https://github.com/EDCD/EDMarketConnector/issues/618) list to ensure this is not a duplicate
- [x] I have checked the [Troubleshooting Guide](https://github.com/EDCD/EDMarketConnector/wiki/Troubleshooting) to check for known workarounds

inutt:
# Description
Fix the detection of tab changes in the settings dialog, and makes the journal/output directory paths display again.

# Type of Change
Bugfix. Fixes #2541.

# How Tested
Restarted the app and opened the dialog several times. Ran linters and pytest.

### DIFF
--- a/myNotebook.py
+++ b/myNotebook.py
@@ -14,6 +14,7 @@ import tkinter as tk
 from tkinter import ttk, messagebox
 from PIL import ImageGrab
 from l10n import translations as tr
+from typing import Callable
 
 if sys.platform == 'win32':
     PAGEFG = 'SystemWindowText'
@@ -171,6 +172,7 @@ class ScrollableNotebook(Notebook):
         self,
         master: ttk.Frame | None = None,
         tabmenu: bool = False,
+        on_tab_change: Callable | None = None,
         *args,
         **kwargs
     ) -> None:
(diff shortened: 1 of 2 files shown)


Rixxan:
Appreciate the submission!

PR #2556, opened 2025-12-08
[Bugfix] Fix type mismatch when reading certain boolean config values

rinkulu:
Config values `no_systray` and `always_ontop` are saved as booleans (internally stored as strings in the .toml file); however, the code that uses these values attempts to read them with the `get_int` method. This method cannot parse the strings `"True"` and `"False"` and always defaults to 0 when encountering them.

From [`config.__init__.Config`](https://github.com/EDCD/EDMarketConnector/blob/c03e4b6f6f7a9247472daa8248629ac9b112b131/config/__init__.py#L286):

```python
    def get_int(self, key: str, default=0) -> int:
        """Adaptive int (handles booleans stored as ints)."""
        val = self.get(key.lower())
        if isinstance(val, int):
            return val
        try:
            return int(val)
        except (ValueError, TypeError):
            return default
```

*Note that this docstring is misleading and does not represent the actual behavior of this method (or maybe I'm just stupid and can't catch its meaning).*

This causes the application to de facto ignore these user settings. This PR fixes the problem by changing the related calls to `get_bool`, also removing excessive constructions around them (e.g., `bool(config.get_int(...))`).

### These changes also fix cases when EDMC would require a restart every time the user saved settings if `no_systray` is set to True.

Another solution to this issue might be rewriting the `get_int` method to make it correctly handle boolean-like strings similarly as `get_bool` does.

### DIFF
--- a/EDMC.py
+++ b/EDMC.py
@@ -39,7 +39,7 @@ import loadout
 import outfitting
 import shipyard
 import stats
-from commodity import COMMODITY_DEFAULT
+from commodity import COMMODITY_CSV
 from config import appcmdname, appversion, config
 from monitor import monitor
 from update import EDMCVersion, Updater, check_for_fdev_updates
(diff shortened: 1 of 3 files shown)


dvdmuckle:
IMO I feel like if you want to get a bool, you should get a bool, so I'm not sure about honoring the docstring here. Then again, if that was in fact the behavior in 5.13.3, we should honor it.

PR #2671, opened 2026-06-01
Download & Install from plugins browser

Niceygy:
Added `dev_install_plugin` in `plugin_browser.py`, that downloads the .ZIP file provided by the plugin registry to the user plugins folder.
Still needs a few things tweaking (I think you have a standard requests object? and more user visible logging would be good) but this is just a first past.

### DIFF
--- a/plugin_browser.py
+++ b/plugin_browser.py
@@ -26,6 +26,7 @@ from EDMCLogging import get_main_logger
 from l10n import translations as tr
 from ttkHyperlinkLabel import HyperlinkLabel
 import semantic_version
+import shutil
 
 logger = get_main_logger()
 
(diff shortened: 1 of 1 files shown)


Niceygy:
Update: uses `shutil.unpack_archive` to extract the archive to a subfolder in the plugins directory names after the plugin's `pluginName`. I do wonder if some plugins will have a trouble if an extra folder is in the ZIP file?

Niceygy:
Also, is there an EDMC internal error handler? At the moment I'm just throwing an exception but thought you might have a more elegant solution.

dwomble:
This presumes the zip file directly contains the plugin (load.py is at the top level of the zip file). That's how my ZIP files are but I've seen plugins where they contain a plugin folder name in the zipfile (load.py is in PluginName/load.py). I think it's probably worth handling this scenario.

